PR #170, opened 2015-09-11
134 document hyperpars

dfalster:
@richfitz Could you please review? 

Documents hyper-parameterisation function.

Also groups all functions for given module under single file. IMO this will make it easier to find relevant functions. 

Finally, for hyperpar in ffdev I thought it best to duplicate the code from hyperpar in ffw16, instead of just linking to the ffw16 function, so that the two strategies can potentially diverge over time.


dfalster:
haven't changed any functionality of the functions themselves.

richfitz:
Looks great to me, sorry this dropped off my radar.

PR #172, opened 2016-01-18
Downgrade version and add DSF as maintainer

richfitz:
Both moves are possibly contentious, so discussion can go here if you disagree :smile:

Also, did you know you can do PRs without a fork?  Pretty sweet. :fork_and_knife:


richfitz:
The master branch contains that same travis fix (f47e1a8) but the merge should still be fine.

PR #214, opened 2019-10-23
WIP abstract strategies

rafaqz:
Cleans up modularity of strategies
- [x] Strategies inherit from a generic Strategy class
- [x] Generic strategy methods are also generic - not related to specific physiology or environmental variables, except where they are universal
- [x] Decouple strategy details from any high level processes


rafaqz:
These changes are all working except `build_schedule` not being found and `lcp_whole_plant` not working for some unknown reason.

`lcp_whole_plant` seems to have so special-casing code I don't quite understand in plant.R. It might be good to discuss on the phone.  There are other issues like swapping `height` for `size` maybe shouldn't extend to FF16 for backwards compat, I just bulk swapped everything to get the tests working.

But otherwise I'm pretty happy this refactor went so painlessly.

dfalster:
feel free to call (mobile) if you want to discuss

PR #220, opened 2019-11-24
Abstract strategies2

rafaqz:
Refactored version with working abstract Strategy parent type, reusable assimilation types, and  passing tests.

Height has not been renamed to size as it created a lot of confusion with various other "size" methods, which will need to be discussed.


dfalster:
Great work @rafaqz. I've had a quick look but will give a detailed review ASAP. Probably early next week, as am at conferences all this week.

rafaqz:
No problem, enjoy the esa conference. The kohyama model is also half done, but I really needed to finish all this basic stuff properly first. Getting my head around the C++/R interface has also been a steep learning curve.

PR #235, opened 2020-03-09
Generalised model

rafaqz:
This PR contains al the work generalising the strategy and environment over the last few months. 

It should be useable as-is, as discussed it might be good to start playing with it to see if the tests are actually covering the real use-cases.

The problem of multiple model environments and templated parameters in RcppR6 seems to taken care of by simply using `Environment` instead of `FF16_Environment`.  The other RcppR6 templating problem remains, and functions that were originally in yaml are now written out.

I guess we can change the name from `hacks.cpp`, but to what?


dfalster:
Great progress @rafaqz, thanks for pushing this forward.

PR #237, opened 2020-03-26
decouple asismilation from ff16

rafaqz:


dfalster:
Happy to merge but apparently one of the checks is failing. @rafaqz - want to investigate?

Issue #244, opened 2020-07-09
tests passing but something isn't right


dfalster:
Having allowed multiple environment's with different strategies, all the tests are passing, but something isn't right. 

Here are examples from the vignettes. They kind of work, but really should be behaving identically to the way they were before

```
devtools::load_all()

pl <- FF16_Plant()
pl$state("height")

pl$set_state("height", 10)

env <- fixed_environment("FF16", 1.0)
pl$compute_vars_phys(env)
```

rates are slightly different

```
# Patches of competing plants

p0 <- scm_base_parameters("FF16")
p1 <- expand_parameters(trait_matrix(0.0825, "lma"), p0, FALSE)
p1$seed_rain <- 20
data1 <- run_scm_collect(p1)

matplot(data1$time, data1$species[[1]]["height", , ], lty=1, col=make_transparent("black", 0.25), type="l",
        las=1, xlab="Time (years)", ylab="Height (m)")
```
Kind of similar but should be identical

# Seed rain

Seed rain isn't reducing like it should, suggest competition isn't working properly:

```
p0 <- scm_base_parameters("FF16")
p1 <- expand_parameters(trait_matrix(0.0825, "lma"), p0, FALSE)

for(s in c(0.1, 10, 100, 1000)){
	p1$seed_rain <- s
	out <- run_scm(p1)
	cat(paste(s, " - ", out$seed_rains, "\n"))
}
0.1  -  1.98425257684682
10  -  198.425257684682
100  -  1984.25257684682
1000  -  19842.5257684682
```

rafaqz:
Interesting. I've been working completely to the tests.

Since coveralls is so high it might be some values that are being used in the tests mask the problem, instead of lines not actually being hit.

dfalster:
Indeed. Need to update tests to catch whatever is going on. 

Note the tests with SCM are much slower than normal, indicating that the algorithm is working harder than it should.

dfalster:
NB - @foxeswithdata & @aornugent are both considering the issues here, so please post updates into this issue thread, so that the other can stay informed. I’ll try and look at this again late next week.

dfalster:
We've narrowed down the area where problem was introduced by running this check:

```

devtools::load_all()
p0 <- scm_base_parameters("FF16")
p1 <- expand_parameters(trait_matrix(0.0825, "lma"), p0, FALSE)
p1$seed_rain <- 20
scm <-  FF16_SCM(p1)
scm$run()
scm$seed_rains
```

[...]

PR #247, opened 2020-07-21
WIP: Kohyama model

aornugent:
This compiles and runs, but I'm having trouble plugging competition into the right place.

```r
p0 <- scm_base_parameters("K93")
p1 <- expand_parameters(trait_matrix(0.059, "b_0"), p0, mutant = F)
types <- extract_RcppR6_template_types(p1, "Parameters")

scm <- do.call('SCM', types)(p1)

scm$run_next()
scm$state$species
# [[1]]                            [,1]        [,2]
# height                   2.000010e+00  2.00000000
# mortality                0.000000e+00  0.00000000
# fecundity                1.997293e-08  0.00000000
# seeds_survival_weighted  1.997293e-08  0.00000000
# log_density             -3.423764e-02 -0.03423743

scm$patch$environment$environment_interpolator$xy
#            [,1]      [,2]
# [1, ] 0.00000000 0.9999843
# [2, ] 0.06250032 0.9999843
# [3, ] 0.12500065 0.9999843
# ...
# [31,] 1.87500970 0.9999843
# [32,] 1.93751002 0.9999843
# [33,] 2.00001035 0.9999843
```

Here you see the environment is the same for all interpolated heights (actually size).

I suspect I've probably got the wrong thing in here:
https://github.com/traitecoevo/plant/blob/a6b7f92480dd19e9da05d19a9914bea3539a858d/inst/include/plant/models/k93_strategy.h#L52

which previously referenced `area_leaf_above(z, height)`, I believe referring to the shading an individual experiences from itself?


dfalster:
Actually, that may be the correct behaviour of the environment. A similar thing happens with the FF16 model

```
scm$patch$environment$environment_interpolator$xy
            [,1] [,2]
 [1,] 0.00000000    1
 [2,] 0.01267660    1
 [3,] 0.02535320    1
 [4,] 0.03802979    1
 [5,] 0.05070639    1
 [6,] 0.06338299    1
 [7,] 0.07605959    1
 [8,] 0.08873618    1
 [9,] 0.10141278    1
[10,] 0.11408938    1
....
```

The reason is you've only taken a single ODE step, so nothing has had time to develop yet.

I'm just updating some other elements of the model and will push these shortly.

dfalster:
Some corrections to model  in commit 0a7bbff1f06f7f7362c52af32cac2c35076a2644.  See further comments in the commit. 

Also adding some more missing features to the interface, including enabling all tests for K93 model. 

These still not passing though

aornugent:
Awesome, thanks. Perhaps I wasn't so far off after all.

After a couple of iterations it runs into the following error:
```
Error in SCM___K93__K93_Env__run_next(self) : Interpolated function as refined as currently possible 
```
which I take to mean that the difference between interpolated points is too small, given the control parameters?

Maybe this is breaking down because the environment is the same when there's only one individual:
```r
> scm$patch$environment$environment_interpolator$xy
            [,1]      [,2]
[1, ] 0.00000000 0.9999843
[2, ] 0.06250032 0.9999843
...
[31,] 1.87500971 0.9999843
[32,] 1.93751003 0.9999843
[33,] 2.00001036 0.9999921
```

aornugent:
Maybe not, I tried introducing a second species with the same result.

dfalster:
Hi Andrew,

Yes, something is up with the environment. I suspect the best pathway forward is for us to first pursue #244.  The FF16 model isn't yet behaving as it should, and I suspect there's something to do with the environment that's causing issues. Once we are confident that is working well, we can revisit K93 with more confidence.

dfalster:
Hi Andrew! Recent commits on devleop branch have updated the scaffolder, incorporating some of the changes we made on this branch. I then ported the Kohyama model to a new branch https://github.com/traitecoevo/plant/tree/kohyama-model2 working with the new scaffolder.

So I suggest checking that out, check if anything is missing then, work on that branch from hereon.

PR #249, opened 2020-08-07
WIP Fix enviro construction

rafaqz:
These changes seem to fix the problem with seed rain. But for some reason the environment always has to be constructed in `parameters.validate()`. I'm not sure if that's a problem or not.

A few constructor tests fail now with spline size not being zero, but the real tests all pass.


rafaqz:
Ok this is updated. It's not automated yet, the hyperpar needs to be passed in. Can we merge this and the Kohiyma model before doing that to get your changes?

Also this is failing tests for empty environment. Is there a reason it needs to be empty at any point? Can we just delete that test?

dfalster:
Closed via b42ad29

PR #252, opened 2020-08-13
get hyperpar from params

rafaqz:
uses "param_hyperpar" func to get the default.

This will need something in the scaffolder ot add the method next to the hyperpar method, as we can only get the "FF16_Strategy" Class name, not "FF16". So we have to link "FF16_hyperpar" to "FF16_Strategy". We could mess with the strings to assume the name has "_Strategy" appended, but its a bit hacky also.


dfalster:
Cool. I'll take a look.

Issue #229, opened 2020-02-26
move k_I into strategy


rafaqz:
currently defined in parameters

dfalster:
Yes. This should be part of the plant’s strategy, as different species may have different values. 

The light env would then be more accuractely interpreted as effective leaf area above.

aornugent:
I'm not fully sure I've understood the order of operations here, but I think the environment constructors need `k_I` to initialise a `Canopy` and do so before a strategy is called. 

`Parameters` uses the `k_I()` function in it's constructor, which appears to be defined in `ff16_environment`, and gets the `k_I` value from `canopy.k_I` which seems a bit circular to me!

dfalster:
One approach is to move k_I into strategy. That would also enable species
to differ in this. I can help with this.

-- 
Daniel Falster

Deputy Director  (Graduate Program) Evolution & Ecology Research Centre & ARC
Future Fellow | School of Biological, Earth & Environmental Sciences |
University
of New South Wales Sydney
P:+61-2-9065 9519 | A: Bdlg E26 Rm 5109 <https://goo.gl/maps/iS3Jc2YmYu62>  |
T: @adaptive_plant <https://twitter.com/adaptive_plant> | W:
http://danielfalster.com

PR #256, opened 2020-09-11
Kohyama93

aornugent:
Here's our first go at implementing a new (non-FF16) model in `plant`.

Some tests fail due to the adaptive environment spline going out of bounds, which can be fixed by lowering the default `k_I` to something on the order of 1e-3.


dfalster:
Only two tests failing now, due to value of k_I. What values have you found work better @aornugent ?

Feels like we need #229 to finish here!

dfalster:
Sorry, I should have said - I'd really like to get all the tests passing before merging the branch in.

aornugent:
> Only two tests failing now, due to value of k_I. What values have you found work better @aornugent ?
I set `k_I = 1e-3` and added `S_D` to the Rcpp class, even though it's hard coded in `k93_strategy.h`, for post-hoc calculations of seed rain. 

Only failing tests are numerics `qk` and `QAG` 🤞

PR #260, opened 2020-09-15
inherited environments

rafaqz:
Inheriting environments is fairly straightforward. 

Since 2015 C++ has been able to inherit constructors as well so we can just do:

```
using FF16_Environment::FF16_Environment;
```

And inherit all the constructors as well as methods. This means the environment.h files are nearly completely empty. You can just add methods for anything you want to change.

Any ideas on how to to this with the scaffolder?


rafaqz:
I don't know how the coverage went down, this deletes a few hundred lines of code, and adds 2

dfalster:
Oh I see, need to fix the scaffolder too. I'll do that.

PR #262, opened 2020-09-15
Canopy

rafaqz:
This PR refactors the FF16_Environment and creates a modular `Canopy` object that can be used in multiple environments. 

Mostly it handles integration over canopy height.

Currently the constructor appears to be is slightly broken - size is not zero when it should be.


dfalster:
Since this is labelled WIP I'm assuming I shouldn't merge yet. Also don't want to merge until all tests are passing.

rafaqz:
Yeah it wont pass tests yet, its just here so you know were things are at. I'll try to get those last few tests passing tomorrow or later in the week.

rafaqz:
I think this has introduced a major slowdown in the SCM tests. The build is stalling, not failing

rafaqz:
Ok so this is passing now after fix the travis yml so it doesn't break the wait limit.

From benchmarking SCM looks only 25% slower. So the build may have just been near the wait limit anyway, and got bumped over it by the 25% increase. It would be good to have more extensive bencmarks in the scripts.

rafaqz:
Oops no it's still broken. I'm struggling with the way testthat works - I'm used to each test set being run separately and its results displayed as it runs - so you can at least see where it gets up to. This is just a black box that says "the tests take too long" with no details. 

I'm still guessing it's just a memory use problem. Benchmarking memory use of SCM was actually lower in this version than develop - but I'll try that in some other ways to see if I can replicate it.

rafaqz:
So using `devtools::test()`  both gives us nice output printed in travis, and also fixes the problem on linux! It doesn't take 40mins either, so there was something strange happening with the timeout.

https://travis-ci.com/github/rafaqz/plant/jobs/397257043#L1628

But it also reveals some weird architecture dependent bugs:
https://travis-ci.com/github/rafaqz/plant/jobs/397257044#L1099

This will also be interesting to fix.

dfalster:
> But it also reveals some weird architecture dependent bugs: https://travis-ci.com/github/rafaqz/plant/jobs/397257044#L1099

I'm still getting a lot of failures here too. The first is the same as in the above, the others seem to be different.

image


I can give you ssh access to iMac if you want to investigate?

aornugent:
Ruh-roh, maybe it's compiler differences? This branch tests fine on my Windows machine:

```
==> Rcpp::compileAttributes()

* Updated R/RcppExports.R

==> devtools::test()

Loading plant
Testing plant
v |  OK F W S | Context
v |   6       | Adaptive interpolator
v |   8       | Build_schedule [9.0 s]
v |  84       | CohortSchedule [0.2 s]
v |  26       | Cohort-FF16
v |  26       | Cohort-FF16r
v |  24       | Cohort-K93
v |   3       | Control
v |  12       | Disturbance
v |  17       | Environment-FF16
v |  17       | Environment-FF16r
v |  17       | Environment-K93
v |   5     1 | fitness support
------------------------------------------------------------------------------
test-fitness-support.R:19: skip: positive_2d
Reason: plant.ml cannot be loaded
------------------------------------------------------------------------------
v |   7       | Gradient
v | 136       | IndividualRunner [1.4 s]
v |  42       | Individual utilities [0.3 s]
v |  24       | Individual-FF16
v |  24       | Individual-FF16r
v |  21       | Individual-K93
v |  20       | Internals
v |   9       | Interpolator
v |  37       | Lorenz (basic ODE) [0.1 s]
v |  87       | Modular [0.1 s]
v |   3       | OdeControl
v | 117       | Parameters [0.2 s]
v |  20       | Patch-FF16
v |  20       | Patch-FF16r
v |  20       | Patch-K93
v |  70       | QAG [0.1 s]
v |  28       | QK
v |  18       | SCM support [20.3 s]
v | 168       | SCM [110.9 s]
v |  59       | Species-FF16
v |  59       | Species-FF16r
v |  59       | Species-K93
v |  42       | StochasticPatchRunner [12.7 s]
v |  60       | StochasticPatch
v | 174       | StochasticSpecies [0.2 s]
v |  16       | Reference Comparison-FF16 [0.1 s]
v |  37       | Strategy-FF16 [17.0 s]
v |  31       | Strategy-FF16r [0.2 s]
v |  11       | Strategy-K93
v |   9       | Support
v |   3       | Tools
v |   8       | Trapezium integration
v |   3       | Uniroot
v |   7       | utils

== Results ===================================================================
Duration: 174.1 s

OK:       1694
Failed:   0
Warnings: 0
Skipped:  1
```

aornugent:
I think this is saying that the `macOS` environment defaults to `clang` unless specified:
https://docs.travis-ci.com/user/languages/cpp/#choosing-compilers-to-test-against

dfalster:
Golly, how frustrating.  But even if you aren't getting the fail on linux & windows, the routines have slowed down way below anything that could be consider reasonable for such a change. 

Daniel (Mac):
```
✔ |   8       | Build_schedule [2.5 s]
✔ | 136       | IndividualRunner [1.1 s]
✔ |  18       | SCM support [13.6 s]
✔ | 168       | SCM [30.3 s]
✔ |  42       | StochasticPatchRunner [1.3 s]
✔ |  37       | Strategy-FF16 [4.6 s]

```

Andrew (windows):
```
v |   8       | Build_schedule [9.0 s]
v | 136       | IndividualRunner [1.4 s]
v |  18       | SCM support [20.3 s]
v | 168       | SCM [110.9 s]
v |  42       | StochasticPatchRunner [12.7 s]
v |  37       | Strategy-FF16 [17.0 s]
```

Travis (linux)
```
✔ |   8       | Build_schedule [12.8 s]
✔ | 136       | IndividualRunner [1.8 s]
✔ |  18       | SCM support [22.5 s]
✔ | 165       | SCM [176.0 s]
✔ |  42       | StochasticPatchRunner [26.4 s]
✔ |  43       | Strategy-FF16 [25.1 s]
```

rafaqz:
That's pretty hard to debug if gcc is fine with things, but clang isn't... I guess we should set the compiler for ci to always be gcc.  It seems insane to me that R uses different compilers on different systems, usually we aim to remove all sources of variation from scientific tool chains. You can manually swap the compiler to gcc: https://stackoverflow.com/questions/52342071/r-not-recognizing-my-alternative-gcc-compiler-after-config-updates.

@dfalster what is that mac benchmark if you cant get it compiled? just the devel branch?

Anyway, I'm playing with valgrind and some things to try and work out what is causing the slowdown.

rafaqz:
Actually I'm getting no slowdown on `canopy` vs `develop`. If a slowdown has occurred it was probably before this.

aornugent:
Compiled `develop` to make sure it wasn't my little laptop slowing things down and have numbers much closer to Daniels:
```
v |   8       | Build_schedule [3.8 s]
v | 136       | IndividualRunner [1.1 s]
v |  18       | SCM support [17.1 s]
v | 168       | SCM [51.9 s]
v |  42       | StochasticPatchRunner [1.8 s]
v |  37       | Strategy-FF16 [8.3 s]
```

rafaqz:
Oh no... I can't replicate that at all. I have essentially identical numbers on develop and canopy.

I've been going back through the commits looking at the test timings, and SCM and patch/stochasticpatch have really blown out while most other thing have held constant. But we have also added more models - SCM has roughly 3x the tests it used to, but is taking about 10x as long.

rafaqz:
@aornugent is your develop branch up to date?

aornugent:
> @aornugent is your develop branch up to date?
> 
> Also are both those tests with the cable plugged in?

Yes++ 

What a nuisance!

rafaqz:
This could be another compiler difference - windows/linux. I don't know. 

Something is triggering a performance loss or even test failures in some contexts but not others, with either change of system of minor code changes. uggh

aornugent:
I tried adding additional compilers to Travis to zero in on what each compiler is doing, but got stuck and gave up. Convincing R to switch might've been the missing piece: https://thecoatlessprofessor.com/programming/r/selecting-an-alternative-compiler-for-r-package-testing-on-travis-ci/#misc-note

aornugent:
I also checked the test time on 4c8a242, before new models were introduced. Fewer tests for SCM but similar run-time.

```
v | 112       | SCM [48.4 s]
```

rafaqz:
Ok so gcc is a bit lenient with errors, maybe more so than clang. I'm going to get rid of all the compiler warnings.

It would be easier to see if 90% weren't suggesting to remove unnecessary brackets... but priorities right?

dfalster:
Thanks for investigating further.

In answer to your question, i compared latest develop vs canopy. Same number of models, only difference are changes in this PR. I can’t get a time for canopy on mac sorry (too many errors), so times are confounded with system differences.

Yes, those bracket warnings which we can’tm fix are annoying!

dfalster:
I can organise ssh access to a mac if that helps. I’ll courirer one to you if needed.

rafaqz:
Do you all have custom set CXXFlags? -O2 or -O3 ?

rafaqz:
ssh or courier either/or. Just my work schedule is er... variable so it would have to be on all the time for a while.

aornugent:
> Do you all have custom set CXXFlags? -O2 or -O3 ?

I don't think so. [`gcc -v` ](https://github.com/traitecoevo/plant/pull/262#issuecomment-709792198) doesn't have anything exotic ( and `pkgbuild::compiler_flags` from R looks like it's only specifying `pedantic`

rafaqz:
I mean in your .R/Makevars or whatever config file. R uses system level settings to compile packages, so there is another source of variability between us. Whether the defaults are set by R or by our package managers, who knows.

aornugent:
Ah!
```
CXX14 = g++ -m$(WIN) -std=c++1y
CXX14FLAGS=-O3 -march=native
CXX11FLAGS=-O3 -march=native
CXXFLAGS=-O3 -Wno-unused-variable -Wno-unused-function
```

rafaqz:
Yep yours is -O3, mine is -O2. So this change is bumping some -O3 level optimisation back down to what I get with -O2 anyway.

That's good. I can set -O3 and debug it.

dfalster:
I have 
```
CC=clang
CXX=clang++ -arch x86_64 -ftemplate-depth-256
CXXFLAGS=-O3 -mtune=native -march=native -Wno-unused-variable -Wno-unused-function  
```

rafaqz:
Now both develop and canopy are equally faster for me:

```
✔ |   8       | Build_schedule [2.9 s]
✔ | 136       | IndividualRunner [1.4 s]
✔ |  18       | SCM support [17.4 s]
✔ | 165       | SCM [40.0 s]
✔ |  42       | StochasticPatchRunner [3.1 s]


✔ |   8       | Build_schedule [2.9 s]
✔ | 136       | IndividualRunner [1.3 s]
✔ |  18       | SCM support [17.8 s]
✔ | 168       | SCM [35.7 s]
✔ |  42       | StochasticPatchRunner [1.9 s]
```

Just a few seconds longer in SCM. and StochasticPatchRunner. There's something to debug but it's not really the drastic change you saw. This may be that I have a more recent gcc version being on bleeding-edge arch linux - and it's not losing the optimisation in the same way your versions are. I don't know. I'll just generally try to make things more efficient and see what happens. I haven't spent much time doing that yet.

This makes me wonder what optimization level travis has for R by default - I bet that its O2 not O3 - that would explain the hang.

dfalster:
Now we are [shaving yaks](https://medium.com/@firehoseproject/a-guide-to-yak-shaving-your-code-d30f98dc759)!

rafaqz:
I think the yak was shaved weeks ago lol.

But personally I understand the whole situation a lot better from this thread. Performance of plant is dependent on the compiler vendor and its version, local system settings, and small code changes that interact with some of those things in some contexts. 

The performance dropping from a change doesn't mean a huge amount necessarily - we could be chasing particular compiler optimisations that drop in and out with context. So it's better to just work on generally optimising things. More to the point I can't solve this current problem because I don't have the context where it's actually a problem.

We also know how to fix travis permanently so that whole saga doesn't happen again. It needs a .R/Makevars with -O3 in it.

rafaqz:
OK travis looks good now :)

The only error seems to be no latex available on mac. Dan do you wan tot try the latest push on your system?

dfalster:
Yay, all tests passing here too. Good find. 

So for posterity sake, the crucial issue was likely the lack of explicit return in below?
```
double get_environment_at_height(double height) const {
     canopy.get_canopy_at_height(height);
   }
```
2nd line should be
```
 return   canopy.get_canopy_at_height(height);
```

While the tests are passing, the new version is still noticeably slower in a few spots. In the the tests, 

- StochasticPatchRunner has increased from 1.3 to 6.9s
- SCM from 25s -> 45s

could this be something to do with creating or saving environment objects ?

aornugent:
`canopy` is slower at my end too, but probably just machine differences at this point (n = 3):
```
//develop
v |   8       | Build_schedule [3.8 s]
v | 136       | IndividualRunner [1.1 s]
v |  18       | SCM support [17.1 s]
v | 168       | SCM [51.9 s]
v |  42       | StochasticPatchRunner [1.8 s]
v |  37       | Strategy-FF16 [8.3 s]

// canopy
v |   8       | Build_schedule [10.5 s]
v | 136       | IndividualRunner [1.9 s]
v |  18       | SCM support [22.7 s]
v | 165       | SCM [141.6 s]
v |  42       | StochasticPatchRunner [19.2 s]
v |  37       | Strategy-FF16 [17.9 s]
```

rafaqz:
I explained this above.

There is a much smaller reduction with gcc 10. So most likely the additional object complexity we have from adding Canopy means one more level of indirection so the compiler is giving up on some optimisation somewhere that it could do before - because nothing is actually different it's just reorganised. 

But it's happening moreso in your older gcc than the latest version.

Possible some manual inlining and const declaration can get the improvement in gcc 8 that gcc 10 is mostly doing on its own.
I'll have to bump back my gcc version and see what I can do.

rafaqz:
Also there was this:
https://github.com/traitecoevo/plant/pull/262/commits/1bbdc5674eff45deeb315e15e12708d96c347557

rafaqz:
@aornugent it's much more likely to be compiler differences

rafaqz:
So one question is - spend time finding optimising things, or everyone install gcc 10 and not worry about it:
https://community.arm.com/developer/tools-software/tools/b/tools-software-ides-blog/posts/gcc-10-better-and-faster

aornugent:
I was thinking about this last night- I like speed, but the Rtools
toolchain is a really common way for non-dev users to run C++/Rcpp
code. It's maintained and distributed by a single postdoc and has only
recently been updated.

Perhaps we could configure Travis to ensure backward compatibility and I
could write up instructions for users that want faster models?

dfalster:
> I explained this above.
Indeed, but I didn't fully fully, so thanks for the extra explanation. I see what you mean now.

> So one question is - spend time finding and optimising problems, or everyone install gcc 10 and not worry about it
I'd prefer to try and optimise. Requiring people to use a non-default compiler is a major hassle and unlikely to happen for most users. Not even sure if it's possible for Mac users without resorting to homebrew and other finicky setups.

And as much as I love the improved structure that comes with the Canopy class, I don't want a solution that comes with such a big speed hit. 

> Possible some manual inlining and const declaration can get the improvement in gcc 8 that gcc 10 is mostly doing on its own.
That would be great if you could.

Also, it's not surprising there are significant speed costs here - the operations involving building and using the canopy are bar far the most expensive when running a patch.

dfalster:
Just checked and https://github.com/traitecoevo/plant/pull/262/commits/0be3b904ebb3f982f404576978f1428e1f501005 did not change speed issue (comparing develop f5ddd7b to latest canopy 321f671) -- both with clean and fresh build

dfalster:
I have some updates on this to post, coming in an hour or so

dfalster:
Great work here @rafaqz, I'm *almost* ready to merge this. 

After further investigation i can confirm that

* there is no loss in speed for FF16 or FF16r between latest branch on develop (3df74ec) and this PR (on my macs). 
* As you suggested, FF16r is a lot slower than FF16, but this was already the case before this PR. The single line change in FF16r compared to FF16 results in a big speed drop. Replacing that code (see [here](https://github.com/traitecoevo/plant/blob/3df74ec0224c57d97ba0a1f2ec1cb7d478e04a5b/src/ff16r_strategy.cpp#L310)) with the equivalent from FF16 recovers all the speed lost 
* There is still a big speed loss between this PR and develop, but that is entirely caused by the K93 strategy. 

Before committing I'd like to take a closer look at the K93 strategy & environment to see if there's any reason why it is running so much slower.

I'm also adding some benchmarking to the develop branch before merging, so we can document any time changes.

aornugent:
Can confirm that the numerical test on #272 passes on `canopy`, albeit much slower, so I think it's good to merge.

I've a copy of develop on https://github.com/traitecoevo/plant/tree/k39-vignette for posterity until resolved. Interested to see what the problem is!

P.S. I love the massive reduction console garbage during compilation!

PR #267, opened 2020-09-25
Inherit from FF16 in FF16r

rafaqz:
This compiles fine and seems to work. 

It's now only a few lines different between FF16r and FF16 - except the constructor in the .cpp file is all repeated. 

We can fix this by moving all field defaults to the header - I think in C++11 you can do xfield = 1 in the definition. Otherwise we can use `xfield(1)` constructors for FF16, and use the regular constructor for FF16r to update that one variable that is different.

Tests are currently failing because all the tests just append "_Env" to the end of the strategy name to get the environment. We'll have to make that a bit smarter if we decide this is the way to go.


src/ff16r_strategy.cpp
```diff
@@ -91,216 +91,9 @@ FF16r_Strategy::FF16r_Strategy() {
   collect_all_auxillary = false;
   // build the string state/aux name to index map
   refresh_indices();
-  name = "FF16r";
+  name = "FF16";
```
dfalster, line 94: Is this change correct? Should name still be FF16r? Not sure we actually need or use the name anyway. I think i added before @rafaqz outlined another approach to detecting strategy type.

dfalster: Otherwise, love seeing all the duplicated code being deleted here.

rafaqz: Oops. Yeah I search and replaced them all so the diff was cleaner to delete all that code, just forgot to swap that one back.

dfalster:
I get lots of tests failing locally, especially in test-individual-runner (similarly for #262). Want me to do look at anything here?

rafaqz:
If you pull it again (fresh as it's been rebased) there are only a few tests failing now - that typo you found was a bug :). 

Some numbers aren't matching, I'm not sure if this is a testing error or actually something wrong with the constructors/methods not overriding properly.

dfalster:
Weird, I'm still getting lots of failures here, similar to before. All similar issue I suspect. I tried doing make clean first, just to be sure.

image


image


As as I can tell, I've pulled the latest successfully:

image

rafaqz:
Weird. I'm not getting those. Maybe if we both do a full clean and recompile and check back

dfalster:
OK, but note both this and canopy branch are all failing on travis too.

dfalster:
Fresh clone and clean build, same thing. But inlace travis, it doesn't stall. Rather terminates within a very short time because of multiple test failures.

rafaqz:
Very weird. Mine only has 3 failures. in test-patch, test-stochastic-patch and test-stochastic-species.
That's after `make clean RcppR6 compile install test`. All for the value of `ode_rates`.

The other failures seem to be a memory use issue - travis hangs, SCM tests take a long time locally. This will be interesting to debug...

dfalster:
> That's after `make clean RcppR6 compile install test`. 

I tried that exact sequence now to be sure, same results as I report above. Darn!

rafaqz:
It would be nice if we could see which tests break in travis, my feeling is that it's failing _after_ those failures you are getting - the SCM tests are the most likely thing to make travis hang like that.

I think we can set it up to print all the test results to the travis log, currently it's just passed/didn't pass. Successes don't show in the logs either:
https://travis-ci.org/github/traitecoevo/plant/jobs/728248270#L578

dfalster:
All the errors I can see here involve a warning `"Maximum number of subdivisions reached`", which is coming from the integration being called by assimilation. You can trigger it simply by calling `p$compute_rates`

e.g.

```
s <- FF16_Strategy()
p <- Individual("FF16", "FF16_Env")(s)
env <- fixed_environment("FF16", 1)
p$compute_rates(env)
```

So something to do with assimilation not accessing environment correctly, perhaps?

dfalster:
I looked some more into this. The adaptive integration of photosynthesis just keeps refining more and more. It is doing this because it has NAN for a value. So this suggests something isn't working, when passing through the environment into the assimilation and into the integrator.

I think it's caused around here, https://github.com/traitecoevo/plant/blob/f5ddd7bd7724ddbc1c77c684d292841dab168a33/inst/include/plant/models/assimilation.h#L51

If at the above point I insert the following lines

```

    std::cout << "start" <<  " "<<0.5*height <<" "<< 
    environment.get_environment_at_height(0) << " "
    << compute_assimilation_h(0.5*height, height, environment)
    << " " << f(0.5*height) <<std::endl; 
    ```

and then run this, I get
```
s <- FF16_Strategy()
p <- Individual("FF16", "FF16_Env")(s)
env <- fixed_environment("FF16", 1)
p$compute_rates(env)
> start 0.172097 1 nan nan
```

So the `compute_assimilation_h` function is failing, so the function `f` is failing, so the integrator is failing. This may also be why scm is taking a long time and failing. 

Intriguingly though, the environment can be accessed from within assimilate. So it's something about passing it through to the next level in compute_assimilation_h. 

Recall, we were doing all of this before, so it's a change introduced with the canopy

**Comment by dfalster** [comment]
Further poking suggests the calls to `environment.get_environment_at_height(z)` within `compute_assimilation_h`, i.e. [here ](https://github.com/traitecoevo/plant/blob/f5ddd7bd7724ddbc1c77c684d292841dab168a33/inst/include/plant/models/assimilation.h#L76), are behaving weirdly. Could it be something to do with [templating for the assimilation object](https://github.com/traitecoevo/plant/blob/f5ddd7bd7724ddbc1c77c684d292841dab168a33/inst/include/plant/models/assimilation.h#L12)?

**Comment by rafaqz** [comment]
It's weird that this can work or fail with the same code - I don't have a way of investigating it as I can't duplicate it at all.

**Comment by rafaqz** [comment]
This is not actually passing it's just not reporting failures properly now.

**Comment by rafaqz** [comment]
Theses tests are failing but I'm not sure what these differences mean between FF16 and FF16r

https://github.com/traitecoevo/plant/blob/develop/tests/testthat/test-stochastic-patch.R#L53-L54
```R
      expect_true(all(patch$ode_rates[-3] > 0.0))
      expect_equal(patch$ode_rates[[3]], 0)
```
What is -3 and 3 for ? we should make these a variable so the intention is clear.

aornugent:
I've just added a new FF16 variant and discovered I need to add `virtual` tags to the underlying FF16 header as well as the overrides. @rafaqz - is there any problem with making all public methods in `ff16_strategy.h` virtual?

rafaqz:
Yeah, it's an issue. `virtual` is slower as it's a run-time method lookup to find the method for the inheriting type. If you use it _everywhere_ you might see some real performance losses. In a few places it shouldn't matter so much. We could benchmark it and see, but it's not the usual practise.

See:
https://stackoverflow.com/questions/4543537/in-c-should-i-almost-always-use-virtual-inheritance

But yes you may have to keep going back to FF16 to add `virtual` when you need them. C++ just isn't really designed well for this kind of modularity, what we are trying to do is hard and clunky however we write it. Either we duplicate hundreds of lines of code, or mess around with `virtual` everywhere, paying the extra cost of object construction and dynamic method lookup.

(There is language that handles this perfectly without `virtual`, and calls methods for inheriting objects without dynamic method lookup or double construction _even when they are defined in user scripts outside the package_. but I wont mention the name here...)

PR #271, opened 2020-10-20
Benchmarking

dfalster:


dfalster:
@rafaqz or @aornugent Can you cast quick eyes on this before i merge?

Tests only fail because of warning, which has been fixed on Canopy branch.

dfalster:
This PR asks for speed benchmarks to be printed as part of travis routines. This would enable (for example) any future changes in speed to be identified at the time or retrospectively. 

image

https://travis-ci.org/github/traitecoevo/plant/jobs/737288556#L1707

tests/testthat/test-scm.R
```diff
@@ -1,9 +1,10 @@
-context("SCM")
+context("SCM-FF16")
 
-strategy_types <- get_list_of_strategy_types()
+strategy_types <- get_list_of_strategy_types()["FF16"]
```
aornugent, line 3: Does this mean we can ditch the K93 case below?

aornugent:
Neat! With the test specialisation, would we expect to have one context per model (e.g. `context('SCM-K93`)? Maybe this could be added to the scaffolder too.

aornugent (approved):

rafaqz:
Seems like you have based travis on an earlier PR of mine. The latest Canopy is different resolves those problems

.travis.yml
```diff
@@ -39,6 +39,7 @@ script: |
   R CMD check plant*tar.gz --no-tests --no-manual 
   Rscript -e 'devtools::install();devtools::test()'
   pass=$?
+  Rscript -e 'plant::run_plant_benchmarks()'
```
rafaqz, line 42: See the latest Canopy PR this code is all different - this version was a failed experiment not the final version.

Maybe we should merge Canopy first or you are going to have to merge and change all of this again later

tests/testthat/test-scm.R
rafaqz: Why not test them all with separate context labels as in Canopy?

.travis.yml
dfalster: Yeah, i discovered that after sending this. Have subsequently added benchmarking onto your PR and reconciled these changes. see https://github.com/rafaqz/plant/tree/strategy_inheritance

So this branch won’t actually get merged, we’ll take the version on your strategy_inheritance branch

tests/testthat/test-scm.R
dfalster: We could. I decided to restrict to FF16 so that tests run quickly. 

Can add FF16r and K93 back once we figure out why they’re running so much slower. 

They’re still run as part of the benchmarking.

dfalster:
Closed via aa6a0a809c60b55982a04d68820e5a8f2cd3f31f

PR #272, opened 2020-10-23
Add K39 numerical checks for 1sp and 3sp

aornugent:
Not sure why it diffed the whole file, but added numerical check for seed rain and ODE times.

### DIFF
--- a/tests/testthat/test-strategy-k93.R
+++ b/tests/testthat/test-strategy-k93.R
@@ -45,7 +45,6 @@ test_that("Reference comparison", {
   h0 <- 10
   p$set_state("height", h0)
 
-
   expect_identical(p$state("height"), h0)
 
   ## Check: Is this redundant now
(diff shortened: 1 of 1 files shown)


tests/testthat/test-strategy-k93.R
```diff
@@ -1,79 +1,103 @@
...
+  p0 <- scm_base_parameters("K93")
+  p0$k_I <- 1e-6
+  p0$disturbance_mean_interval <- 200
+  
+  # Use single sp. defaults
+  p1 <- expand_parameters(trait_matrix(0.059, "b_0"), p0, mutant = FALSE)
+  p1$seed_rain <- 20
+  
+  out <- run_scm(p1)
+  expect_equal(out$seed_rains, 0.11391, tolerance = 1e-5)
```
dfalster, line 83: I get a fail here:

```
test-strategy-k93.R:83: failure: K93 seed rain is unchanged
out$seed_rains not equal to 0.11391.
1/1 mismatches
[1] 0.114 - 0.114 == 3.18e-05
```
Change tolerance?

tests/testthat/test-strategy-k93.R
```diff
@@ -1,79 +1,103 @@
...
+                       0.044, 0.044, 0.044), 
+                     c("b_0", "b_1", "b_2",
+                       "c_0", "c_1", "d_0", "d_1"))
+  
+  p2 <- expand_parameters(sp, p0, mutant = FALSE)
+  p2$seed_rain <- c(20, 20, 20)
+  out <- run_scm(p2)
+  
+  expect_equal(out$seed_rains, c(0.00030, 0.34744, 0.01303), tolerance = 1e-5)
+  expect_equal(length(out$ode_times), 1017)
```
dfalster, line 102: I get a fail here too.

Maybe asking too much for this to be equal across arhcitectures

```
test-strategy-k93.R:102: failure: K93 seed rain is unchanged
length(out$ode_times) not equal to 1017.
1/1 mismatches
[1] 992 - 1017 == -25
```

tests/testthat/test-strategy-k93.R
```diff
@@ -1,79 +1,103 @@
...
+  expect_equal(ret, trait_matrix(numeric(0), "b_0"))
+})
+
+
+test_that("K93 seed rain is unchanged", {
+  
+  # Generic parameters
+  p0 <- scm_base_parameters("K93")
+  p0$k_I <- 1e-6
+  p0$disturbance_mean_interval <- 200
```
dfalster, line 76: The long disturbance regime (200 years) causes this to run very slowly - 240s. Ideally tests would take no longer than 10s. Until we identify the cause of the slowdown, could we  change to an example with interval of 10 years, so that tests run quickly?

Perhaps leave the old one in there but commented out?

The example won't match the one from paper, but that's ok. As long as we can detect if the output of the model changes from what it is currently, that will be sufficient.

dfalster (changes requested):
> Not sure why it diffed the whole file, but added numerical check for seed rain and ODE times.

Change of line endings perhaps?

Thanks for this!!! Good to know numerical results haven't changed with introduction of Canopy. See requested changes.

dfalster:
Apparently travis checks failed, but no really. New tests passed so you can ignore that.

tests/testthat/test-strategy-k93.R
aornugent: Done

tests/testthat/test-strategy-k93.R
aornugent: This one is harder for me to verify. Do you mind checking whether there's a major discrepancy with the reduced disturbance frequency, please?

tests/testthat/test-strategy-k93.R
aornugent: Good point! I was lazy and just re-used the model parameters, but it's muuuch faster now.

tests/testthat/test-strategy-k93.R
aornugent: Otherwise could fall back to just testing a couple of specific ODE times, like the one species case above.

PR #280, opened 2021-02-19
Remove plant.ml dependency

### LINKED ISSUE #269: Failed to install plant from R 4.0
remotes::install_github("traitecoevo/plant", dependencies=TRUE) fails to install under R 4.0 with the following message:
Downloading GitHub repo traitecoevo/plant@HEAD
Skipping 1 packages not available: plant.ml
Installing 5 packages: tinytex, rmarkdown, plant.ml, numDeriv, deSolve
Installing packages into ‘/home/timaeus/R/x86_64-pc-linux-gnu-library/4.0’
(as ‘lib’ is unspecified)
Error: Failed to install 'plant' from GitHub:
  (converted from warning) package ‘plant.ml’ is not available (for R version 4.0.2)

This continues even after issuing the following command: 
Sys.setenv(R_REMOTES_NO_ERRORS_FROM_WARNINGS="false")

Sys.getenv("R_REMOTES_NO_ERRORS_FROM_WARNINGS")
[1] "false"

I was able to get around this by manually downloading the code (your Option 2 for installation)

aornugent:
Fixes #269

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -33,7 +33,6 @@ Suggests:
     parallel,
     RcppR6 (>= 0.2.3),
     knitr,
-    plant.ml,
     rmarkdown,
     bench
 VignetteBuilder: knitr
--- a/tests/testthat/test-strategy-k93.R
+++ b/tests/testthat/test-strategy-k93.R
@@ -4,7 +4,7 @@ context("Strategy-K93")
 
 test_that("Defaults", {
   expected <- list(
-   height_0 = 2.0,                   
+   height_0 = 2.0,
    b_0 = 0.059,
    b_1 = 0.012,
    b_2 = 0.00041,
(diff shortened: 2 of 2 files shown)


aornugent:
Weird, the seed rain check on K93 failed. So nothing broken, but something unstable in the model results.

dfalster:
I note that plant.ml is only in suggests, so as it is it's not required to install. Should we leave as this optional extra?

dfalster:
We could also revisit what it does and where the best place to put it is.

That said, happy to remove it for now and reinsert at a later date.

dfalster:
Re test failure: It looks like the number of ODE steps changed. We experienced this before -- on different systems small precision differences lead to different numbers of ODE steps. Perhaps this test should be disabled? I seem to recall  we disabled other strategies for the same reason.

aornugent:
I didn't investigate plant.ml beyond it's role in #269. Maybe `dependencies=TRUE` is overriding the optional Suggests.

Let's take it out for now and I'll file an issue to revisit.

dfalster:
ok

PR #290, opened 2021-04-16
Move disturbance from environment to patch

aornugent:
This PR decouples environment and disturbance, to later allow patches to be initialised with a pre-determined size distribution.

It's almost finished, but several (important) tests fail because the mean disturbance interval is not correctly set when creating the patch. I think this currently occurs in the environment constructors of each strategy?


aornugent:
All tests now passing (locally at least!)

dfalster (approved):

PR #293, opened 2021-04-21
Move k_I into strategy, better value for k93

### LINKED ISSUE #287: k93 model failing -- need to set seed rain
@kunstler reports that in v2.0.0, the k93 strategy fails to run:
```
library(plant)
p0 <- scm_base_parameters("K93")
p1 <- expand_parameters(trait_matrix(0.0825, "b_0"), p = p0, mutant = FALSE)
data1 <- run_scm_collect(p1)
> Error in SCM___K93__K93_Env__run_next(self) :
  Environmental interpolator has gone out of bounds, try lowering the extinction coefficient k_I
```

We've encountered this problem before: `k_I` needs to be set to a lower value. This was in the test but that i commented out https://github.com/traitecoevo/plant/blob/develop/tests/testthat/test-strategy-k93.R#L75

So we should update:
- vignettes (https://github.com/traitecoevo/plant/blob/develop/inst/docs/strategy.Rmd#L497) (though, perhaps not needed if doing next step
- reimplement basic test at https://github.com/traitecoevo/plant/blob/develop/tests/testthat/test-strategy-k93.R#L75
- can we put a better value in the k93 environment constructor, so that this works without needing to manually setting a better value? https://github.com/traitecoevo/plant/blob/develop/inst/include/plant/models/k93_environment.h

Thanks @kunstler for noticing this!

dfalster:
Shifts the k_I parameter into the plant strategy

Fixes #287 #229


src/ff16_environment.cpp
```diff
@@ -2,8 +2,4 @@
 
 namespace plant {
 
-const double get_k_I(const FF16_Environment environment) {
```
aornugent, line 5: Can we delete this file if it's empty?

aornugent (approved):
This works great.

Issue #206, opened 2019-07-09
Some renaming of functions, files, classes


dfalster:
Gradually thinking some renaming of functions, files, classes will help make the material more accessible. For example:

Variables and functions:

- [ ] `height` -> `size`
- [ ] `plant` -> `individual`
- [ ] `compute_vars_phys` --> `compute_vital_rates`
- [ ] `scm` -> `solve_characteristic_method`
- [ ] `area_leaf_above` -> `competition` (or other)
- [ ] `germination_probability` -> `establishment_probability`

files: 

- [ ] `plant.h` -> `individuals.h`
- [ ] `ff16strategy` -> `strategy_ff16`
- [ ]  Move supporting tools (like qag, qk, ) into subfolder?

aornugent:
It strikes me that there is significant duplication for simple environments. `K93Environment` and `FF16Environment` are effectively the same.

Brainstorming, with no consideratoin for how this will affect compilation, there might be room for components that are not specific to one strategy but part of a larger class of models.

For example:
```
inst
├──include
|   ├── evolutionary
|   |   ├── StochasticPatch.h
|   |   └── StochasticSpecies.h
|   ├── fish
|   |   └── 3DEnvironment.h
|   ├── generic
|   |   ├── Cohort.h
|   |   ├── Environment.h
|   |   ├── Individual.h
|   |   ├── Patch.h
|   |   ├── Species
|   |   └── Strategy.h
|   ├── models
|   |   ├── FF16.h
|   |   ├── FF16evo.h
|   |   ├── FF16r.h
|   |   └── K93.h
|   ├── numeric
|   |   ├── QAG.h
|   |   ├── Lorenz.h
|   |   ├── QK.h
|   |   |    .....
|   ├── plant
|   |   ├── LightEnvironment.h
|   |   └── WaterEnvironment.h
|   └── SCM.h
└── RcppR6_classes.yml
```
would permit `FF16` and `K93` to use the same `plant\LightEnvironment`, while still providing room to specialise.

aornugent:
- [x] Update names of utility functions, e.g. `grow_plant_to_size` -> `grow_individual_to_size`
- [ ] Change name of result in utility functions e.g. `res$plant` -> `res$individual`

aornugent:
- [ ]  change `lai_error` in `run_scm_error` to e.g. `competition_error`

dfalster:
Note @devmitch - I added something to the list, rename `cohort` to `node`

[...]

PR #297, opened 2021-04-28
Renaming of seed_rain(s)

aornugent:
Seed rain as a term was overloaded, meaning both arrival and production of seed. Given #206 we also took the opportunity to generalise these terms.

Changes include variants of:
* `seed_rain_[in]` -> `birth_rate`
* `seed_rain_[out]` -> `net_reproductive_ratio` and `offspring_production`
* `add_seeds` -> `introduce_new_cohort`
* `seed_survival_weighted` -> `offspring_produced_survival_weighted`

That last one is still up for grabs, but it should be easy to find and replace all if we think of a better name.


inst/include/plant/environment.h
```diff
@@ -42,20 +42,20 @@ class Environment {
 
   void clear_environment() {}
 
-  double seed_rain_dt() const {
-    if (seed_rain.empty()) {
-      Rcpp::stop("Cannot get seed rain for empty environment");
+  double offspring_arriving_dt() const {
```
dfalster, line 45: Why are we using  `offspring_arriving_dt` here instead of `birth_rate_dt`? The new term may be confusing. 

Also, are we moving the broth_rate stuff out of environment? Or is that in next step, after successful renaming

dfalster (changes requested):
Looks very promising, great work. One query in the code.

We'd better update the vignettes too.

dfalster (approved):

PR #299, opened 2021-05-03
Fixed time patch

aornugent:
Originally, runtime was determined by mean_disturbance_interval, such that almost no patches remained in the meta-population.

This PR exposes the max_runtime as a control parameter and effectively decouples runtime from the disturbance regime.


aornugent:
Blegh, rogue commits. Will tidy and try again.

PR #301, opened 2021-05-03
Refactor metapopulation disturbances

aornugent:
This PR refactors the existing `Disturbance` class into a base `Disturbance_Regime` class from which derived classes `No_Disturbance_Regime` and `Weibull_Disturbance_Regime` inherit from, in order to permit both meta-population and.. non-metapopulation runs.


inst/include/plant/disturbance_regime.h
```diff
@@ -0,0 +1,19 @@
...
+#define PLANT_PLANT_DISTURBANCE_REGIME_H_
+
+#include <vector>
+
+namespace plant {
+
+class Disturbance_Regime {
+public:
+  virtual double density(double time) const {return 0.0;};
+  virtual double pr_survival(double time) const {return 0.0;};
```
aornugent, line 12: @dfalster - this fixes the previous compilation problem, but means that `no_disturbance.h` below is pretty trivial.

Options I can see are either:
1. treating 'no disturbance' as the base case (as you suggested)
1. returning `NA_REAL` as the default

I don't mind either really, other than conceptual clarity of a base class with 'no disturbance' as a special case which side-steps the weirdness of 'Weibull disturbance' inheriting from 'no disturbance'.

dfalster:
Tests failed :(. Was recent commit meant to work, or just a step on the way?

dfalster:
Everything compiles on my machine, but many tests failing.

aornugent:
Thanks - I was stashing changes before little bro kicked me off his gaming PC 😅 

Will resolve tests and request merge later this week. I wonder if we can set Github Actions to ignore draft PRs?

dfalster:
Should be possible. https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request

aornugent:
Almost there - some of the schedule building tests are failing after the change to `max_patch_lifetime` that I haven't quite figured out.

For some reason, it sets a custom schedule correctly, but when we pull out the `SCM$Parameters` object and use it to create a second `SCM` it falls back to the default schedule rather than using the custom one.

aornugent:
All tests passing 🤞 

I ended up refactoring some of the schedule building code to something simpler, but please check I haven't broken some intended design feature. Also, `stochastic.R` has a `patch_density` variable that I've taken out as I can't see how it was supposed to be derived.

dfalster (approved):
Great work Andrew!

PR #302, opened 2021-05-23
Remove k_I from parameters

aornugent:
The light extinction coefficient (k_I) has now been moved into strategies. These are the last few breadcrumbs and can now be removed.


dfalster (approved):
Good catch

PR #305, opened 2021-08-02
Enable state variables in Environment

aornugent:
I'd like to merge this work onto the `develop` branch, in order to make another PR that separates Environment and Control objects so that we can better configure soil state (these commits are also on #306, but aren't directly related).

@itowers1 - do you mind if your notebooks also get pushed on to the develop branch? We can continue using the `water` branch after the PR is merged, it'll just tidy up the commit history.


dfalster (approved):
Looks great, well done. One requested change - can we delete the pdf before merging in?

I'd also like to rename `Water_strategy` to something more consistent with the other work, maybe `FF16w`, but appreciate this may not be the right time with two PRs in motion simultaneously. So happy to flag that for future if you prefer

aornugent:
Thanks @dfalster  - will file an issue to rename `Water -> FF16w`.

PR #313, opened 2021-08-30
Decouple assimilation

aornugent:
I took this one step further and moved the integrator into assimilator which is initialised per strategy anyway. This sets up a future where Control objects can be decoupled from strategies.

Let's check the benchmarks and verify there are no slowdowns. We can roll back a single commit to keep the integrator as part of FF16_Strategy.

### DIFF
--- a/inst/include/plant/control.h
+++ b/inst/include/plant/control.h
@@ -26,7 +26,6 @@ namespace plant {
 
 struct Control {
   Control();
-  void initialize();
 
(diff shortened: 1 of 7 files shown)


aornugent:
The results look similar to [the previous run](https://github.com/traitecoevo/plant/runs/3276575161?check_suite_focus=true) so this looks ok to me.

dfalster (approved):
Awesome! Great idea to embed in Assimilator

PR #314, opened 2021-09-10
Refactor control

aornugent:
One step closer to #306.

This removes the Control object from Parameters and passes it directly to SCM or Patch APIs as required. This is handled on the R side with what I hope are sensible defaults, so most often users won't notice it's even there.

```
result <- run_scm_collect(p = expand_parameters(p0, trait_matrix), 
                          control = scm_base_control())
```

All tests pass, but I didn't test the carrying capacity or equilibrium scripts as I'm not totally sure what the correct output should be. Please let me know if there's anything that looks like it needs further investigation.

Next step will be repeating this exercise with Environment.

### DIFF
--- a/R/build_schedule.R
+++ b/R/build_schedule.R
@@ -8,23 +8,24 @@
 ##'
 ##' @title Build Cohort Schedule
 ##' @param p Parameters object
+##' @param ctrl Control object
 ##' @return A Parameters object, with schedule components set.  The
 ##' output offspring produced is also available as an attribute
 ##' \code{birth_rate}.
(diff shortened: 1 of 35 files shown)


dfalster:
This looks great, well done @aornugent 

> I didn't test the carrying capacity or equilibrium scripts as I'm not totally sure what the correct output should be. Please let me know if there's anything that looks like it needs further investigation.

That's OK, I can review those. But that won't happen today. So feel free to merge, move on with environment, I can get back to you about that soon. Please assign me a GH issue

PR #323, opened 2021-11-08
Exposing auxiliary variables

devmitch:
This PR exposes the auxiliary variables in `FF16_environment.h` to the R API, and includes a very basic R test. Still need to expose the variables guarded by `collect_all_auxiliary`, ie `area_sapwood`.


dfalster (approved):
Nice! I can see it was bit tricky. I have run it locally and can see it works. Well done!

aornugent (approved):

PR #304, opened 2021-06-11
Initialise patch state

aornugent:
Early progress for solving patches with arbitrary initial conditions. The included test shows how this has been added to `run_scm_collect` which may or may not the desired API design, but useful for now. It's also hard coded for the single species case which will need to be updated/

Next steps will be adapting the default cohort introduction schedule to something smarter and then thinking about how to use `build_schedule` to see if we can arrive at the same solution as starting a patch from scratch.

PR #324, opened 2021-11-15
FF16 Rainfall API

devmitch:
This PR begins to implement the rainfall model in the FF16 environment. Currently only a few necessary getters, setters and spline computation functions are exposed, we can add/remove depending on what will be required for the model. 

A basic test is included, and some slight refactoring was done to the `Interpolator` class.

This may need to wait for #304 to be merged.


aornugent:
Nice one @devmitch! I think the next step is to show that it accumulates correctly - i.e. accessing the spline from `environment.compute_rates()` and checking that the total input rainfall for the duration of a simulation is equivalent to integrating the area under the spline. The code on [L57](https://github.com/traitecoevo/plant/blob/develop/inst/include/plant/models/ff16_environment.h#L57) is a bit nonsense but is testing the same thing. 

We now want to replace `soil_infiltration_rate` with `rainfall_spline` and update the [FF16w tests](https://github.com/traitecoevo/plant/blob/develop/tests/testthat/test-strategy-ff16w.R#L76). I think `FF16_Environment` holds a `time` variable but you might like to check the call stack and double check how `compute_rates` is called.

dfalster:
Hi @devmitch 

great work. I'm really happy to see this coming together. Well done setting it up and extending in a way that keeps all tests passing (and adds new tests)

I agree with Andrew's suggested next steps.

devmitch:
I've set a default spline of y = 0 to stay in line with the default soil infiltration rate of 0.0 in `FF16w_make_environment` . I used domain [0, 10] to construct it, I hope the spline isn't volatile when you try extrapolate points outside the domain like x = 1000, might investigate that later.
Now the rainfall spline will be set in either of 3 ways:
Through the default argument (y=0) of FF16w_make_environment
Through passing in customs arguments of x, y points to FF16w_make_environment
Doing either of the above and after construction, setting the spline through rainfall_init(x, y)

src/interpolator.cpp
```diff
@@ -21,6 +21,9 @@ void Interpolator::init(const std::vector<double>& x_,
 // Compute the interpolated function from the points contained in 'x'
 // and 'y'.
 void Interpolator::initialise() {
+  if (not std::is_sorted(x.begin(), x.end())) {
+    util::stop("spline control points must be in non-descending order");
```
dfalster, line 25: non-descending -> ascending?

Also need to be unique? 

"spline control points must be unique and in ascending order"

R/ff16w.R
```diff
@@ -64,16 +64,17 @@ FF16w_StochasticPatchRunner <- function(p) {
...
+##' @param rainfall_x list of x (time) positions for rainfall spline
+##' @param rainfall_y list of y (rainfall) positions for rainfall spline
 FF16w_make_environment <- function(canopy_light_tol = 1e-4, 
                                    canopy_light_nbase = 17,
                                    canopy_light_max_depth = 16, 
                                    canopy_rescale_usually = TRUE,
                                    soil_number_of_depths = 1,
                                    soil_initial_state = 0.0,
-                                   soil_infiltration_rate = 0.0) {
+                                   rainfall_x = seq(0, 9, 1),
```
dfalster, line 76: You're right that need x & y. I would rather pass in a single data frame than an array, so perhaps change this accordingly?

src/interpolator.cpp
```diff
@@ -84,9 +101,10 @@ std::vector<double> Interpolator::r_eval(std::vector<double> u) const {
   check_active();
   const size_t n = u.size();
   std::vector<double> ret(n);
-  for (size_t i = 0; i < n; ++i) {
-    ret[i] = eval(u[i]);
-  }
+  // for (size_t i = 0; i < n; ++i) {
+  //   ret[i] = operator()(u[i]);
+  // }
+  std::transform(u.begin(), u.end(), ret.begin(), [&](double x){return operator()(x);});
```
dfalster, line 107: Fancy!

tests/testthat/test-environment.R
```diff
@@ -41,3 +41,35 @@ for (x in names(strategy_types)) {
...
+  
+  ## interpolated range of points
+  expect_equal(env$rainfall_eval_range(c(-7, 1, 7.8345)), c(49, 1, 61.37939025), tolerance=1e-2)
+  
+  ## clear points, set new points and recompute spline
+  env$rainfall_clear_points()
+  x <- seq(-10, 10, 1)
+  y <- 2*x
+  env$rainfall_add_points(x, y)
+  env$rainfall_recompute()
```
dfalster, line 72: I wonder if recompute should be a separate step, or instead happen automatically at the end of `rainfall_add_points`? I can't imagine a use case where you added points in several steps.

tests/testthat/test-environment.R
```diff
@@ -41,3 +41,35 @@ for (x in names(strategy_types)) {
...
+  x <- seq(-10, 10, 2)
+  y <- x^2
+  env$rainfall_init(x, y)
+  
+  ## discrete points used for spline creation
+  expect_equal(env$rainfall_eval(2), 4)
+  expect_equal(env$rainfall_eval(-2), 4)
+  
+  ## interpolated points
+  expect_equal(env$rainfall_eval(3), 9, tolerance=1e-2)
```
dfalster, line 59: tolerance of 0.01 is very generous. I would have expected it to be accurate at much higher resolution. Does it not pass with default tolerance ?

Similar everywhere tolerance is used.

tests/testthat/test-interpolator.R
```diff
@@ -26,6 +26,7 @@ test_that("Splines require sensible data", {
   s <- Interpolator()
   expect_error(s$init(c(1, 2, 3), 1), "Incorrect length input")
   expect_error(s$init(numeric(0), numeric(0)), "insufficient number of points")
+  expect_error(s$init(c(3, 2, 1), c(1, 1, 1)), "spline control points must be in non-descending order")
```
dfalster, line 29: will need to change to follow suggested change in comment above

tests/testthat/test-strategy-ff16w.R
```diff
@@ -17,19 +17,23 @@ test_that("FF16w Environment", {
...
-  e$set_soil_infiltration_rate(10)
+  x <- seq(0, 9, 1)
+  y <- rep(1, 10)
+  e$rainfall_init(x, y)
   expect_equal(e$soil$states, 0)
   e$compute_rates()
-  expect_equal(e$soil$rates, 10)
+  expect_equal(e$soil$rates, 1)
   
+  # not needed anymore? should we still be able to set the states manually?
```
dfalster, line 27: probably. We will occasionally want to specify an exact environment for a non-dynamic use

tests/testthat/test-strategy-ff16w.R
```diff
@@ -52,31 +56,43 @@ test_that("FF16w Environment", {
...
+               sapply(seq(1, 10), function(x) { (out$time^2)/x }),
                tolerance = .01)
   
+  # check the states are correct
+  # model is currently 1/z * int(0, t)(f(x)), where z is depth, 
+  # x is time, f is spline, t is number of years
   expect_equal(out$patch$environment$soil$states,
-               c(105, 52, 35, 26, 21, 17, 15, 13, 11, 10),
+               sapply(seq(1, 10), function(i) { (1/i) * integrate(integrand, 0, out$time)$value }),
                tolerance = 0.1)
```
dfalster, line 94: again with tolerance

tests/testthat/test-strategy-ff16w.R
```diff
@@ -52,31 +56,43 @@ test_that("FF16w Environment", {
...
   expect_equal(out$patch$environment$soil$rates, 
-               c(1.00, 0.50, 0.33, 0.25, 0.20, 0.17, 0.14, 0.13, 0.11, 0.10),
+               sapply(seq(1, 10), function(x) { (out$time^2)/x }),
                tolerance = .01)
   
+  # check the states are correct
+  # model is currently 1/z * int(0, t)(f(x)), where z is depth, 
+  # x is time, f is spline, t is number of years
   expect_equal(out$patch$environment$soil$states,
-               c(105, 52, 35, 26, 21, 17, 15, 13, 11, 10),
```
dfalster, line 77: So this is checking that rainfall accumulates correctly over time?

dfalster (commented):
Really great progress @devmitch .

Some minor suggestions included in this review. @aornugent may have more to mention

src/interpolator.cpp
devmitch: Yeah that's correct, the x points are required to be strictly ascending and therefore unique.

https://github.com/traitecoevo/plant/blob/a8f8a7b26059a7e82718bda5e85b5cfaa518b9ec/src/tk_spline.cpp#L151

src/interpolator.cpp
devmitch: Thanks! I still don't like that I have to explicitly call `operator()` for the faster version of the evaluation. Might take another look at these functions later.

tests/testthat/test-environment.R
devmitch: Yeah I agree - I was trying to model it after `Canopy.h` and `rescale_canopy()`. I'll probably integrate it into add_points and add_point (might not even need add_point if add_points can take in a single point in each array or in one data frame)

R/ff16w.R
devmitch: So single data frame for `FF15w_make_environment` which is split into two arrays to pass into `rainfall_add_points()`?

tests/testthat/test-environment.R
devmitch: On that line specifically I was able to pass with `0.001 (1e-3)` tolerance at most - however, if I create the spline with more control points by using a smaller step value:
```R
x <- seq(-10, 10, 0.41) # as opposed to step of 2
y <- x^2
env$rainfall_init(x, y)
```
then the tests pass with a much smaller tolerance of `1e-7` or `1e-6`, so I think its a limitation of the spline when given few control points. Ill change the tests to use more control points.

tests/testthat/test-strategy-ff16w.R
devmitch: Was able to pass this with tolerance `1e-5` by using step value of 0.1:
```
x <- seq(0, 110, 0.1)
```
This is obviously going to slow the tests down a little bit since the spline will be created with 1100 control points, but the numbers for this test are quite large so accuracy would unsurprisingly suffer. Trying to use tolerance `1e-6` yields:
```
10/10 mismatches (average diff: 1)
[1] 389415 - 389414 == 1
[2] 194708 - 194707 == 1
[3] 129806 - 129805 == 1
[4]  97354 -  97353 == 1
[5]  77884 -  77883 == 1
[6]  64903 -  64902 == 1
[7]  55632 -  55631 == 1
[8]  48678 -  48677 == 1
[9]  43269 -  43268 == 1
```
Maybe I could try with a lower order polynomial than quadratic...

tests/testthat/test-strategy-ff16w.R
devmitch: Yes - its probably best I just hardcode the values. This line:
```function(i) { (1/i) * integrate(integrand, 0, out$time)$value }```
Integrates the spline at each soil depth and checks the result against the numerical integration from `run_scm`.

**Comment by devmitch** [comment]
After discussions with @dfalster and @itowers1, the rainfall spline (which now exists in the `Environment` class) has been replaced with a set of splines called **extrinsic drivers**. 

Construction of the rainfall spline is now done in either of 3 ways:
```R
# default rainfall spline of y = 1
env <- make_environment("FF16w")

# rainfall spline of y = 5
env <- make_environment("FF16w", rainfall=5)

# custom rainfall spline based on set of control points
env <- make_environment("FF16w")
x <- seq(0, 10, 1)
y <- x^2
env$set_extrinsic_driver("rainfall", x, y) # overwrites previous spline of y = 1
```

Querying the spline now requires the user to pass in the name of the extrinsic driver they wish to query:
```R
# determine the rainfall value at time 84
res <- env$extrinsic_driver_evaluate("rainfall", 84)
# determine the rainfall values over the range of integer times from 0 to 100
res <- env$extrinsic_driver_evaluate_range("rainfall", seq(0, 100, 1))
```

The list of potentially active extrinsic_drivers is now also available from any environment:
```
env <- make_environment("FF16w")
expect_equal(env$get_extrinsic_driver_names(), c("rainfall"))
```

Two more things that need to be looked at:
1. One test is still failing in `test-patch.R`:  
```R  
# context: env is an FF16w environment  
env_size <- env$ode_size  
env_state <- patch$ode_state  
env_rates <- patch$ode_rates  
expect_equal(patch$ode_size, env_size)  
  
# either 0 or numeric(0)  
expect_identical(patch$ode_state, numeric(env_size))  
expect_identical(patch$ode_rates, numeric(env_size)) # FAILS, is actually 1  
```  
I'm not exactly sure what the relationship between `ode_rates` and `env_size` is, and I'm not sure how to resolve this. This was previously passing as `ode_rates` was 0 because initially there was no rainfall / no soil infiltration.

2. The rainfall spline is being constructed in the `FF16_Environment` constructor, though I'm not sure it should be however since rainfall is an `FF16w` specific feature. However as `compute_rates` needs the rainfall spline, `FF16_Environment` needs to have it as well. Bit confused here...

Thoughts @dfalster @aornugent ?

aornugent:
Tackling the failing test first!

```diff
# context: env is an FF16w environment  
env_size <- env$ode_size  
env_state <- patch$ode_state  
env_rates <- patch$ode_rates  

expect_equal(patch$ode_size, env_size)  

# these two are typos that previously passed because we initialised at 0
-expect_identical(patch$ode_state, numeric(env_size))  
+expect_identical(patch$ode_state, numeric(env_state)) 
-expect_identical(patch$ode_rates, numeric(env_size)) 
+expect_identical(patch$ode_rates, numeric(env_rates))
```

aornugent:
This is awesome work @devmitch. I really like the approach of storing multiple drivers in a map- it makes it much more general and hopefully we can apply the same pattern to a variety of new model processes.

To your second point:

> The rainfall spline is being constructed in the FF16_Environment constructor, though I'm not sure it should be 

I wasn't able to find a way to inherit from `FF16_Environment` and extend it with custom methods, like rainfall, while also inheriting from `FF16_Strategy` as we do in `FF16w_Strategy` due to templating. In the end, we decided that it was ok to extend `FF16_Environment` as long as it maintained backwards compatibility with the `FF16_Strategy` - that is, unused methods were ok as long as we were still able to reproduce the original results published by Dan.

In this case, the uninitialised driver fails loudly, which I think is a good thing:

```
Error in FF16_Environment__extrinsic_driver_evaluate(self, driver_name,  : 
  Interpolator not initialised -- cannot evaluate 
```

My only outstanding question is what happens when a spline has been initialised for a shorter duration than the model simulation? I've done a little tinkering and the spline will continue to interpolate outside the bounds of which it has been initialised, but often with strange behaviour:

```r
env$set_extrinsic_driver("rainfall", x = c(1, 2, 3), y = c(10, 1, 0))
env$extrinsic_driver_evaluate_range('rainfall', c(1, 2, 3, 10, 50, 100, 1000))
> 10   1   0   7  47  97 997
```

I suggest then that we stick with the philosophy of 'fail loudly' and have a simulation stop if it goes beyond the length of the spline. 

This might require modifying `interpolator.h` to store `min_x` and `max_x`, then adding a check in `environment::extrinsic_driver_evaluate_*`. I think (but should confirm) that we don't want to do this in `interpolator::evaluate_*` because we sometimes do extrapolate outside of the range of control points in `canopy.h`.

I'd also suggest running this check in the Patch constructor which has access to `parameters.max_patch_lifetime` and the environment, which provides an opportunity to add additional context to the error message.

Let me know if you'd like to discuss!

devmitch:
> I wasn't able to find a way to inherit from `FF16_Environment` and extend it with custom methods, like rainfall, while also inheriting from `FF16_Strategy` as we do in `FF16w_Strategy` due to templating. In the end, we decided that it was ok to extend `FF16_Environment` as long as it maintained backwards compatibility with the `FF16_Strategy` - that is, unused methods were ok as long as we were still able to reproduce the original results published by Dan.

Ah I see, makes sense. I think I was confusing myself a little as well when I was considering this.

> My only outstanding question is what happens when a spline has been initialised for a shorter duration than the model simulation? I've done a little tinkering and the spline will continue to interpolate outside the bounds of which it has been initialised, but often with strange behaviour

Yeah it seems to be smart enough to extrapolate correctly for simple curves like the linear y = k spline constructed for `FF16w_make_environment()` which is constructed with points `(0, k), (1, k), ... , (10, k)` yet correctly places `(1000, k)` on the spline. However for curves like your example with low control points and no real pattern it seems to be volatile. 

> This might require modifying `interpolator.h` to store `min_x` and `max_x`, then adding a check in `environment::extrinsic_driver_evaluate_*`. I think (but should confirm) that we don't want to do this in `interpolator::evaluate_*` because we sometimes do extrapolate outside of the range of control points in `canopy.h`.

Another solution that comes to me is adding a boolean flag in `Interpolator` called `extrapolate`, which is false by default. The logic for `interpolator::eval()` would look like:

```cpp
double Interpolator::eval(double u) const {
  check_active();
  if (not extrapolate and (u < min() or u > max())) {
    util::stop("Extrapolation disabled and evaluation point of " + u + " outside of interpolated domain.");
  }
  return tk_spline(u);
}
```

It then seems ideal to also provide a faster, less safe method for spline evaluation as well (this is how it is done in the STL with maps for example - .at() checks the key exists before getting the value, operator[] does not)
```cpp
double interpolator::operator()(double u) const {
  return tk_spline(u);
}
```

However, if we were to disable extrapolation for rainfall splines, how do we know what domain to construct the default spline of y = k in `FF16w_make_environment` with if we don't have access to `parameters.max_patch_lifetime`? Should `max_lifetime` also be an argument in `FF16w_make_environment`?

aornugent:
> Another solution that comes to me is adding a boolean flag in Interpolator called extrapolate, which is false by default and can be changed with setExtrapolation() for example.

Sounds very tidy! Could we then enable extrapolation in the default constructor, but reset the flag to `FALSE` when a custom spline is created using `set_extrinsic_driver`?

>  Should max_lifetime also be an argument in FF16w_make_environment?

Not if we can avoid it. It'd be nice to re-use the same environment for a variety of simulations. My suggestion to run the extrapolation check in `Patch` was only really to make the code clear that it should test for compatibility between environment and parameters at the start of a run, if `extrapolate = FALSE` then it should fail loudly regardless.

I see the tests are failing!

```
...
40: test_check("plant")
An irrecoverable exception occurred. R is aborting now ...
Segmentation fault (core dumped)
```

My intuition is it doesn't like calling things that haven't been initialised on Windows and Ubuntu (strange that it passes on OSX). All tests passed locally on my Ubuntu machine the other day, which is a bit curly. Maybe try `make rebuild` and push the Rcpp and NAMESPACE files? 🤷‍♂️

dfalster:
Good work @devmitch . I agree with @aornugent 's suggestions above.

devmitch:
> Not if we can avoid it. It'd be nice to re-use the same environment for a variety of simulations. My suggestion to run the extrapolation check in Patch was only really to make the code clear that it should test for compatibility between environment and parameters at the start of a run, if extrapolate = FALSE then it should fail loudly regardless.

The following is currently how the default spline of y = k is generated in `FF16w_make_environment`:
```cpp
FF16w_make_environment <- function(canopy_light_tol = 1e-4, 
                                   canopy_light_nbase = 17,
                                   canopy_light_max_depth = 16, 
                                   canopy_rescale_usually = TRUE,
                                   soil_number_of_depths = 1,
                                   soil_initial_state = 0.0,
                                   rainfall = 1) {
  # ...
  x <- seq(0, 10, 1)
  y <- rep(rainfall, 11)
  e$set_extrinsic_driver("rainfall", x, y)
}
```
If we disable extrapolation for the rainfall spline, then evaluating it at any point > 10 will throw an error as extrapolation is disabled and we are attempting to evaluate the spline above the maximum control point. Since the normal patch lifetime is up to 105 years, this obviously won't work well when we try to evaluate these larger numbers. Since I'm guessing patch lifetime is arbitrary, how can we construct the spline with a large enough domain of control points to account for any arbitrary patch lifetime?

aornugent:
Ah ha- if we're handling construction on the R side:

```diff
FF16w_make_environment <- function(canopy_light_tol = 1e-4, 
                                   canopy_light_nbase = 17,
                                   canopy_light_max_depth = 16, 
                                   canopy_rescale_usually = TRUE,
                                   soil_number_of_depths = 1,
                                   soil_initial_state = 0.0,
                                   rainfall = 1) {
  # ...
  x <- seq(0, 10, 1)
  y <- rep(rainfall, 11)
- e$set_extrinsic_driver("rainfall", x, y)
+ e$set_extrinsic_driver("rainfall", x, y, extrapolate = TRUE)
}
```

Users can then update the environment with more complex rainfall scenarios:
```r
env <- FF16w_make_environment(rainfall = 1)
x <- seq(0, 105, len = 100)
y <- x^2

# extrapolate = FALSE by default
env$set_extrinsic_driver("rainfall, x, y)
```

Then, if the patch lifetime exceeds the length of the control points, the simulation will fail (hopefully with a useful error). This is particularly useful where the extrinisic driver comes from real data- we want to prevent the case where people start analysing results that go beyond their domain of interest. If they want to forecast (i.e. extrapolate beyond the available data) they can make an informed decision to set `extrapolate = TRUE` or create synthetic data that represents a rainfall scenario of their choice. 
____

>  Maybe try make rebuild and push the Rcpp and NAMESPACE files?
> Git tells me there are no changes to these files after running this command...

Darn, that's more concerning then. I'll try to check out these changes tonight and see what might be happening, please let me know if you find anything that might address the problem. My hunch is still that the uninitialised splines will be causing trouble in FF16 and K93, but my hunches are nearly always wrong so more methodical investigation is needed.

devmitch:
> Darn, that's more concerning then. I'll try to check out these changes tonight and see what might be happening, please let me know if you find anything that might address the problem. My hunch is still that the uninitialised splines will be causing trouble in FF16 and K93, but my hunches are nearly always wrong so more methodical investigation is needed.

I wasn't able to figure out much, google tells me that it might have something to do with the R packages installed, not sure.

Anyway, it seems like Rcpp can't handle default arguments, so I've gone with the following for the default linear splin in `FF16w_make_environment`:
```cpp
x <- seq(0, 10, 1)
y <- rep(rainfall, 11)
e$set_extrinsic_driver("rainfall", x, y)
e$extrinsic_driver_extrapolation("rainfall", TRUE);
```
(unless there is another way you know of to get them working...)

PR #327, opened 2021-12-13
Rainfall api final tests

devmitch:
Just some tests that I forgot to include in the previous PR that check the interpolator correctly fails when extrapolation is disabled and an attempt to evaluate the interpolated function outside the domain of points that were used to create it is made.


aornugent (approved):
Suggest "setExtrapolate" -> "set_extrapolation"

dfalster:
Wonderful, great work @devmitch @aornugent ! Very excited to have this feature implemented. 

The recent commits in this PR were from @devmitch, working off my iMac. Just assigning credit where credit is due.

PR #328, opened 2021-12-13
FF16w drawdown through evapotranspiration

aornugent:
This attempts to provide some measure of cohort impact on shared, extractable resources that are represented as state variables in the environment. It's clunky ~and doesn't yet work~ - I can already see a few ways to improve the general approach, but posting it here for review in case there is a more streamlined approach that absolves the need for clunky-ness all together.

In short; we expose a strategy specific consumption method that is called from the patch during compute rates to provide the integrated impact of all cohorts for a given species across all resource pools (e.g. soil layers). The consumption totals are then passed into environment the rate of change is updated to reflect the impact of all species on a given resource pool.


aornugent:
A smattering of print statements led me to the observation that `water_exfiltration` is negative, but I'm not sure why. 

```r
  p0 <- scm_base_parameters("FF16w")
  env <- make_environment("FF16w")
  ctrl <- scm_base_control()
  
  out <- run_scm(p1, env, ctrl)
  out$patch$environment$soil$states
```

devmitch:
`water_exfiltration` was negative as the order of cohort heights being added to the vector of x-coordinates for trapezium integration was descending - trapezium integration requires ascending order:
```cpp
double trapezium(const ContainerX& x, const ContainerY& y) {
// ...
// x1 is ahead of x0, so if descending order ie x0 > x1, then x1 - x0 < 0 => negative result
while (x1 != x.end()) {
  tot += (*x1++ - *x0++) * (*y1++ + *y0++);
}
return tot * 0.5;
}
```
Adding them in ascending order yields a positive result:
```cpp
for(auto it = cohorts.rbegin(); it != cohorts.rend(); ++it) {
  heights.push_back(it->height());
  consumption.push_back(it->compute_consumption(i));
}
```

inst/include/plant/models/ff16_environment.h
```diff
@@ -55,9 +54,23 @@ class FF16_Environment : public Environment {
...
 
-  virtual void compute_rates() {
+  // currently only used by FF16w but present in both FF16 and K93
+  // avoids uninitialised variables/splines because vars.state_size = 0
+  // perhaps not completely user safe, but I think diffcult to break from R API
+  virtual void compute_rates(std::vector<std::vector<double>> const& resource_depletion) {
+    // treat each soil layer as a separate resource pool
     for (size_t i = 0; i < vars.state_size; i++) {
-      vars.set_rate(i, extrinsic_driver_evaluate("rainfall", time) / (i+1));
+      double water_infiltration = extrinsic_drivers["rainfall"].eval(time) / (i+1);
```
dfalster, line 63: I think we can remove the `/ (i+1)` term here.

aornugent: Agree - it's a leftover test of soil depths, but will be replaced with drainage in time.

inst/include/plant/species.h
```diff
@@ -192,6 +195,45 @@ std::vector<double> Species<T,E>::net_reproduction_ratio_by_cohort() const {
...
 }
 
+template <typename T, typename E>
+std::vector<double> Species<T,E>::integrate_shared_resource_consumption(const E& environment) const {
+  int n_resources = environment.vars.state_size;
+  std::vector<double> ret;
+  ret.reserve(n_resources);
+
+  for(size_t i = 0; i < n_resources; i++){
+    std::vector<double> heights;
```
aornugent, line 205: It's inefficient to get cohort heights within the this loop because they're the constant for all resources (soil layers). I'll make a suggestion elsewhere how we can remove the loop over `n_resources` altogether.

inst/include/plant/species.h
```diff
@@ -192,6 +195,45 @@ std::vector<double> Species<T,E>::net_reproduction_ratio_by_cohort() const {
...
+    consumption.reserve(size());
+    // cohort heights are in descending order - we need ascending for integration
+    for(auto it = cohorts.rbegin(); it != cohorts.rend(); ++it) {
+      heights.push_back(it->height());
+      consumption.push_back(it->compute_consumption(i));
+    }
+
+    // integrate over cohorts; use sum when only a few cohorts
+    if(size() < 3) {
+      ret.push_back(accumulate(consumption.begin(), consumption.end(), 0));
```
aornugent, line 218: @dfalster - this is an obvious crutch. Is there a better approach to integration with few points?

inst/include/plant/species.h
```diff
@@ -192,6 +195,45 @@ std::vector<double> Species<T,E>::net_reproduction_ratio_by_cohort() const {
...
+      ret.push_back(util::trapezium(heights, consumption));
+    }
+  }
+  return ret;
+}
+
+// Temporary - refactor into something more generic
+template <typename T, typename E>
+std::vector<double> Species<T,E>::auxilliary_variable_by_cohort(std::string aux_name) const {
+  std::vector<double> ret;
```
aornugent, line 229: I don't think this is used any more but could be a replacement pattern for other `variable_by_cohort` methods.

src/ff16w_strategy.cpp
```diff
@@ -31,6 +31,58 @@ double FF16w_Strategy::net_mass_production_dt(const FF16_Environment& environmen
   return net_mass_production_dt_A(assimilation_, respiration_, turnover_);
 }
 
+double FF16w_Strategy::evapotranspiration_dt(double area_leaf) const {
+  return 2 * area_leaf;
```
aornugent, line 35: @itowers1 - this is the interesting part of the model, but is currently doing very little.

inst/include/plant/models/ff16w_strategy.h
```diff
@@ -12,13 +12,39 @@ class FF16w_Strategy: public FF16_Strategy {
...
+    }
+    return ret;
+  }
+
+  double evapotranspiration_dt(double area_leaf) const;
+  double water_access(const FF16_Environment& environment,
+                      double height, double area_leaf_);
+
+  // temporarily treating all soil layers as the same
+  double compute_consumption(int resource_index) {
```
aornugent, line 33: Rather than passing an index, we can pass a pointer to the environment and loop over the number of resources (soil layers), returning a vector for each cohort.

inst/include/plant/patch.h
```diff
@@ -176,12 +176,20 @@ void Patch<T,E>::rescale_environment() {
 
 template <typename T, typename E>
 void Patch<T,E>::compute_rates() {
-  for (size_t i = 0; i < size(); ++i) {
-    double pr_patch_survival = survival_weighting->pr_survival(time());
+  double pr_patch_survival = survival_weighting->pr_survival(time());
+  std::vector<std::vector<double>> resource_depletion; // n_species x n_layers
```
dfalster, line 180: Since size of this object is already known, could create and reserve size on patch creation

aornugent:
Closing in favour of #329

Issue #321, opened 2021-11-08
Interpolate size distributions to specified heights and times


dfalster:

dfalster:
The solver returns size distributions at times and sizes determined by the adaptive solver. The user will likely want these at specific times or heights.

itowers1:
interpolate_to_times yields extremely small negative values for mostly fecundity and in one instance area_heartwood, mass_heartwood. Truncate at zero?

itowers1:
Mortality apparently unsuccessfully interpolated for species 1 with respect to height for all steps. Should have a number of interpolated values for a given height = number of steps included in plot, but seem to have just one at the moment because mortality(step=150, height=c(1,5,10)) is NaN.

dfalster:
Flagged in the repo-wide TODO audit #473 (Tier 3): related to the untested tidy-output interpolation gap in `tests/testthat/test-tidy-outputs.R:83-98`.

dfalster:
Implemented in commit `65da50ab` via the exported functions `interpolate_to_times()` and `interpolate_to_heights()` in `R/tidy_outputs.R`. These take tidied SCM output and spline-interpolate each state variable to user-specified times or heights (out-of-range points returned as `NA`), and are exercised by tests in `tests/testthat/test-tidy-outputs.R`. The core ask of this issue is delivered, so closing.

(The separate question of interpolation accuracy / negative-value truncation raised in the comments is tracked under the test-coverage audit, #482.)

Issue #309, opened 2021-08-09
Expand state variables available in output for FF16


dfalster:
This relates to #296 

The role of previous plant plus object was to give access to more of the internals. We were reluctant to store all of them for sake of speed. The redesign of strategy internals means this can now be accomplished using just a single strategy object, with a switch (`collect_all_auxillary`) to store more or fewer of the internals. 

I have two suggested pathways here. Arguably both would be good additions. 

1. Expand optional auxiliary variables collected  in `FF16_strategy.h` (at https://github.com/traitecoevo/plant/blob/99c443f81b98e1284e14dcda4102a31108ca039c/inst/include/plant/models/ff16_strategy.h#L39 and https://github.com/traitecoevo/plant/blob/99c443f81b98e1284e14dcda4102a31108ca039c/src/ff16_strategy.cpp#L129) to enable collecting more variables. The option `collect_all_auxillary` in the constructor (see  https://github.com/traitecoevo/plant/blob/99c443f81b98e1284e14dcda4102a31108ca039c/src/ff16_strategy.cpp#L10) can then be triggered only when running `run_scm_collect`
2. Expose all the functions in strategy via R to enable more post-hoc calculations. Given height, we can calculate leaf area. Given leaf area, we can calculate root mass, etc.... Currently the way strategy is defined in `RcppR6_classes.yml` doesn't expose all the strategy functions. But it could. (would need  to change from a list-like construct to something more like `Canopy`.

aornugent:
The first step being implemented is to add `ode_aux` to `Patch` so that it can be queried from an `SCM` object.  This will allow us to return the addtional state variables in `run_scm_collect` using a `collect_aux` flag.

Later we will address the second step by re-factoring the RcppR6 definition for `*_Strategy` objects to expose methods as well as attributes by switching from a list class to a constructor class (which isn't super well defined in the Rcpp docs)

dfalster:
This is achieved via commit referenced above, 65da50ab8550436ac32d50817b871903f86d469b

Issue #308, opened 2021-08-09
Tidy patch outputs


dfalster:
The outputs of `run_scm_collect` are not particularly user friendly. 

suggest tidying into tidy data format, organising by cohort and species. 

Here is a prototype 

```{r}

tidy_species <- function(data) {
  
  # get dimensions of data = number of steps * number of cohorts
  dimensions <- dim(data[1,,] )
  
  # establish data structure for results    
  data_species <- 
    tidyr::expand_grid(
      step = seq_len(dimensions[1]), 
      cohort = seq_len(dimensions[2])
    )
  
  # retrieve bnames of all tracked variables
  vars <- data[,1,1] %>% names()
  
  # bind each onto main data frame
  for(v in vars) {
    data_species[[v]] <- 
      results$species[[1]][v, , ] %>% 
      as.data.frame %>% tidyr::as_tibble() %>%
      tidyr::pivot_longer(cols=starts_with("V"), names_to = "cohort") %>% 
      pull(value)
  }
  
  data_species %>% mutate(density = exp(log_density)) %>% select(-log_density)
}




tidy_env <- function(env) {
  tibble(step = seq_len(length(env))) %>%
    left_join(by = "step", 
    env %>% purrr::map_df(as_tibble, .id= "step") %>% mutate(step = as.integer(step))
    )
}





tidy_patch <- function(results) {

  out <- results
  
  data <- tibble(
    step = seq_len(length(results$time)),
    time = results$time, 
    patch_density = results$patch_density
    )
  
  out[["species"]] <- 
    left_join(by = "step", data,
      purrr::map_df(results$species, tidy_species, .id="species")
    )
  
  out[["env"]] <- 
    left_join(by = "step", data,
      tidy_env(results$env)
    )
  
  out[["n_spp"]] <- length(results$species)
  
  out[["patch_density"]] <- NULL
  
  out
}

run_scm_collect() %>% tidy_patch()
```

Could even become the default for scm, or an option in scm

dfalster:
Tidy outputs should enable easier calculation of garages values for output 

In general, we want to calculate 3 types of output

1. each cohort at each heigh at each time (default raw output) 
2. total or average of some variable for each species at each patch age --> obtained by integrating over n at given patch age
3. some variable at common size at each patch age --> obtaining by interpolating along n at given patch age
4. Averages of any of the above over all patch ages (i.e. #89 )

This should be easier with tidy outputs. E.g. for no 2 above

```

[...]

PR #330, opened 2022-01-17
Revive exponentially decaying recruitment (and tidy patch)

aornugent:
Daniel and I tinkered with this model during our investigation of lockup, but it never made it onto Github. A single parameter `reproduction_decay` is introduced that governs the change in establishment probability over time.

We can see that as `recruitment_decay` increases, the establishment of late successional cohorts declines to zero, leaving only the first wave of cohorts having a material density.

**Recruitment decay = 0**

image

**Recruitment decay = 0.2**

image

**Recruitment decay = 5**

image

I haven't caught up with Daniel's fancy `tidy_patch` framework, but here's how to edit this parameter and plot results:

```r
devtools::load_all()

p0 <- scm_base_parameters("FF16r")
p0$strategy_default$recruitment_decay = 0

env <- make_environment("FF16r")
ctrl <- scm_base_control()

out <- run_scm_collect(p1, env, ctrl)

# Relativise the log densities onto (-4, max)
rel <- function(x, xmin = -4) {
  x[x < xmin] <- xmin
  xmax <- max(x, na.rm=TRUE)
  (x - xmin) / (xmax - xmin)
}

t <- out$time
h <- out$species[[1]]["height", , ]

d <- out$species[[1]]["log_density", , ]
rd <- rel(d)


n <- length(t)
x <- matrix(rep(t, ncol(h)), nrow(h))
col <- matrix(make_transparent("black", rd), nrow(d))

plot(NA, xlim=range(t), ylim=range(h, na.rm=TRUE),
     las=1, xlab="Time (years)", ylab="Cohort height (m)")
segments(x[-1, ], h[-1, ], x[-n, ], h[-n, ], col=col[-n, ], lend="butt")

```


aornugent:
Oh my bad - I thought FF16r had fallen out of use, other than as an example of strategy inheritance, but I now see that it's called in the StochasticSpecies test. I can re-enable the custom `fraction_allocation_reproduction` and then either:
1. leave recruitment decay in FF16r and set the default to 0; or
2. move recruitment decay into another strategy, e.g. FF16decay.

Becca-90:
That's awesome!

dfalster:
Hi all, Nice Andrew! 

The above two commits 

1. implement the `recruitment_decay` option on the FF16 model (and thereby all FF16 models), with default value of 0. So set it to something higher (e.g. 5) to get the effect Andrew demonstrated
2. Pull in changes from tidy_patch branch. 


So @Becca-90, you should 

1. install this version
2. Add a setting for recruitment_decay into your scripts, e.g. in the `run_my_patch` function

dfalster:
Well it seems I'm making a hatchet job out of this. Will fix.

dfalster:
Ok, working now ;)

dfalster:
Hi @aornugent -- I've updated all the tidy_patch materials. Is this Ok to merge into develop?

dfalster (approved):
Nice work with recruitment decay. This branch now also includes all changes for tidy_patch.

dfalster:
Addresses #321, #309, #308, #331

vignettes/example.Rmd
```diff
@@ -0,0 +1,276 @@
+---
+title: "Example worklfow"
```
aornugent, line 2: `worklfow` -> `workflow`

R/tidy_plots.R
```diff
@@ -0,0 +1,32 @@
...
+    dplyr::mutate(relative_log_density = rel(.data$log_density)) %>%
+    ggplot2::ggplot(ggplot2::aes(x = .data$time, y = .data$height, color = as.character(.data$species),
+                        group = interaction(.data$cohort, .data$species), alpha = .data$relative_log_density)) +
+    ggplot2::geom_line() + 
+    ggplot2::labs(x = "Time (years)",
+         y = "Height (m)",
+         color = "Species", 
+         alpha = "Relative log(Density [/m/m2])") +
+    ggplot2::scale_alpha(range=c(0.01,1))+ 
+    ggplot2::theme_classic()
```
aornugent, line 29: Suggest `ggplot2::coord_cartesian(expand = F)` to set lower axis limits to (0, 0)

vignettes/example.Rmd
```diff
@@ -0,0 +1,276 @@
...
+  traits =  trait_matrix(c(0.07), c("lma")),
+  B_lf1 = 1,
+  seed_rain = 100,
+  p0 = base_parameters(),
+  find_equilbrium = FALSE,
+  latitude = 28.182
+  ) {
+  
+  hyper_par_fn = make_FF16_hyperpar(B_lf1 = B_lf1, latitude = latitude)
+  p1 <- expand_parameters(traits, p0, hyper_par_fn, mutant = FALSE)
```
aornugent, line 64: I think `p0` should change `base_parameters()`; otherwise we need a line above saving `p0 <- base_parameters()`

vignettes/example.Rmd
```diff
@@ -0,0 +1,276 @@
...
+  p1$seed_rain <- seed_rain
+  
+  if(find_equilbrium)
+    result <- equilibrium_seed_rain(p1)
+  else
+    result <- build_schedule(p1)
```
aornugent, line 70: This suggests we should always call `build_schedule` which might be a good idea, but means we don't get the blindingly fast results of the FF16 defaults. Would this work instead?

```r
run_my_patch <- function(
  ...
  find_equilibrium = FALSE,
  optimise_schedule = FALSE,
  ...
  
  if(find_equilbrium)
    result <- equilibrium_seed_rain(p1)
  else if(optimise_schedule)
    result <- build_schedule(p1)
  else
    result <- p1

  run_scm_collect(result)
}
```

vignettes/example.Rmd
```diff
@@ -0,0 +1,276 @@
...
+  tidy_patch()
+
+results_tidy
+```
+
+By pulling out the 'species' object from our list of tidied result outputs, we can plot the size distibution for each of the species in the patch. In this plot, each line represents the height over time of a characteristic originating at a given height (given by seed mass) at each step. Note that steps in this case are not equivalent to years, where the distance between steps is instead determined by the cohort spacing algorithm. We can take advantage of varying line transparency in ggplot2 to plot the relative log(Density) of each characteristic the size distribution of each species in the pathc. 
+
+```{r}
+results_tidy$species %>%
+  drop_na() %>%
```
aornugent, line 99: `drop_na()` is namepaced in `tidyr`, but we could re-export it if it's integral to the primary workflow.

vignettes/example.Rmd
```diff
@@ -0,0 +1,276 @@
...
+results_tidy <- results %>%
+  tidy_patch()
+
+```
+
+
+Lets interpolate our results from the `run_mypatch` above for timepoints 1,5,10,1000 using the funcion `interpolate_to_times`. Note that time in units of years. 
+
+```{r}
+times <- c(1, 5, 10, 1000)
```
aornugent, line 198: I get `NA`'s for times longer than the simulation, maybe change `1000` -> `100`

vignettes/example.Rmd
aornugent: Derp, I see it now.

aornugent:
Looking good @dfalster ! I love the new framework for visualisation. I've left a few comments on the workflow doc for your review, but hopefully all minor. @Becca-90 - I've updated the test script:

**Recruitment decay = 5**

```r
base_parameters <- function() {
  p0$strategy_default$recruitment_decay = 5
  
  return(p0)
}

run_decay_patch <- function(
  traits =  trait_matrix(c(0.07), c("lma")),
  B_lf1 = 1,
  seed_rain = 100,
  p0 = base_parameters(),
  find_equilbrium = FALSE,
  optimise_schedule = FALSE,
  latitude = 28.182
) {
  
  hyper_par_fn = make_FF16_hyperpar(B_lf1 = B_lf1, latitude = latitude)
  p1 <- expand_parameters(traits, p0, hyper_par_fn, mutant = FALSE)
  p1$seed_rain <- seed_rain
  
  if(find_equilbrium)
    result <- equilibrium_seed_rain(p1)
  else if(optimise_schedule)
    result <- build_schedule(p1)
  else
    result <- p1
  
  # gather outputs at each time step
  run_scm_collect(result)
}

out <- run_decay_patch()

results <- out %>% tidy_patch() %>% expand_state_FF16()
totals <- results$species %>% integrate_over_size_distribution()

totals %>% 
  ggplot(aes(time, individuals, colour=species)) +
  geom_line() +
  coord_cartesian(expand = F) + 
  theme_classic()
```

Which confirms that recruitment of individuals drops after the first wave:

image

vignettes/example.Rmd
dfalster: Thanks. That was intentional, 1000 being outside the range.

vignettes/example.Rmd
dfalster: Good suggestion

vignettes/example.Rmd
dfalster: Let's see. For now it's in `imports` and loaded as a library for the vignette

dfalster:
Thanks @aornugent

PR #332, opened 2022-01-29
Renaming variables

devmitch:
Begins to rename some of the variables/names mentioned in #206, starting with `plant -> individual`. 

I've changed the variables in `Cohort` and `IndividualRunner`, as well as updated the relevant R file so the tests still pass. Not sure whether to stick with the wording "individual" or "indv" as member variables. Still need to update relevant docs.

In some of the R files, there are still lots of mentions of "plant" in place of "individual" - if I change these, anyone using these functions will probably have to update their scripts. This could also include the function names themselves - for example:
https://github.com/traitecoevo/plant/blob/3b6fd9cf4311e2f7850b0d116fa31dea92d84323/R/plant.R#L23-L50
Thoughts?


dfalster:
Great! I'll check in more detail later. But a quick answer-- 

> Not sure whether to stick with the wording "individual" or "indv" as member variables. 
Can you give an example where this is used? 

> In some of the R files, there are still lots of mentions of "plant" in place of "individual" - if I change these, anyone using these functions will probably have to update their scripts. This could also include the function names themselves 

Let's go ahead and change them.

devmitch:
> Can you give an example where this is used?

For example, this seems a bit verbose
https://github.com/traitecoevo/plant/blob/3b6fd9cf4311e2f7850b0d116fa31dea92d84323/inst/include/plant/cohort.h#L58
and can be shortened like I did in `IndividualRunner`
https://github.com/traitecoevo/plant/blob/3b6fd9cf4311e2f7850b0d116fa31dea92d84323/inst/include/plant/plant_runner.h#L35

"indv" is used elsewhere so I wasn't sure:
https://github.com/traitecoevo/plant/blob/5902bebeb81ecba38fd92f0d4785d76ee6eaa401/vignettes/individuals.Rmd#L33-L40

dfalster:
Hi @devmitch 

Ah I see. I prefer verbose in the R and C++ functions, so individual over individual. There's no doubt what we're talking about. 

The example with `indiv` you linked to is when someone is using the package, so then you can use whatever variable name you want. `indiv` is fine in that context.

devmitch:
There was also a comment in Individual.h:
`// TODO recame lcp_whole_plant to competition_compensation_point`
which I did - is this still desired?

dfalster:
> There was also a comment in Individual.h....which I did - is this still desired?

Please rename. Thinking about it now, `resource_compensation_point` is a better name, thank you!

src/plant_tools.cpp
```diff
@@ -4,27 +4,27 @@
 // [[Rcpp::export]]
 plant::Internals FF16_oderunner_plant_internals(
```
dfalster, line 5: FF16_oderunner_plant_internals should become FF16_oderunner_individual_internals

dfalster: And similar below. perhaps search on oderunner_plant_internals

dfalster (approved):
Nice work. One more suggested change in comments.

devmitch:
There are some variables in `Control.h` that seem to still be conceptually tied to the idea of an actual plant:
https://github.com/traitecoevo/plant/blob/5b4bdcb7066db41a5e33f5b2915b57b0c8b3b652/inst/include/plant/control.h#L35-L36
not sure how to handle these.

devmitch:
On the advice of @aornugent I've renamed:
`plant_seed_tol -> offspring_production_tol`
`plant_seed_iterations -> offspring_production_iterations`
`survival_plant -> survival_individual`
`plant_tools -> individual_tools`

Looks like this is ready to merge.

dfalster (approved):
Great work, all looks very good to me. If tests are passing you can merge. I suggest doing a squash merge into develop. better if @devmitch does the merge, as it then gets recorded as his work

PR #335, opened 2022-02-24
More renaming of variables

devmitch:
Continuing to tick off the list of tasks in #206, this PR renames instances of `Cohort` to `Node` in variables, functions and files.

There may be some variables lingering on the R side that still reference a cohort by name, such as "coh" (which I renamed), or simply "c" (which are harder to find).

If these changes are approved, I can also deal with the `height` -> `individual_size` rename on this branch/PR as well.


dfalster (approved):
Excellent work @devmitch!

I have not suggestions or changes.

dfalster:
Great work team!

PR #339, opened 2022-05-09
Move `compute_assimilation` into FF16 Strategy as a lambda

aornugent:
Draft of changes to modularise how assimilation is calculated. Where previously the per-leaf photosynthetic canopy was hard-coded, it now resides in `FF16_Strategy::compute_assimilation` which is passed to `assimilation::assimilate` for integration along the distribution of canopy openness
 during calculation of `net_mass_production_dt`.

I've left the leaf level approximation in `assimilation` for now, which can be accessed from the `assimilator` object -

```c++
// ff16_strategy.cpp
double FF16_Strategy::compute_assimilation(double z, double height,
                                     const FF16_Environment& environment) {
  return assimilator.assimilation_leaf(environment.get_environment_at_height(z)) * assimilator.q(z, height);
}
```

When FF16w comes online, we can substitute in a new method while also reusing `q` to access information about the distribution of leaf area -

```c++
// ff16w_strategy.cpp
double FF16w_Strategy::compute_assimilation(double z, double height,
                                     const FF16_Environment& environment) {
  // something like..
  return leaf::assimilation(environment.get_environment_at_height(z)) * assimilator.q(z, height);
}
```

The only requirements are to return a double and scale the result by `area_leaf`:

```c++
// assimilation.h
double assimilate(std::function<double(double)> f, ...) {
    ...
    return area_leaf * A;
}
```

We can modify this last line if additional scaling terms are needed and set them to 1 in FF16 to maintain backwards compatibility.

### DIFF
--- a/inst/include/plant/models/assimilation.h
+++ b/inst/include/plant/models/assimilation.h
@@ -3,6 +3,7 @@
 #define PLANT_PLANT_ASSIMILATION_H_
 
 #include <memory>
+#include <functional>
 #include <plant/environment.h>
 #include <plant/control.h>
 #include <plant/qag.h> // quadrature::intervals_type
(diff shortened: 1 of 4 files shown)


dfalster (approved):
Thanks for progressing this. 

Looks OK but I'm a bit confused, as I thought the main motivation was to pull the leaf level equation out of the assimilator thing and pass that in as a function?

Anyway, happy for you to progress if you have a plan!

aornugent:
Merged and extended in `simple_water_model` branch.

Issue #344, opened 2022-07-27
Incorrect estimates of integrated density from `integrate_over_size_distribution`


Becca-90:
library(plant)
    library(dplyr)

    ## 
    ## Attaching package: 'dplyr'

    ## The following objects are masked from 'package:stats':
    ## 
    ##     filter, lag

    ## The following objects are masked from 'package:base':
    ## 
    ##     intersect, setdiff, setequal, union

    library(tidyr)
    library(ggplot2)
    library(purrr)

Let’s assume a known density wrt height, to test against. Let

$$n(*H*) = a \\, H ^{−b}$$

Then we can directly integrate to compare results against ‘plant’
integration

$$N= \int\_{1}^{20} n(H) \\, dH = a \bigg\[\frac{H^{1-b}}{1-b}\bigg\]^{20}\_{1}$$
$$H\_{T}= \int\_{1}^{20}N(H) \\, H \\, dH = a \bigg\[\frac{H^{2-b}}{2-b}\bigg\]^{20}\_{1}$$

Now generate data to tests

    ## set density = a*H^-b, height sequence 1:20, b = 3/2, a = starting density per unit area
    ## create stand of individuals with heights obtained and resulting density over time
    ## so we have constant height growth increase, therefore assume time steps are non-linear

    a = 10
    b = 3/2
    plant_stand <- 
      tibble(
        time = 1,
        step=2,
        patch_density = 1, 
        species = 1, 
        height = seq(20,1, by = -0.05)
        ) %>% 
      mutate(
        cohort = seq_len(n()) %>% rev(),
        density = a*height^(-b)
      )

Now compare solutions

    # integrate with plant inbuilt solver
    stand_integrate <- integrate_over_size_distribution(plant_stand)

    # Analyitcal soluntions
    # N: a/(1-b) H ^ (1-b)
    # H: a/(2-b) H ^ (2-b)

    N_analytical <- a/(1-b) *20 ^ (1-b) - a/(1-b) *1 ^ (1-b)
    N_trapezium <- -plant:::trapezium(plant_stand$height, plant_stand$density)
    N_plant<- stand_integrate$individuals

    c(N_analytical, N_trapezium, N_plant)

    ## [1]  15.52786  15.53099 241.21153

    H_analytical <- a/(2-b) *20 ^ (2-b) - a/(2-b) *1 ^ (2-b)
    H_trapezium <- -plant:::trapezium(plant_stand$height, plant_stand$height*plant_stand$density)
[...]

PR #345, opened 2022-07-27
Proposed fix to issue with integration of density

Becca-90:
Issue #344

### DIFF
--- a/R/tidy_outputs.R
+++ b/R/tidy_outputs.R
@@ -192,5 +192,5 @@ integrate_over_size_distribution <- function(tidy_species_data) {
       dplyr::across(where(is.double) & !c(.data$density, .data$density_integrated, .data$min_height) , ~-trapezium(height, density * .x)), 
       .groups = "drop"
     ) %>% 
-    rename(density = density_integrated)
+    dplyr::rename(density = density_integrated)
 }
--- a/R/tidy_outputs.R
+++ b/R/tidy_outputs.R
@@ -187,9 +187,10 @@ integrate_over_size_distribution <- function(tidy_species_data) {
     dplyr::filter(.data$step > 1) %>% 
     dplyr::group_by(.data$step, .data$time, .data$patch_density, .data$species) %>% 
     dplyr::summarise(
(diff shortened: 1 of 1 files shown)


aornugent:
Thanks @Becca-90!

You can see above that your PR failed the automated tests - which isn't a problem because it tells you what to fix! Clicking through one of the `R-CMD-check` actions above and scrolling all the way to the bottom takes me to [this error message](https://github.com/traitecoevo/plant/runs/7539336119?check_suite_focus=true#step:9:289).

The fix is simple: you need to add the `dplyr::` namespace to `rename`.

You can run these tests locally, before pushing your changes with `devtools::test()` or the keyboard shortcut `Ctrl + Shift + T`.

I wasn't able to tell from just reading your code how these changes fixed the issue. Could you please describe your solution?

Becca-90:
Morning!

So the issue was in the last part of the integrate_across_size_distribution function. We previously had individuals calculate like this:

```{r}
    dplyr::summarise(
      individuals = -trapezium(.data$height, .data$density),
      min_height = min(.data$height),
      dplyr::across(where(is.double), ~ -trapezium(height, density*.x)),
      .groups="drop"
    )
```
it was basically causing the trapezium integration to be done twice for individuals and giving some crazy output... so in the second call of trapezium (below) we just had to isolate everything that isn't already integrated to integrate across... I think that's essentially what was going on!
We also renamed individuals as density integrated (a bit clearer?)

```{r}
    dplyr::summarise(
      density_integrated = -plant:::trapezium(.data$height, .data$density), 
      min_height = min(.data$height),
      dplyr::across(where(is.double) & !c(density, density_integrated, min_height) , ~-plant:::trapezium(height, density * .x)), 
        .groups = "drop") %>% 
    rename(density = density_integrated)
```

aornugent (approved):

aornugent:
@dfalster - I'm happy to merge this without a dedicated test, which can be submitted in a second PR. Thoughts?

dfalster:
I'll add a test tomorrow. The template code is already there

PR #355, opened 2022-11-14
Add missing tests for fitness & equilibrium functions

dfalster:
Despite extensive tests, somehow various functions related to fitness and demographic equilibrium went untested. Presumably because they were slow. Lack of tests here enabled loss of functionality as we changed some of the interface. 

This PR adds some tests on master, which will then be ported back to develop and updated as needed.

Here is the list of functions idnetified
# Equilbirum.R

- [x] equilibrium_birth_rate 
- [x] equilibrium_birth_rate_iteration 
- [x] equilibrium_birth_rate_solve 
- [x] equilibrium_birth_rate_solve_robust 
- [x] equilibrium_birth_rate_hybrid 
- [ ] make_equilibrium_runner 
- [ ] equilibrium_runner_cleanup 
- [ ] equilibrium_birth_rate_solve_target

# fitness_support

- [x] bounds 
- [x] check_bounds 
- [x] bounds_infinite 
- [x] check_point 
- [x] positive_1d 
- [x] positive_1d_bracket  (tested via running positive_1d)

# fitness_trait_space

- [x] max_fitness
- [x] viable_fitness

# Fitness 

- [x] fitness_landscape
- [x] max_growth_rate
- [x] carrying_capacity [function removed]


scripts/carrying_capacity.R
```diff
@@ -1,7 +1,38 @@
...
+                              parallel = FALSE) {
+     f <- function(x) {
+          ## NOTE: This is not very pretty because matrix_to_list, which we
+          ## use to iterate over this, does not preserve array-ness or
+          ## names.
+          p <- expand_parameters(trait_matrix(x, traits),
+               remove_residents(p),
+               mutant = FALSE
+          )
+          p$seed_rain <- seed_rain
```
aornugent, line 24: This is now out of date - but because you've just moved to `scripts` then perhaps it's sufficient only to note that it needs revisiting? i.e. it won't break the package in the meantime.

tests/testthat/test-support-fitness.R
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  ans <- positive_2d(f, c(0, 0), -2, 2)
+})
+
+
+test_that("bounds", {
+
+  params <- scm_base_parameters("FF16")
+  params$strategy_default$hmat <- 0.5
```
aornugent, line 34: I don't think these are used?

aornugent (approved):

Issue #346, opened 2022-08-09
Benchmark machines for comparison


dfalster:
Hi @aornugent @Becca-90 @fjrrobinson @devmitch @itowers1 @pzylstra 

We've been discussing speed lately and so i thought it would be good to compare speed of plant across our different machines. I actually setup a function `run_plant_benchmarks` for this very purpose, building off the [bench](https://github.com/r-lib/bench) package. So can you please post results in the issue below noting

- your machine [model, os]
- output of the tests

To run, please checkout the develop branch then

```
devtools::load_all()
run_plant_benchmarks()
```

dfalster:
Here's output from Daniels 2014 iMac:


```
# A tibble: 6 × 14
  expression     strategy      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory                 time           gc      
  <bch:expr>     <chr>    <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>                 <list>         <list>  
1 scm            FF16        1.02s    1.02s     0.977   17.06MB    0.977     1     1      1.02s <NULL> <Rprofmem [690 × 3]>   <bench_tm [1]> <tibble>
2 build_schedule FF16        2.88s    2.88s     0.347    8.04MB    0.347     1     1      2.88s <NULL> <Rprofmem [4,830 × 3]> <bench_tm [1]> <tibble>
3 scm            FF16w    984.23ms 984.23ms     1.02    17.07MB    1.02      1     1   984.23ms <NULL> <Rprofmem [692 × 3]>   <bench_tm [1]> <tibble>
4 build_schedule FF16w       2.91s    2.91s     0.343    7.97MB    0.343     1     1      2.91s <NULL> <Rprofmem [4,648 × 3]> <bench_tm [1]> <tibble>
5 scm            K93      723.63ms 723.63ms     1.38    52.48KB    0         1     0   723.63ms <NULL> <Rprofmem [119 × 3]>   <bench_tm [1]> <tibble>
6 build_schedule K93         8.69s    8.69s     0.115   16.66MB    0.115     1     1      8.69s <NULL> <Rprofmem [8,453 × 3]> <bench_tm [1]> <tibble>
```

dfalster:
Daniel's 2018 MacBook Pro (2.7 GHz Quad-Core Intel Core i7)

```
# A tibble: 6 × 14
[...]

PR #365, opened 2023-05-24
Set default compilation to -O2 for faster run times

dfalster:
Addresses slow run times identified in  #346. Using `pkgbuild::compile_dll()` has by default `debug=TRUE` which turns off optimisation in compiler


aornugent (approved):

aornugent:
Some numerical instability between compiler flags. Ubuntu and Windows didn't seem to care, so maybe it's a `clang` thing.

```r
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure ('test-strategy-ff16w.R:103:3'): Rainfall spline basic run ──────────
out$offspring_production not equal to 16.88961056463.
1/1 mismatches
[1] 16.9 - 16.9 == -0.000172

[ FAIL 1 | WARN 3 | SKIP 0 | PASS 2205 ]
```

PR #371, opened 2023-07-22
Adopt pointers for environment objects

aornugent:
In #362 we dramatically increase the number of environments by caching them at each ODE step. We therefore want to adopt pointers so that we can switch to environments held in memory, rather than copying them out of the cache.

This PR attempts to introduce a `shared_ptr` for environment objects following the same pattern established by Strategy.

I'm pretty new at this, but I think `shared_ptr` is a smart pointer (vs. raw pointers) that can be copied between objects (e.g. into the cache) - https://learn.microsoft.com/en-us/cpp/cpp/pointers-cpp

I've plumbed the `environment_ptr_type` into a variety of places and switched from `environment.member_function()` to `environment->member_function()` as appropriate.

I'm now a bit stuck debugging RcppR6_classes.yml as it expects `Environment` types but is receiving `environment_ptr_type`. I can see `Node` and `Individual` have a `make_*` constructor that maybe handles the type discrepancy for strategies, so maybe there's a solution there we can follow.

Any advice very welcome.


aornugent:
Closing in favour of #362

PR #382, opened 2024-01-24
Generalise growth optimisation routine

dfalster:


R/individual.R
```diff
@@ -286,15 +286,13 @@ optimise_individual_rate_at_size_by_trait <- function(
...
-
-    # grow inidividual to specified size
-    res <- grow_individual_to_height(indv,
-      height = size, env,
-      time_max = 100, warn = TRUE, filter = TRUE
-    )
-
-    res <- res$rate[colnames(res$rate) == rate]
+    # set inidividual at specified size
+    indv$set_state(size_name, size)
```
dfalster: Here you've opted to set state directly. This works for height but won't work so well for other attributes, e..g duh, biomass etc. If we use `grow_individual_to_size` you can do this.

On the other hand, setting state directly will be faster for height. Maybe an option to `set_state_via_ode` so we can switch between options?

Can you get it working with `grow_individual_to_size`

dfalster: @itowers1 can you try this out?

tests/testthat/test-individual.R
```diff
@@ -136,3 +136,25 @@ test_that("resource_compensation_point", {
...
+  
+  size_res <- optimise_individual_rate_at_size_by_trait(bounds, log_scale = TRUE, tol = 0.001, size = 3, type = names(strategy_types)[x], size_name = "height")
+  #check outcome real and positive
+  expect_true(res > 0)
+  
+  #check that height and size are convergent with same parameters
+  height_res <- optimise_individual_rate_at_height_by_trait(bounds = bounds, height = 3)
+  expect_true(height_res == size_res)  
+  
+  #set bounds - too low for positive growth
```
dfalster: Nice, can you add another option where bounds don't contain the root? I'm not sure how that behaves

tests/testthat/test-individual.R
```diff
@@ -136,3 +136,25 @@ test_that("resource_compensation_point", {
...
 })
+
+
+test_that("Maximise individual outcome", {
+  #set bounds
+  bounds = bounds(lma=c(0.01, 0.08))
+  
+  size_res <- optimise_individual_rate_at_size_by_trait(bounds, log_scale = TRUE, tol = 0.001, size = 3, type = names(strategy_types)[x], size_name = "height")
+  #check outcome real and positive
+  expect_true(res > 0)
```
dfalster: Add a test which looks at the optimised value and tests against this

e.g. 

`expect_equal(height_res$value, 0.56)`

where `height_res` is the output traits, and 0.56 is the values of the trait you get running this and are satisfied it's reasonable. 

The purpose of hardcoding a test like this is to ensure that down the track the model keeps producing the same result.

dfalster (commented):
@itowers1 A few moments on this PR, can you address?

tests/testthat/test-individual.R
itowers1: Have implemented a hardcode for both height growth rate and area heartwood growth rate. Unfortunately falls down when iterating across FF16 and FF16r, but only for area heartwood growth rate, but surprisingly not height growth rate. Could make hardcoded tests unique to certain stratgies?

tests/testthat/test-individual.R
itowers1: Done, if positive growth is possible, then finds the bound value which maximises growth.

R/individual.R
itowers1: Done.

R/individual.R
```diff
@@ -283,33 +284,54 @@ optimise_individual_rate_at_size_by_trait <- function(
...
     indv <- do.call("Individual", types)(s) # equiavlent to calling Individual<FF16w,FF16_Env> or FF16_individual(s)
-
-    # set inidividual at specified size
-    indv$set_state(size_name, size)
-    # compute rates given environment
-    indv$compute_rates(env)
     
-    #filter ode rate based on size name
-    res <- indv$ode_rates[indv$ode_names == size_name]
+    if(set_state_by_ode & size_name == "height"){
```
dfalster: should this be `! set_state_by_ode` not the "not" missing

tests/testthat/test-individual.R
dfalster: > Could make hardcoded tests unique to certain strategies?
Good idea. Just add an if statement

R/individual.R
itowers1: My idea here was that in order to be passed through to having height set, rather than grown, you need to both have set_state_by_ode be TRUE AND have the desired property be optimised as height_growth.

dfalster: Makes sense. Except for me, `set_state_by_ode` means grow the plant, not set state directly. Maybe change it to `set_state_directly`? then the logic of the term aligns with your implementation.

itowers1: Ah I see, sorry I misinterpreted the meaning of the expression. Agreed, will change now.

dfalster: Confirming it's a bad name, if we both interpret differently!

tests/testthat/test-individual.R
itowers1: This error test seems to be failing on Windows and Ubuntu, but not Mac, is that unusual?

dfalster: Nice work!

Looking at the failure, the results aren't that different.

This highlights the difference in numerical results across operating systems. 

https://github.com/traitecoevo/plant/actions/runs/7648848770/job/20842305908#step:6:1

So just increase the tolerance allowed in the test.

PR #384, opened 2024-01-30
Clarify methods used to create spline of light environment and calculate assimilation

dfalster:
This PR attempts to simplify the language and methods used to both

- calculate the splines of light availability within the canopy
- calculate a plants assimilation, integrating over these splines.

Each of the above uses some numerical methods, and the language used for these was not very clear. There were also some options that were not really needed. 

Changes include

- renamed things to do with `canopy` -> `light_availability`  (canopy was too vague)
  - [x] `canopy.h` -> `resource_spline.h` (generalised)
   -[x] also pass in full function to integrator, so can simplify and generalise `Resource_spline` (though currently limited on scale 0-1, see #385 )
  - [x] `canopy.compute_canopy` -> `light_availability.compute_environment` (consistent with names elsewhere)
  - [x] control variables `canopy_light_*` -> `light_availability_spline_*`
  - [x] `canopy::set_fixed_canopy` -> `light_availability::set_fixed_value`
  - [x] `get_canopy_at_height` & `canopy_openness` -> `get_value_at_height`
- removed
   - [x] options for adaptive integration of assimilation, reusing intervals    
   - [x] assimiolation class, preferable to call integrator straight from within strategy


inst/include/plant/resource_spline.h
```diff
@@ -0,0 +1,136 @@
...
+#include <plant/interpolator.h>
+#include <plant/adaptive_interpolator.h>
+#include <plant/ode_interface.h>
+#include <plant/util.h>
+
+using namespace Rcpp;
+
+namespace plant {
+
+class Resource_spline {
```
aornugent: Minor nitpick, but I think classes have mostly used CamelCase like StochasticPatch and AdaptiveInterpolator. We do use underscores for *_Strategy and *_Environment, but I think this would be best as ResourceSpline.

inst/RcppR6_classes.yml
```diff
@@ -749,21 +745,22 @@ StochasticPatchRunner:
...
+  name_cpp: "plant::Resource_spline"
   constructor:
     args:
       - tol: double
       - nbase: double
       - max_depth: double
+      - rescale_usually: bool
   methods:
-    canopy_openness:
+    get_value_at_height:
```
itowers1, line 757: Possibly, "get_resource_value_at_height"?

inst/include/plant/qag.h
```diff
@@ -1,4 +1,15 @@
-// -*-c++-*-
+// The class `QAG` provides methods to use Gauss-Kronrod quadrature to numerically integration of a function, e.g. `double QAG::integrate(Function f, double a, double b)`.
```
itowers1: integrate a

inst/include/plant/qk.h
```diff
@@ -1,4 +1,14 @@
 // -*-c++-*-
+
+// The class `QK` provides methods to use Gauss-Kronrod quadrature to numerically integration of a function. This class is called via the class QAG, so users should review that file first.
```
itowers1: numerically integrate a function

inst/include/plant/models/ff16_environment.h
```diff
@@ -42,15 +41,12 @@ class FF16_Environment : public Environment {
   }
 
   double get_environment_at_height(double height) const {
```
itowers1, line 43: So this just calculates light availability at this point? Will the method be expanded to return all environmental values at a given height (i.e. + soil moisture) or will this always be with respect to light. There will eventually be a completely new strategy so no need for this function to refer to anything but light availability I suppose.

inst/RcppR6_classes.yml
itowers1, line 757: Possibly, "get_resource_value_at_height"?

inst/include/plant/models/ff16_environment.h
dfalster: Yes, may eventually return a vector. In FF16 it's just a single number, light availability.

inst/RcppR6_classes.yml
dfalster: Possible, but I'll leave as is. The specification of a resource actually comes when you call it, e.g. 

`ight_availability.get_value_at_height`

inst/include/plant/resource_spline.h
dfalster: Agree

dfalster:
Thanks Andrew, Isaac: Good feedback. I addressed those concerns in commit above. Once it passes I'll merge.

PR #388, opened 2024-02-05
Remove fitness functions, consolidate vignettes 

dfalster:
All fitness related functions are moving to package {plant.assembly}. This will simplify what is presented in this package. 

This PR

- removes functions: `fitness_landscape`, `solve_max_fitness`, `viable_fitness`, `fundamental_fitness`, `assembly_parameters`, `equilibrium_birth_rate`
- removes all parameters/controls related to the above, i.e. `equilibrium_*`
- removes the slow vignettes, that takes ages to build (these mostly related to fitness functions)

Additionally

- remove dockerfile and scripts (not being used and way out of date)
- remove folder `inst/slow_vignettes` (no longer needed)
- reinstate vignettes in the build (so we can regularly check they run)
- fix errors in vignettes that arose from run not running them as part of build


aornugent:
Looks good. I haven't tested that the functionality is reproduced at the other end, but happy to look at the corresponding PR at plant.assembly when helpful.

aornugent (approved):

Issue #378, opened 2023-08-10
Tissue masses being assigned to different turnover pools


itowers1:
Spotted on develop branch. Line 171 for turnover function definition and Line 220 for use of turnover function. Argument order differs on Line 220, whereby turnover for sapwood is calculated based on bark mass and vice versa.

dfalster:
Closed via #389

PR #389, opened 2024-02-27
Fix order of inputs in net_mass_production_dt

aornugent:
Addresses #378.


dfalster:
Gah! I was initially surprised this didn't break anything, i.e. cause tests to fail. But  we have generally set rates of bark and sapwood turnover to be equal, so the error wouldn't show up until you decoupled those rates.  

How did you find this @aornugent >?

aornugent:
@itowers1 found it! I just rediscovered it when looking at turnover in #380.

PR #413, opened 2024-05-23
Clarify ode methods: Document code and vignettes

### LINKED ISSUE #411: [simplify interface] Clarify how to control step sizes in ode stepper
It's currently unclear how to 

- adjust controls ode time stepping
- control steps sizes within a given range
- set minimum / max step sizes

dfalster:
This PR better documents the ODE stepper used in plant, both in code and via a new article. In addition, we 

- move relevant files for ode_solver into a subdirectory `inst/include/plant/ode_solver`
- enable system to continue when  minimum step size is reached
- rename some items for clarity
   - `advance` -> `advance_adaptive`
   - `node_schedule_ode_times` -> `ode_times`
- clarify how to control time steps (closes #411, needed for TF24 Soil water milestone)


itowers1:
Also includes some renaming of plant -> individual

pkgdown/_pkgdown.yml
```diff
@@ -26,10 +26,20 @@ navbar:
         - text: "Theory"
         - text: "Demography of individuals, patches, and metacommunities"
           href: articles/demography.html
-        - text: "Numerical methods for solving equations"
+        - text: "Overview of numerical methods used in `plant`"
+          href: articles/methods/numerical_methods.html
+        - text: "Solving system dyanmics via the characteristic method"
```
itowers1: dynamics

src/ode_control.cpp
```diff
@@ -42,11 +42,14 @@ double OdeControl::adjust_step_size(size_t dim, size_t ord,
     if (r < 0.2) {
       r = 0.2;
     }
-    step_size *= r;
-    last_step_size_shrank = true;
-    if (step_size < step_size_min) {
-      step_size = step_size_min;
-      util::stop("Step size became too small");
+    double new_step = step_size * r;
```
itowers1, line 45: Is `new step'`s only purpose to evaluate whether `step size` is too small. Is not returned later on except on line 51, so could possibly used a comment line to indicate that the `new_step` is not actually passed on as the `new_step` value.

itowers1 (commented):
See comments. Otherwise, will the error emerging from allowing the model to proceed with the minimum step size be acceptable to users? Perhaps model needs an associated warning outputted with results that lets users know  that model was requesting a smaller step size? Perhaps this is what `last_step_size_shrank` does?

dfalster:
Thanks @itowers1 ! Hmm, good suggestion about the warning. If we added it here, you'd get maybe 1000s of messages in the terminal (one every step). Perhaps, best to add at the end of the run, as a post run assessment. That might require storing a flag. We can't just look at the step sizes as some will be short due to cohort introduction. I'm going to add this as a new issue, as it will take a little more programming than I want to do on this branch. See link below.

src/ode_control.cpp
dfalster: Thanks for checking, but actually it is potentially used in the next if statement, see 

```
    if(new_step < step_size) {
      step_size = new_step;
      last_step_size_shrank = true;
    }
 ```

PR #446, opened 2025-08-05
Simplify scm & environment interface

dfalster:
As we add TF24, I was noticing the layer of patches and clutter surrounding scm and environment objects. This commit simplifies

Now a single environment object,  

- e.g. `Environment("FF16")`. 
- Other variants removed (i.e. `make_environment`, `FF16_make_environment`, `FF16_fixed_environment`, `test_environment`). 

Changed outputs so collect happens on C++ side

- move `get_state` and `get_aux` into relevant classes
- `run_scm_collect` simplified, calls `run_scm`
- runs `tidy_patch` by default
- remove various ancillary functions: `scm_patch`, `make_scm_integrate`, `scm_state`, `patch_to_internals`, `scm_to_internals`, `species_to_internals`

Remove other small helpers 

- `first`, `last` replaced with dplyr versions
- `modify_list`

Updates throughout to make the above work with remaining examples and tests

TF24_Environment:

- add ability to access & vary soil number of layers and parameters after initial class setup
- more testing of environment


R/benchmark.R
```diff
@@ -45,15 +45,12 @@ run_resource_consumption_benchmarks <- function(its = 10) {
     p0 <- scm_base_parameters("TF24")
     p0$max_patch_lifetime = 10
     
-    p1 <- expand_parameters(trait_matrix(0.0825, "lma"), p0, TF24_hyperpar,FALSE)
-    
-    env <- make_environment("TF24", 
-                            soil_number_of_depths = layers,
-                            soil_initial_state = rep(1, layers))
+    p1 <- expand_parameters(trait_matrix(0.0825, "lma"), p0)
```
itowers1, line 48: No hyperpar object argument required anymore? Maybe just add that into the merge description if so.

R/individual.R
```diff
@@ -49,6 +50,14 @@ grow_individual_to_size <- function(individual, sizes, size_name, env,
   ret
 }
 
+vlapply <- function(X, FUN, ...) {
```
itowers1, line 53: Should these go in util.R?

R/util.R
```diff
@@ -105,20 +89,6 @@ validate <- function(x, ...) {
 }
 
 
-
-vlapply <- function(X, FUN, ...) {
```
itowers1, line 109: Oh right they have been moved from util.R, no probs!

tests/testthat/test-disturbance.R
```diff
@@ -19,7 +19,7 @@ test_that("No disturbance", {
 
   # integration test to show constant density  
   time <- seq(0, 100, len = 1e3)
-  d <- obj$density(time)
+  d <- purrr::map_dbl(time, ~obj$density(.x))
```
itowers1, line 22: Does obj$density only take one value at a time now?

itowers1 (approved):

R/individual.R
dfalster: I was trying to get rid of them! But couldn't quite. I moved them here in the file where they are used. some people would disagree with this tactic

tests/testthat/test-disturbance.R
dfalster: Correct

PR #449, opened 2025-09-19
remove ci functions

itowers1:
Simply just removing the optimisation of ci as compared to psi stem using the sperry method as it not required anymore.


dfalster (approved):
Nice!

Issue #274, opened 2020-10-29
Update scaffolder for strategy inheritance


aornugent:
The new approach to model inheritance requires some changes to the scaffolder. 

- [ ]  Pass in an additional `Environment` template - e.g. `FF16r` uses the `FF16_Environment`
- [ ]  Use inheritance approach demonstrated in `FF16r` to overload the `FF16` class, rather than copy the `FF16` strategy wholesale
- [ ]  Don't add RcppR6 bindings or tests for components that are reused (i.e. the scaffolder should not generate any mention of `FF16r_Environment` because it doesn't exist)

aornugent:
Logic in `test-cohort.R` needs reworking too:

```
for (x in names(strategy_types)) {
  s <- strategy_types[[x]]()
  e <- environment_types[[x]]
  ...
}
```
where there are more strategy types than environments:
```
> names(strategy_types)
[1] "FF16"    "FF16ppa" "FF16r"   "K93"    
> environment_types
$FF16
[1] "FF16_Env"

$FF16r
[1] "FF16_Env"

$K93
[1] "K93_Env"
```

Issue #253, opened 2020-08-28
Add bounds (or checks) to AdaptiveInterpolator


aornugent:
Need to check that our spline doesn't go below zero, as identified in `K93` with high values of `k_I`. Reprex to follow.

aornugent:
This might be as simple as `clamp`-ing the result. Not sure what the overhead is.
https://en.cppreference.com/w/cpp/algorithm/clamp

dfalster:
Triage note (still open), with a pointer on **where** the fix belongs given the odelia migration.

The interpolator core never clamps the result — `Interpolator`/`AdaptiveInterpolator` only validate the x-domain (`check_bounds`), not the y-value. The specific case this issue raised (K93 light spline undershooting below zero at high `k_I`) is still unguarded: `K93_Environment::get_environment_at_height` → `ResourceSpline::get_value_at_height` (`resource_spline.h:79`) has no lower floor. The `E <= 0 → 0` guard added in #490 (`81cfeab1`) is FF16-only and only on the PPA stepped-light path; the smooth/default path passes negative values straight through.

Importantly, the fix should **not** go into odelia. The `Interpolator` class has migrated there (#464/#456), but clamping spline output to `[0, ∞)` is a plant-domain assumption about resource availability, not a property of cubic interpolation — odelia is general-purpose and other consumers may legitimately want negative values. `AdaptiveInterpolator` stays in `plant` regardless.

The clean fix is a single floor at the shared consumer chokepoint `ResourceSpline::get_value_at_height` (`inst/include/plant/resource_spline.h`), which covers FF16/K93/TF24 at once and is migration-proof. Implementing there.

PR #456, opened 2026-03-05
refactor: migrate ODE and interpolation core to odelia

### LINKED ISSUE #26: Load odelia's DLL globally in .onLoad so linking packages can resolve compiled symbols (e.g. XAD Tape)
## Problem

Packages that `LinkingTo: odelia` and instantiate odelia's `Solver` (e.g. [`plant`](https://github.com/traitecoevo/plant)) fail to load with an unresolved symbol:

```
unable to load shared object '.../plant.so':
  symbol not found in flat namespace '__ZN3xad4TapeIdLm1EED1Ev'
```

i.e. `xad::Tape<double, 1>::~Tape()`.

## Root cause

odelia is *mostly* header-only, but parts of the bundled **XAD** autodiff runtime (notably `xad::Tape`'s destructor) are **compiled**, not header-only — those symbols live in `odelia.so`, not in the headers. A consumer's shared object therefore carries an undefined reference that must resolve to `odelia.so` at load time.

R loads each package DLL with `RTLD_LOCAL` by default, so `odelia.so`'s symbols are private and invisible to the consumer's DLL. odelia already ships `odelia_load_dll(local = FALSE)` (which re-loads `odelia.so` with `RTLD_GLOBAL`), but nothing calls it early enough — the consumer can't, because its own DLL is loaded before its `.onLoad` runs (this is also why `devtools`/`pkgload` eager-loading fails).

## Proposed fix

Have **odelia make itself global on load**:

```r
# R/zzz.R
.onLoad <- function(libname, pkgname) {
  odelia_load_dll(local = FALSE)
}
```

R loads an imported namespace (and runs its `.onLoad`) **before** it loads the importing package's own DLL:

```
load <consumer> namespace
 +- import odelia -> loadNamespace(odelia) -> odelia .onLoad -> odelia.so global   <- here
 +- useDynLib(<consumer>) -> load <consumer>.so                                     <- xad symbol resolves
```

This holds for both `library()` and `devtools`/`pkgload`, so it fixes every consumer — current and future — with no per-package, OS-specific build hacks.

## Why this over the alternatives

- **Consumer-side `.onLoad`** — runs too late under pkgload (consumer DLL is eager-loaded first). Doesn't work reliably.
- **Consumer-side Makevars link to `odelia.so`** — what `plant` does today as a stopgap; it's macOS-specific (needs an `install_name_tool` fixup because `odelia.so`'s install name is bare), embeds odelia's absolute path, and is wrong on Linux/CI. Fragile and duplicated per consumer.
- **odelia `.onLoad` global load (this issue)** — one line, one place, cross-platform.

## Tradeoff to consider

`RTLD_GLOBAL` puts odelia's/XAD's symbols in the process-wide namespace, so in principle two libraries defining the same symbol could collide. Here the risk is negligible (symbols are `xad::…` / `odelia::…`, small ecosystem). If you'd rather avoid global symbols, the alternative is to export a portable `odelia::LdFlags()` helper (the RcppParallel pattern) for consumers to link against — more explicit, more work, still per-consumer. I'd lean toward the `.onLoad` for simplicity.

## Acceptance criteria

- [ ] `R/zzz.R` with `.onLoad` calling `odelia_load_dll(local = FALSE)`.
- [ ] A downstream package that `LinkingTo: odelia` and uses `Solver` loads via both `library()` and `devtools::load_all()` with no Makevars linker hacks.

## Context

Surfaced while reconciling `plant`'s `odelia` branch (which removes plant's in-tree ODE solver in favour of odelia) with the move to `odelia@xad`. plant currently carries a `src/Makevars` link + `install_name_tool` fixup + `R/zzz.R` workaround that can be removed once this lands.

### LINKED ISSUE #464: [AD] Migrate ODE and interpolation core to the standalone odelia package
## Motivation

The core ODE stepping machinery and interpolation utilities are generic and not specific to `plant`. They have been spun out into a standalone package, [`odelia`](https://github.com/traitecoevo/odelia), so they can be developed, tested, and reused independently. This issue tracks migrating `plant` to depend on `odelia` and removing the in-tree implementations.

## Scope

- Add `odelia` as a dependency (DESCRIPTION: Remotes/Imports/LinkingTo).
- Replace internal ODE/interpolator implementations with the `odelia` equivalents.
- Regenerate Rcpp/R6 bindings for the new types (`std::vector`, `odelia::*`).
- Remove legacy solver/interpolator code (`inst/include/plant/ode_solver/*`, `interpolator.*`, `tk_spline.*`, old test helpers).
- Adjust tests and docs to match the new solver/interpolator behaviour (including deterministic stochastic patch insertion).

## Open questions / follow-ups

- [ ] Compatibility wrappers (`ode_solver_r6.h`, `ode_solver_compat.h`) were needed and solver usage was updated across core headers/source — understand whether these wrappers can be avoided / removed.
- [ ] Mutant tests caused a memory explosion and were skipped — needs attention.

## Implementation

Implemented in #456.

### LINKED ISSUE #491: [AD] Remove plant-side XAD/odelia linker workaround once odelia loads its DLL globally
## Background

On the `odelia` branch, plant no longer ships its own ODE solver and instead `LinkingTo: odelia`. Since the move to `odelia@xad`, part of the bundled **XAD** autodiff runtime (`xad::Tape`'s destructor) is compiled rather than header-only, so `plant.so` carries an undefined reference (`__ZN3xad4TapeIdLm1EED1Ev`) that must resolve to `odelia.so` at load time. R loads each DLL with `RTLD_LOCAL`, so without intervention plant fails to load:

```
unable to load shared object '.../plant.so':
  symbol not found in flat namespace '__ZN3xad4TapeIdLm1EED1Ev'
```

## Current workaround (to be removed)

To unblock the merge, plant currently carries a stopgap (commit `145a60e1`):

1. **`src/Makevars`** — links `plant.so` against `odelia.so` and, on macOS, rewrites the load command to odelia's absolute path via `install_name_tool` (odelia.so's install name is bare, so dyld can't otherwise find it).
2. **`R/zzz.R`** — `.onLoad` calling `odelia::odelia_load_dll(local = FALSE)`.

This works but is **macOS-specific**, **embeds odelia's absolute path** (wrong on Linux/CI, breaks if odelia is reinstalled elsewhere), and duplicates per-consumer plumbing that belongs upstream.

## Proper fix (upstream)

The clean fix is one line in **odelia** — load its DLL globally in its own `.onLoad` — tracked in [traitecoevo/odelia#26](https://github.com/traitecoevo/odelia/issues/26). Once odelia loads global before plant's DLL, plant's undefined `xad` symbol resolves via flat-namespace lookup with **no plant-side linker hacks at all**.

## Action (once odelia#26 lands + odelia reinstalled)

- [ ] Revert `src/Makevars` to the plain version (drop the `odelia.so` link, `ODELIA_LIBDIR`, and the `install_name_tool` fixup target).
- [ ] Remove `R/zzz.R` (no longer needed).
- [ ] Confirm plant loads and the suite passes via **both** `library(plant)` and `devtools::load_all()` / `devtools::test()`.
- [ ] Confirm a clean `R CMD check` / CI build with no linker warnings.

## Related

- traitecoevo/odelia#26 (upstream `.onLoad` global-load fix)
- `odelia` branch, commit `145a60e1` (introduced the workaround)

dfalster:
Closes #464. Closes #491.

The core ODE-stepping and interpolation machinery has been spun out of plant into the standalone **[odelia](https://github.com/traitecoevo/odelia)** package; this PR migrates plant onto it.

## What this does

- Add odelia as a dependency (`DESCRIPTION` remotes/imports/linking).
- Replace plant's internal ODE solver and interpolator with odelia equivalents (`odelia::ode::Solver`, `odelia/interpolator.hpp`).
- Regenerate Rcpp/R6 bindings for the new types (`std::vector`, `odelia::*`).
- Remove legacy solver/interpolator code (`inst/include/plant/ode_solver/*`, `interpolator.*`, `tk_spline.*`, old test helpers) — and the temporary compatibility shims (`ode_solver_r6.h`, `ode_solver_compat.h`) that bridged the transition.
- Adjust tests/docs to the new solver/interpolator behaviour (incl. deterministic stochastic patch insertion).

## Resolved review items

- [x] **Compatibility wrappers** (`ode_solver_r6.h`, `ode_solver_compat.h`) — no longer needed; removed. Core headers/source now use `odelia::ode::Solver` directly.
- [x] **Mutant tests memory explosion** — fixed and re-enabled (`80a7a1ca` "Re-enable mutants (fixed)", `9cc45c41` "Fix test-mutant speed"). Verified: `test-mutant.R` passes and process RSS is flat (~212 MB) across repeated 3-resident × 10-mutant runs — no leak/explosion.

## XAD/odelia load workaround removed (closes #491)

Because plant `LinkingTo: odelia` and references compiled XAD symbols (`xad::Tape`) that live in `odelia.so`, plant previously carried a macOS-specific stopgap (link `plant.so` against `odelia.so` + `install_name_tool` fixup in `src/Makevars`, plus a `.onLoad` in `R/zzz.R`). With **[odelia#28](https://github.com/traitecoevo/odelia/pull/28)** (closes [odelia#26](https://github.com/traitecoevo/odelia/issues/26)) loading odelia's DLL globally in its own `.onLoad`, that stopgap is removed.

One non-obvious requirement remains: odelia's global `.onLoad` only helps if odelia's **namespace** is loaded before plant's `useDynLib` (R processes NAMESPACE imports before DLLs; `LinkingTo` is C++-only and does not load the namespace). So plant now carries a standard `@importFrom odelia odelia_load_dll` (`R/odelia.R`, mirroring the `@importFrom Rcpp evalCpp` idiom) — **not** a linker hack, and it also clears the latent "Imports field not imported from: odelia" NOTE. A fresh `library(plant)` and `devtools::load_all()` both load with no `xad::Tape` symbol error.

## Sync + scaffolder

- Merged latest `develop` into the branch (brings #492 profile-plant skill, #493 K93 SCM perf, #494 FF16 `pow` sharing, #495/#274 new-strategy scaffolder→skill, #496 docs→Overstorey, #497/#253 resource-spline floor, #501 tidyselect).
- Fixed the relocated scaffolder (`scripts/new_strategy_scaffolder.R`, moved in #495 before the solver swap) to emit `odelia::ode::Solver` + `obj.get_system()` instead of the removed `plant::ode::Runner`.

## Verification

- `make rebuild` + full suite: **FAIL 0, WARN 0, PASS 2148** (1 skip = pre-existing empty TF24 test).
- `library(plant)` and `devtools::load_all()` both clean — no symbol errors.
- Removing the linker workaround is runtime-neutral (interleaved A/B, sampling off): scm/build_schedule timings unchanged within machine drift.

## Related

- [odelia#26](https://github.com/traitecoevo/odelia/issues/26) / [odelia#28](https://github.com/traitecoevo/odelia/pull/28) — upstream global `.onLoad` DLL fix.
- #274 / #495 — new-strategy scaffolder migrated to a skill.
- #253 / #497 — resource-spline floor at 0.

## Windows linking (resolved)

All R-CMD-check jobs are green (macOS + Windows). Windows can't use odelia#28's load-time symbol resolution (mingw forbids undefined symbols in a DLL), so `plant.dll` links directly against odelia's single compiled `odelia.dll` at build time via a Windows-only `src/Makevars.win`. This keeps one shared `Tape`/`active_tape_` in the process (consistent with macOS/Linux). The architecture and downstream contract are documented in odelia (traitecoevo/odelia#30); the Windows gap was tracked in traitecoevo/odelia#29.


dfalster:
Mutants running and tests passing, but 1.5x original time. Have been investigating why but no answers yet

dfalster:
Speeds now resolved!

dfalster:
Tracking issue: #464

PR #457, opened 2026-03-16
Relicense from GPL-2 to AGPL-3

dfalster:
This PR changes the license of plant from GPL-2 to AGPL-3.

**Reason**

The ODE solver code from plant has been migrated to standalone package [odelia] (https://github.com/traitecoevo/odelia/). In the near term, we will remove this code from plant and make odelia a dependency. Further, we are looking to vendor [XAD](https://github.com/auto-differentiation/xad) into odelia, a C++ automatic differentiation library released under AGPL-3. GPL-2 and AGPL-3 are legally incompatible — they cannot be combined in the same binary — so incorporating XAD requires upgrading our license. AGPL-3 is the appropriate target because it is the license under which XAD is distributed and it is the most natural copyleft license for a combined work of this kind.

**What changes**

- LICENSE updated to AGPL-3
- DESCRIPTION updated from License: GPL-2 to License: AGPL-3
- CLA.md added, requiring contributors to grant the maintainers the right to sublicense future contributions

**Practical impact on users**

For typical use of plant as a local scientific computing tool, the change from GPL-2 to AGPL-3 has no practical effect. The additional AGPL-3 obligation — that those running the software as a network service must make their source available — does not apply to ordinary research use.

**Consent**

All copyright holders on both packages have been consulted and have given consent. This to be confirmed via approval of this PR


dfalster:
Hi @aornugent @richfitz @itowers1 ! As discussed, we're looking to relicense plant from GPL-2 to AGPL-3. Can you give your formal approval for this change by approving this PR?

I've also added a Contributor License Agreement to set expectations for future license changes. For example, if we no longer need AGPL-3 and want to switch to something else.

aornugent:
I approve.

richfitz (approved):

Issue #470, opened 2026-06-18
[speed] Enable link-time optimization (LTO) to inline cross-TU hot-path calls


dfalster:
## Motivation

Profiling the FF16 benchmarks (#466) showed that the single biggest source of avoidable overhead in the hot path is **cross-translation-unit (cross-TU) function calls that cannot be inlined**.

The build currently uses `CXX_STD = CXX20` in `src/Makevars` with **no LTO** and **no `UseLTO` in `DESCRIPTION`**. Without LTO, a function defined in a `.cpp` file (its own translation unit) cannot be inlined into a caller that lives in a different TU — most importantly the templated `Individual<FF16>` / `Node` / `Species` code that drives the ODE hot loop. Every such call is emitted as a real, non-inlinable function call at every node of every timestep.

The inline-usage audit on the `optimising` branch confirmed this empirically: simply **moving small strategy helpers (`compute_competition`, `compute_competition_by_ratio`, `area_leaf`) from `src/ff16_strategy.cpp` into the header** — removing their cross-TU barrier — produced **~1.3x**, the largest single step of the whole optimisation campaign. The identical move for TF24 gave another ~1.3x.

## Why this deserves its own issue

Moving functions into headers only works for *small* helpers. The two remaining dominant hot symbols — `FF16_Strategy::assimilation()` (~90 samples) and `FF16_Strategy::compute_rates()` (~29 samples) — are too large to reasonably move into headers. **LTO is the only way to inline them across the TU boundary**, and it does so for *every* cross-TU call in the program at once, with zero code movement.

This is a build-configuration change independent of the per-function source optimisations on the `optimising` branch, so it is tracked separately here.

## Proposed experiment

1. Add `UseLTO: yes` to `DESCRIPTION` (preferred over hand-adding `-flto` to `src/Makevars` — it uses R's toolchain-aware LTO machinery).
2. `make compile`, then a same-session A/B (sample off), recompiling between states:
   ```sh
   PLANT_PROFILE_REPEATS=20 Rscript scripts/profile-benchmarks.R FF16
   ```
3. Re-verify **bit-identical** output. LTO reorders/fuses operations but should not change results absent `-ffast-math`. Run the targeted suite (`test-interpolator`, `test-strategy-ff16`, `test-strategy-tf24`, `test-individual`) plus the strict reference tests.

## Caveats / risks

[...]

Issue #466, opened 2026-06-16
[speed] Investigate speedups through code optimisation


dfalster:
Running the more expensive hydraulic model has highlighted the need to optimise for speed. While we'll optimise the hydraulic model itself, this is also a good time to revisit cumulative issues about speed throughout the codebase

This is also more achievable now with Codex or Claude. 

Below is a report generated from Codex, looking at profiling speed across FF16 and looking at potential speedups. 

----

# FF16 profiling pass, 2026-06-16

Goal: profile the `run_plant_benchmarks()` FF16 scenarios with the package DLL
compiled through `make compile`, then identify optimisation targets.

## Reproduce

```sh
make compile
Rscript scripts/profile-benchmarks.R FF16
PLANT_PROFILE_REPEATS=20 PLANT_SAMPLE_SECONDS=5 Rscript scripts/profile-benchmarks.R FF16
```

The second command uses macOS `/usr/bin/sample` from inside the profiling script.
On macOS this may require allowing the command to inspect the R process.

Baseline environment:

- Git commit: `4b5c43d1`
- Build path: `make compile` then `pkgload::load_all(compile = FALSE, debug = FALSE)`
- R: `R version 4.6.0 (2026-04-24)`
- Platform: macOS arm64
- Profiling repeat count: `PLANT_PROFILE_REPEATS=20`
- Native sampling window: `PLANT_SAMPLE_SECONDS=5`

Outputs from this run:

- `tmp/profile-benchmarks/20260616-213341/benchmark-summary.csv`
- `tmp/profile-benchmarks/20260616-213341/FF16-scm.sample.txt`
- `tmp/profile-benchmarks/20260616-213341/FF16-build_schedule.sample.txt`
- matching `.Rprof`, `*-by-total.csv`, `*-by-self.csv`, and `*.sample.log`

## Timings

One-iteration `bench::mark()` timing with the existing compiled DLL. These are
the headline baseline values to compare after each optimisation:

| expression | strategy | min | median | itr/sec | mem_alloc | total_time |
|---|---|---:|---:|---:|---:|---:|
| `scm` | FF16 | 242 ms | 242 ms | 4.139 | 17.7 MB | 242 ms |
| `build_schedule` | FF16 | 656 ms | 656 ms | 1.525 | 730.4 KB | 656 ms |

Repeated profiling totals from the same run (`PLANT_PROFILE_REPEATS=20`) give a
less noisy cumulative baseline:

[...]

Issue #465, opened 2026-06-16
[speed] Optimise `Q(z, H)` for common integer values of `eta`


dfalster:
## Background

Profiling of the FF16 model indicates a substantial amount of time is spent in transcendental functions, particularly `pow()` and `exp()`. One frequently-called function is:

```cpp
double FF16_Strategy::Q(double z, double height) const {
  if (z > height) {
    return 0.0;
  }
  const double tmp = 1.0 - pow(z / height, eta);
  return tmp * tmp;
}
```

where `eta` is typically set to `12.0`.

Because `eta` is a runtime parameter, the compiler cannot generally optimise

```cpp
pow(x, eta)
```

into a multiplication chain, even when `eta == 12`.

As a result, every evaluation of `Q()` incurs the cost of a general-purpose `pow()` call.

## Proposed change

Introduce specialised implementations for common integer values of `eta` and select the implementation once during model initialisation.

For the common case `eta = 12`:

```cpp
inline double Q_eta12(double z, double height) {
  if (z > height) {
    return 0.0;
  }

  const double u = z / height;

  const double u2  = u * u;
  const double u4  = u2 * u2;
  const double u8  = u4 * u4;
  const double u12 = u8 * u4;

  const double tmp = 1.0 - u12;
  return tmp * tmp;
}
```

Retain the current implementation as a general fallback:

```cpp
inline double Q_general(double z, double height, double eta) {
  if (z > height) {
    return 0.0;
  }

  const double tmp = 1.0 - std::pow(z / height, eta);
  return tmp * tmp;
}
```

At startup, select which implementation to use based on the configured value of `eta`.

## Implementation notes

Prefer selecting the implementation once during initialisation rather than testing

```cpp
if (eta == 12)
```

inside every call to `Q()`.

Possible approaches:

- mode flag selected at startup
- separate simulation paths
- function object selected at startup

Avoid introducing branches into the hottest evaluation loops if possible.

## Expected benefit

The exact gain is unknown and should be benchmarked.

Potential sources of improvement:

- elimination of a large number of `pow()` calls
- replacement with ~5 floating-point multiplications
- improved compiler optimisation opportunities

Because the shading calculation already uses spline-based acceleration and caches `A_l(H)`, this is expected to be a **constant-factor optimisation**, not an algorithmic one.

The optimisation is most likely to be beneficial when:

- `Q()` is evaluated many times per timestep
- `eta` remains fixed for an entire simulation
[...]

Issue #435, opened 2024-08-27
TF24 hydraulics: Performance improvement for spline class


hol430:
Hi all, we've adapted your approach for using splines to interpolate water potential from transpiration and vice versa in the lpj-guess dave model. This has been an excellent optimisation for us, so thanks for sharing the approach! One thing that we noticed however, was that a lot of time was spent in `std::lower_bound()` as called from [`spline::operator ()`](https://github.com/traitecoevo/plant/blob/develop/src/tk_spline.cpp#L207). This is called a lot in our model, and 20% of the total runtime was spent inside this particular call to `std::lower_bound()`.

What's happening here is we're searching through the x-values to find a start point for the interpolation. `std::lower_bound()` is doing a binary search internally, so it's reasonably efficient, but what we did to try and speed things up was to do a one-time check (when we initialise the spline object) to check if the x-values are equally-spaced. If they are, then we can eliminate the call to `std::lower_bound()` at time-of-interpolation by looking at our x value relative to the full range of x values (ie `(x - xmin) / (xmax - xmin)`). We then multiply that by the number of x-values and floor() the result, and end up with something comparable to the output of `std::lower_bound()` in significantly less time than it would have otherwise taken.

After doing that (and ensuring we chose equidistant x-values), we managed to speed up our simulations by 18%, so I thought I'd share our findings here in case it's of use to you guys. I get that your model works differently, so if this is not really an issue for you guys (e.g. if this code is not called so often), then feel free to close this issue :).

Edit: didn't mean to label this as a bug, but I couldn't see a better category.

dfalster:
Apologigies for the slow response @hol430. Brilliant! very glad the interpolation helped you speed up your code. And great idea about the additional optimisation. We'll implement that also.

dfalster:
Thanks @hol430 — we implemented this and it's a clear win. Reporting back in case it's useful.

We're working on a new TF24 plant-hydraulics model (on our `root_water_uptake` branch) whose solver hammers these splines extremely hard: for every individual at every ODE sub-step it solves the soil→root→stem→leaf continuum via nested 1-D root-finds plus a golden-section profit optimisation, each iteration evaluating the transpiration / root-vulnerability splines many times. Profiling confirmed your finding — `std::lower_bound` inside `spline::operator()` was the single hottest leaf.

[...]

PR #471, opened 2026-06-18
Optimise FF16/TF24 hot paths (~2.7-3.5x FF16, ~1.3x TF24)

dfalster:
Profiling the `run_plant_benchmarks()` FF16 scenarios (#466) showed nearly all time in the SCM derivative loop: `compute_rates` -> assimilation quadrature and the competition/environment rebuild. This PR works through optimisation targets 1-5 from that profiling pass, plus an inline-usage audit and an ordered spline-lookup change. All changes are bit-identical except two reciprocal-multiply reorderings, where the affected reference tests were relaxed to an explicit `1e-4` tolerance.

**Cumulative effect** (honest same-session, sample-off figures vs the original `4b5c43d1` baseline): **~2.7x one-iteration `scm` and ~3.5x one-iteration `build_schedule` on FF16** (~3.4x on the 20-repeat profile totals), and **~1.3x on TF24**.

## Algorithmic / structural
- **Eta-specialised canopy shape**: new `CanopyShape` helper (FF16/TF24/K93) picks a multiplication-chain `u^eta` for common `eta` (1,2,4,8,10,12) once at `prepare_strategy()`, avoiding libm `pow()` per quadrature point; caches `1/eta`. (#465, libm `pow` cost from #361)
- **Ratio-first competition/assimilation**: `q()`/`Q()` and the competition helpers take `u = z/H` plus a cached `1/height`, hoisting the division out of inner loops.
- **Cached dependent aux**: `competition_effect` (= `area_leaf(height)`) and a new `height_inverse` aux are computed once on state-set and reused by the competition and assimilation paths.
- **`tk::spline` index without per-point binary search**: uniform grids use the O(1) index from #435; **non-uniform grids now seed an O(1) proportional guess from the average spacing and nudge to the exact segment**, converging to the same `idx` as `std::lower_bound` (bit-identical) while replacing the O(log n) search with O(1)-amortised stepping. A diagnostic bypass showed the light-spline query is ~2/3 of the `build_schedule` run, and this change alone is ~1.17x one-iteration `build_schedule`. `operator()` was also moved into the header so it inlines into the quadrature loop. (#435)
- `growth_rate_gradient()` reuses a `thread_local` scratch `Individual` instead of copy-constructing `Internals` every call.

## Per-call overhead
- Drop `std::function` type erasure in the assimilation integrand so the templated `QK::integrate` inlines the integrand at each quadrature point.
- Hoist the invariant light-spline cap out of the integrand; use unchecked spline eval where crown bounds are already guaranteed; inline `Interpolator` accessors.
- Replace remaining `std::map<string,int>::at` lookups in the FF16 rate path with fixed integer slots (`constexpr` indices + cached aux indices in `Individual`).
- Move small hot strategy helpers (`area_leaf`, `compute_competition*`, `update_dependent_aux`) and `util::is_finite` into headers so they inline across the templated `Individual`/`Node`/`Species` boundary (this build has no LTO; enabling LTO is tracked in #470).
- De-duplicate a shared `pow(area_leaf, a_l2)` in the allocation derivatives.

## Tooling & docs
- Add `scripts/profile-benchmarks.R` harness and `notes/profile-ff16-2026-06-16.md` with per-change benchmarks and methodology.
- Add `agents.md` section 12 documenting the repo-wide optimisation strategies, with inline comments at the fragile sites (index constants must track `state_names()`/`aux_names()` ordering).

An earlier *hinted* non-uniform spline lookup (low/centre/high interval hints threaded through the call stack) was tried and backed out as slower; the guess+nudge approach above succeeds because it removes `lower_bound` on the hot grid rather than layering hints on top of it. See the notes.

Targeted tests pass: `test-interpolator`, `test-strategy-ff16`, `test-strategy-tf24`, `test-strategy-k93`, `test-individual`, `test-node`, `test-scm`, `test-environment`.

Refs #466. Follow-up: #470 (LTO).

### DIFF
--- a/inst/include/plant/models/ff16_strategy.h
+++ b/inst/include/plant/models/ff16_strategy.h
@@ -100,7 +100,16 @@ public:
   void compute_rates(const FF16_Environment& environment,
                 Internals& vars);
 
-  void update_dependent_aux(const int index, Internals& vars);
+  // Inline (header): called per state-set / ODE-state update from templated
+  // Individual<FF16> code, so inlining avoids a cross-TU call (no LTO build)
+  // and lets the now-inline area_leaf fold in.
+  void update_dependent_aux(const int index, Internals& vars) {
+    if (index == HEIGHT_INDEX) {
+      double height = vars.state(HEIGHT_INDEX);
+      vars.set_aux(COMPETITION_EFFECT_AUX_INDEX, area_leaf(height));
+      vars.set_aux(HEIGHT_INVERSE_AUX_INDEX, 1.0 / height);
+    }
+  }
 
   // * Mass production
   // [eqn 12] Gross annual CO2 assimilation
--- a/inst/include/plant/models/ff16_strategy.h
+++ b/inst/include/plant/models/ff16_strategy.h
@@ -14,6 +14,12 @@ public:
   typedef std::shared_ptr<FF16_Strategy> ptr;
   FF16_Strategy();
 
+  static constexpr int AREA_HEARTWOOD_INDEX = 3;
+  static constexpr int MASS_HEARTWOOD_INDEX = 4;
+  static constexpr int COMPETITION_EFFECT_AUX_INDEX = 0;
+  static constexpr int NET_MASS_PRODUCTION_DT_AUX_INDEX = 1;
+  static constexpr int AREA_SAPWOOD_AUX_INDEX = 2;
+
   // Overrides ----------------------------------------------
 
   // update this when the length of state_names changes
--- a/inst/include/plant/models/ff16_strategy.h
+++ b/inst/include/plant/models/ff16_strategy.h
@@ -14,6 +14,13 @@ public:
   typedef std::shared_ptr<FF16_Strategy> ptr;
   FF16_Strategy();
 
+  // Fixed integer slots for the hot ODE rate path, used instead of
+  // state_index.at("...") / aux_index.at("...") string-map lookups (those map
+  // lookups showed up in profiling, see #466). These MUST stay in sync with
+  // the order of state_names() and aux_names() below: *_INDEX is the position
+  // of that name in state_names(), *_AUX_INDEX the position in aux_names(). If
+  // you add/reorder a name there, update these constants (refresh_indices()
+  // still validates the named maps used by the R-facing paths).
   static constexpr int AREA_HEARTWOOD_INDEX = 3;
   static constexpr int MASS_HEARTWOOD_INDEX = 4;
   static constexpr int COMPETITION_EFFECT_AUX_INDEX = 0;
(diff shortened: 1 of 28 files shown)


dfalster:
Hi @aornugent @itowers1 

Huge progress speeding up plant, with identical numerical results. Have requested a light review from you, partly FYI, and partly in case I've advertentantly done something silly. 

I'm not expecting you to spend more than a few minutes reviewing  

ONce merged, I'll pull these changes into Isaac's water branch

inst/include/plant/models/ff16_strategy.h
```diff
@@ -176,12 +208,21 @@ class FF16_Strategy: public Strategy<FF16_Environment> {
...
+                             double height_inverse) const {
+    return compute_competition_by_ratio(z * height_inverse, area_leaf_);
+  }
+  double compute_competition_by_ratio(double z_over_height,
+                                      double area_leaf_) const {
+    return k_I * area_leaf_ * canopy_shape.Q(z_over_height);
+  }
 
-  // [eqn  9] Probability density of leaf area at height `z`
-  double q(double z, double height) const;
```
itowers1, line 182: Just leaving a comment here as this seemed like a big change

inst/include/plant/models/tf24_strategy.h
```diff
@@ -129,6 +153,9 @@ class TF24_Strategy: public Strategy<TF24_Environment> {
 
   virtual double net_mass_production_dt(const TF24_Environment& environment,
                                 double height, double area_leaf_);
+  double net_mass_production_dt(const TF24_Environment& environment,
+                                double height, double area_leaf_,
```
itowers1, line 162: Two definitions of net_mass_production_dt, one virtual one not

itowers1 (approved):
Seems fine to me and if tests are passing, no problems. Left two comments but they are just points to check that stood out to me, nothing wrong with them necessarily.

dfalster:
Pushed `3d4c1b1e`: **reuse `area_sapwood`/`mass_sapwood` from `net_mass_production_dt()` in `compute_rates()`** (FF16 + TF24), implementing the second "considered but not implemented" item under profiling target 7.

`compute_rates()` was recomputing `area_sapwood(area_leaf_)` and `mass_sapwood(area_sapwood_, height)` after `net_mass_production_dt()` had already computed the identical values internally (they feed respiration/turnover). The change adds a by-reference worker overload that reports the two intermediates, makes the existing 4-arg form a thin wrapper (so no caller signature changes and the large TF24 body isn't duplicated), and reuses the captured values for the `mass_heartwood` rate and the `area_sapwood` aux.

- **Bit-identical** — out-refs receive exactly the values the body already computed; no `pow()`, no reordering. No tolerance changes.
- **Timing-neutral within noise** — same-session sample-off A/B vs parent HEAD `8b6487ec`: one-iter `scm` 85.1/87.4 ms vs 87.5/87.2; `build_schedule` 186.5/194.1 ms vs 191.9/191.9; 20-rep `.Call` totals overlap (trend slightly up, which can only be noise). Kept as a clean dedup, not presented as a speedup — same category as the `pow()`-dedup and §6 rows.
- C++-only (worker not in `RcppR6_classes.yml`), so `make compile` suffices.
- Tests pass, no tolerance changes: `test-interpolator` 14, `test-strategy-ff16` 47, `test-strategy-tf24` 41 (+ empty-test skip), `test-individual` 129.

Detailed write-up + timing-history row added under §7 of `notes/profile-ff16-2026-06-16.md`.

PR #499, opened 2026-06-22
Initialise a patch from existing nodes (revive #304)

dfalster:
Revives the intent of #304 ("Initialise patch state") on current development:
let an SCM patch start from **pre-existing nodes** instead of always growing
from empty — either to **resume** an exported run, or to **seed** an arbitrary
initial size distribution at patch age 0.

#304 did much of this on the R side against an old codebase and stalled on
numerical instability. This reimplements the intent around a clean
export/re-import primitive.

## Status
The `odelia` refactor (#456) has now merged into `develop`, and `develop` has
been merged back into this branch. The diff is therefore down to just the
additive changes below; the full local test suite is green (**2175 pass / 0
fail**, 1 pre-existing empty-test skip).

## Design
The initial condition is carried **on the `Parameters` object** (applied in
`Patch::reset()`), so `run_scm(p)` just works, the run stays serialisable/
reproducible, and the seeding survives the reset at the start of every run /
schedule refinement.

- **`Parameters`**: `initial_state` (flat ODE state), `n_initial_cohorts`, the
  per-node birth bookkeeping (`initial_node_times`, `initial_patch_density`,
  `initial_pr_patch_survival`) and `initial_time`.
- **`Patch::set_initial_state()`** (from `reset()`): introduces the nodes, loads
  ODE state, restores birth bookkeeping, then does a full
  `compute_environment(false)` (a rescale of the not-yet-built light spline reads
  uninitialised grid state). A guard rejects initial conditions whose log-density
  rates would explode to non-finite values.
- **`SCM::run_next_impl`**: one additive "integrate the gap" branch — when the
  next scheduled introduction is in the future (a resumed run starting
  mid-schedule), integrate up to it without introducing. The solver clock follows
  the resumed patch age via `set_state_from_system` (no new solver API needed).
- **`Node`/`Species`**: birth-field getters/setter + per-node accessors for the R
  export helper.

## R API (`R/scm_support.R`)
- `export_patch_state(scm, step)` — capture a patch's full state (nodes + birth
  bookkeeping + residual schedule).
- `set_initial_state(p, state)` — write it onto `Parameters`.
- `make_initial_state(p, heights, densities)` — build a state from a specified
  size distribution at patch age 0.

## Tests (`tests/testthat/test-initial-state.R`)
Round-trip resume reproduces the run (seed loads **bit-exact**; forward fitness
within ~1e-5, the residual being adaptive step-size divergence across the
checkpoint), determinism of re-import, K93 multi-species, distribution seeding,
the density guard, and compatibility of seeded runs with tidy outputs /
`plot_size_distribution`.

## Docs
An Overstorey guide ("Initialising a patch") is parked under
`overstorey-staging/guides/` (`.Rbuildignore`'d); it moves into the overstorey
repo when this merges.


dfalster:
### Worked example: seeding an arbitrary initial size distribution

The new `make_initial_state()` path, run through the standard tidy output and
`plot_size_distribution()` (this is the example added to the staged Overstorey
guide in this PR; figures regenerate from chunks when it renders).

**Seeded initial size distribution at *t = 0*** (log-normal)

initial distribution

**Patch evolution from the seeded distribution** (`plot_size_distribution` on the tidy table)

size distribution

**Aggregate dynamics** (`FF16_expand_state` + `integrate_over_size_distribution`)

aggregate dynamics

**Same shape, decreasing initial density**

density panel

Issue #537, opened 2026-06-25
[AutoDiff] Inventory of numerical (finite-difference) derivatives and AD replacement opportunities


dfalster:
Parent epic: #472. Precedent: #527 / PR #531 replaced TF24f's finite-difference
profit gradient with an exact forward-mode AD + IFT gradient
(`Leaf::dprofit_droot_collar_psi`). This issue inventories the **remaining
numerical (finite-difference) derivatives** in the code and the opportunity to
replace each with AD/analytic derivatives.

## Guiding constraint: local AD must not block whole-run AD (#472)

The long-term goal (#472, scope B) is to wrap an **entire simulation run** in an
AD evaluation for parameter calibration — i.e. template the core scalar type and
let derivatives of emergent outputs w.r.t. parameters flow end-to-end.

That changes how we should add *local* derivatives now. The existing
`dprofit_droot_collar_psi` is a deliberate **cut-point**: it is hard-coded to
`xad::fwd<double>` and takes/returns plain `double`. It is fast and exact for
its one job, but as written it would:

- **not compile** against a future `Leaf<T>` (signature + members are `double`), and
- be **opaque to an outer AD pass** — derivatives w.r.t. leaf parameters
  (`vcmax`, `b`, `c`, …) cannot propagate through a `double`-valued black box.

So the design rule for any new derivative work below: **prefer derivative code
that is templated on the scalar type** (composes with an outer AD pass) over a
hand-rolled `xad::fwd<double>`→`double` pass. Where a hand-derivative is kept for
speed, register it as a *custom adjoint / IFT checkpoint* the outer tape can
traverse, and keep a templated reference path alongside it. Every site below
should be tagged as a known "AD cut-point" so the scope-B migration knows where
the chain is currently broken.

---

### A1. `Node::growth_rate_gradient` — d(growth rate)/d(height)  ⭐ highest-impact
[`inst/include/plant/node.h:188-217`](inst/include/plant/node.h#L188-L217)

The single most pervasive FD in the model. The SCM size-density transport needs
`∂g/∂h` (growth-rate gradient w.r.t. height) for **every node, every quadrature
point, every timestep**. Currently computed by perturbing height and re-running
the whole strategy `compute_rates`/`assimilation` chain via `util::gradient_fd`
(default **backward** difference, `node_gradient_eps = 1e-6`) or
`util::gradient_richardson` (off by default).

- Differentiates `Individual::growth_rate_given_height(h)` → full strategy
[...]

Issue #472, opened 2026-06-18
[AutoDiff] Enable automatic differentiation (AD) across plant for gradient-based calibration


dfalster:
## Context

Issues #467 (logistic vulnerability curve) and #468 (closed-form Weibull
transform) both concluded that special-function-free / smooth hydraulic curves
are **not** worth adopting for the SCM forward-simulation core — the transforms
are already O(1) spline lookups and the runtime gains are marginal. In both
cases the genuine motivation that remained was **automatic differentiation
(AD)**: smooth, differentiable models for gradient-based calibration and
Bayesian inference of model parameters.

That motivation is **not specific to the hydraulic path** — it applies to
`plant` as a whole. This issue tracks AD as a repo-wide capability and collects
what an AD-usable path requires across the codebase. The hydraulic curves are
kept below as a worked sub-example, but the blockers and the framework decision
are shared by every strategy (FF16, TF24, …), the canopy/light environment, the
demographic ODE, and the SCM.

## Why the codebase is not AD-friendly today

The scalar type is **hardcoded to `double` everywhere**, and several constructs
break a naive "swap `double` for an AD type" approach:

1. **`double` is baked into the core.** `state_type` is `std::vector<double>`
   (`ode_solver/ode_interface.h`, `ode_control.h`), and strategies, `Individual`,
   `Species`, `Patch`, `SCM` all store and pass `double`. AD requires the scalar
   to be a template/typedef parameter (an `AD<double>` or similar) threaded
   through the whole stack — the single largest piece of work.
2. **Spline / interpolator transforms.** `Interpolator` / `tk::spline` and the
   adaptive light-environment splines (`resource_spline.h`,
   `adaptive_interpolator.h`, used for the canopy `compute_competition`), plus the
   hydraulic transforms in `leaf_model`. Splines built from knots are not
   differentiable w.r.t. the parameters that generated the knots — the knots are
   recomputed per parameter set, not traced.
3. **Root-finders and optimisers.** `uniroot.h` (bisection), `optimize.h`
   (Brent / golden-section) are used for the light-compensation point, the
   collar-`psi` and `Ci` solves, the profit maximisation, and more. Gradients
   through an iterative solver need the implicit function theorem (IFT) or an
   AD-traceable fixed point, not a re-run of the loop.
4. **The adaptive ODE solver.** The RKCK stepper (`ode_step.h`,
   `ode_control.h`) takes error-controlled, parameter-dependent step sizes.
[...]

Issue #550, opened 2026-07-01
[TF24 hydraulics] SCM cohort-density blow-up under extreme seasonal drought


aornugent:
## Summary

Under extreme seasonal drought regimes, `run_scm()` aborts with a non-finite density error in TF24, even though all soil moisture values and individual physiology remain finite.

## Error Message

```
Error: Detected non-finite contribution
```

This is raised in `Species::compute_competition` (`inst/include/plant/species.h`) when a node's density-weighted competition contribution becomes non-finite.

## Root Cause

C++ instrumentation shows:
- The offending node's **DENSITY** has become `+Inf`
- The individual's physiology is healthy (finite height ~8-10 m, finite leaf area, finite per-individual competition ~2-3)
- All soil layers have finite moisture

The blow-up occurs in the SCM characteristic (size-density) equations under the explicit fixed-step integrator, not in the physiological model itself.

## Minimal Reproduction

```r
pkgload::load_all(".")

mpl       <- 20      # max patch lifetime (yr); blows up for ~20-70
mean_rain <- 0.5     # mean rainfall
amp       <- 0.4     # seasonal amplitude -> rainfall sweeps [0.1, 0.9]
ndepths   <- 5

p0 <- scm_base_parameters("TF24", "TF24_Env")
p0$max_patch_lifetime <- mpl
p1 <- add_strategies(p0, trait_matrix(0.07, "lma"))

env <- Environment("TF24")
env$set_soil_number_of_depths(ndepths)
env$set_soil_water_state(rep(0.2, ndepths))

x <- seq(0, mpl, length.out = mpl * 6)
y <- amp * sin(2 * pi * x) + mean_rain        # seasonal, strictly positive
env$extrinsic_drivers_set_variable("rainfall", x = x, y = y)

# Trigger the error
result <- run_scm(p1, env)
```

## Expected Behavior

The SCM should complete successfully or fail gracefully with a descriptive error message indicating the issue with the integration.

## Environment

- R version: 4.6.1
- Platform: Linux 6.8.0-124-generic

dfalster:
## Investigation report

Reproduced on `develop` and traced to root cause. Summary of what's going on, what we can fix now, and why "just make it complete" turns out to be blocked on the model rather than on a knob.

### Root cause: the SCM size-density equation blows up

The failure is in the SCM (size-density / method-of-characteristics) equations, not the physiology or the soil model. Each node integrates

```
[...]

PR #541, opened 2026-06-26
[AutoDiff] Scalar-template the state/params + exact AD growth-rate gradient (#472 scope B, Milestone C foundations + #537 A1)

dfalster:
Continues end-to-end AD for FF16 (#472 scope B; builds on #540). Scalar-templates
the per-plant state container and the FF16 parameter struct, bridges a live
`FF16_Strategy` to the AD kernel, and delivers the **exact growth-rate gradient**
(#537 item A1). Fully additive — every templated type keeps a `double` alias, so
the R/RcppR6 interface and all existing behaviour are bit-identical (FF16
reference-comparison test unchanged; full suite FAIL 0).

## Commits

1. **`Internals` → `basic_internals<S>`** (+ `using Internals = …<double>`): the
   per-plant state vector can hold an AD active type.
2. **`FF16_Pars` → `basic_FF16_Pars<S>`** (+ alias): traits can be AD active
   types. RcppR6-safe (referenced only by name, no forward declarations).
3. **`FF16_Strategy::prod_pars()`**: gathers the net-production kernel's params
   from a prepared strategy's actual `pars` + derived `eta_c`. Lifting to
   `FF16ProdPars<ad>` gives reverse-mode **trait gradients from the live model**
   (`test-ff16-live-prod-pars-ad.R`, ~1e-11 vs FD).
4. **Exact growth-rate gradient (#537 A1)**: height-growth kernel pieces
   (`ff16_fraction_allocation_growth`, `ff16_dheight_darea_leaf`,
   `ff16_darea_leaf_dmass_live`, `ff16_height_dt_crown_top`) mirroring
   `compute_rates`' `dheight/dt` assembly. The kernel reproduces a **live**
   crown-top `FF16_Strategy`'s `dheight/dt` exactly, and reverse-mode AD gives
   `d(growth rate)/d(height)` — currently a finite difference in
   `Node::growth_rate_gradient` — matching a fine FD of the live model to ~1e-8,
   with trait gradients in the same sweep (`test-ff16-growth-rate-ad.R`).

## Validation

- Full suite `FAIL 0`; FF16 reference comparison bit-identical (the `double`
  methods `area_leaf`/`assimilation_leaf`/`respiration`/`turnover`/
  `net_mass_production_dt_A` route through the kernel as a single source of truth).
- AD tests link the XAD tape from odelia's `.so` (and `plant.so` for the live
  strategy's compiled methods); they run on the installed package in CI and skip
  under `load_all`.

## Scope / next

[...]

PR #553, opened 2026-07-01
Prototype reverse-mode AD of SCM outputs

dfalster:
Adds exact reverse-mode automatic differentiation of SCM outputs with respect
to traits and the birth-rate driver, for FF16, TF24 and TF24f. One backward
sweep returns the derivative of a scalar emergent output w.r.t. all inputs at
once — the shape of a calibration objective or a selection gradient — at
machine precision, with no per-trait re-run and no step-size lottery.

The C++ core gains a third template axis, the scalar type `S`, defaulting to
`double`. Every existing use and the whole R interface still instantiates
`double` and is bit-identical; the reference suite is untouched.

This is a prototype, not a finalised API: the access-point names and the engine
structure are expected to change. Every access point is either finite and
validated against a finite difference over the same reconstruction, or gated
with a clear error — never plausible-but-wrong. `notes/ad-refactor-optimize-roadmap.md`
has the follow-ups and the guide added here has the techniques and the limits.

Refs #472 (scope B), #537.


dfalster:
@aornugent Could you provide a review on this before merging? We've demonstrated AD capability with one prototype implementation. We ended up implementing outside of the SCM object, but I'm not sure this is the right long-term solution. I'm thinking we land this PR and then re-implement, refine in a new round of work. I'm open to variants.   I'm interested in your thoughts.

aornugent:
This is huge! 

The spike proves that this is possible — exact reverse-mode gradients through the SCM for the strategies I thought would be hardest, along with validation of that the results are accurate and flushing out the real difficulties on the way

I have opened #559 with a design roadmap for what comes next using this branch as the basis and keeping the tests you developed for regression. 

Where #559 goes further is arguing that most of the surrounding machinery isn't needed. plant
already links odelia, which is where the ODE solver, the differentiable spline, and the XAD tape
live. Build on that and a lot falls away: the gradient becomes `run_mutant`/`run` differentiated
rather than three `*_emergent.cpp` engines; the R-side `Rcpp::as<Environment>` round-trip becomes
a native boundary; `stand_stage_history` collapses into one "record the adaptive nodes, replay
them fixed" step; `<T,E,S>` goes back to `<T,E>`; five hand-managed tapes become odelia's one.
The reasoning is all in the PR.

A useful side effect: because that machinery belongs in odelia rather than plant, it becomes
general AD infrastructure — reverse-mode gradients and Jacobians for any model built on odelia,
not just the SCM.

So the recommendation is to hold off merging the prototype and land the small odelia surface
first, using these Jacobians as the oracle, then port the plant path onto it. Merging as-is means
taking on a second tape lifecycle and scalar axis that the next step just unwinds. Would be good
to get your read on the direction before anyone starts building.

PS. I got here by working in a meta-repo with plant and odelia as submodules, which gave Claude access to both codebases and XAD at once (rather than working on plant with odelia installed).

dfalster:
Thank you @aornugent ! I agree 100% with your proposed directions. I'll hold off on merging this as we develop a better long-term solution.

dfalster:

---

## Status: prototype

This is a **prototype** of automatic differentiation in `plant` — a working,
end-to-end-validated spike that establishes the capability and the techniques, **not a
finalised API**. It is deliberately additive and bit-identical to the reference model
(see below), so it is safe to land as a foundation, but the access-point names, the engine
structure, and several honest-scope gaps are **expected to be refined later** (see
"Deferred" and `notes/ad-refactor-optimize-roadmap.md`). Suggest we land in develop then refine in a new feature branch.

## Summary

Adds **exact reverse-mode automatic differentiation** of `plant`'s SCM outputs w.r.t.
traits and the birth-rate driver, for FF16, TF24 and TF24f (#472 scope B, #537). One
backward sweep returns the derivative of a scalar emergent output w.r.t. *all* inputs at
once — the "many inputs → one scalar" shape of a calibration objective or a selection
gradient — at machine precision, with no per-trait re-run and no step-size lottery.

The full narrative (how it was built, the techniques, the validation, the honest scope)
is in the guide added here: `overstorey-staging/guides/autodiff-trait-gradients.qmd`.
This description is the map; the guide is the territory.

## The discipline: additive and bit-identical

The C++ core gains a **third template axis** — the scalar type `S`, defaulting to
`double`. Every differentiable piece is templated on `S`; AD only ever happens in C++ by
instantiating with an XAD active type (`odelia`'s reverse-mode tape). Because `S` defaults
to `double`, every existing use and every line of the RcppR6/R interface still names the
`double` type and is **bit-for-bit unchanged**. The ordinary model, and the reference test
suite, are untouched — the AD is purely additive. The whole spike landed as a long series
of individually-verified, bit-identical steps rather than one rewrite.

Two recurring principles do all the work (both detailed in the guide):

1. **Never differentiate a solver — differentiate its converged point.** Root-finds via
   the implicit function theorem, optimisations via the envelope theorem, adaptive
   integrators via a **frozen-schedule replay** (a `double` discovery pass finds the
   schedule + resident light; a second pass replays it with the trait active).
2. **The derivative of a built quantity must come from the same construction as the
   value** — the same spline, the same harvested operating point.

## Prototype access points

The prototype surface (names/signatures may change on refinement). All dispatch on
strategy; all are ordinary compiled package calls (no on-the-fly `sourceCpp`). Given a
`run_scm(save_RK45_cache = TRUE)` resident:

| Call | Differentiates | Strategies |
|---|---|---|
| `stand_gradient(scm, metrics, traits, species, feedback)` | metrics × traits Jacobian of emergent stand outputs (canonical entry) | FF16, TF24f (all metrics); TF24 (offspring only) |
| `offspring_production_gradient(scm, traits, species)` | `d(offspring_production)/d(trait)` — the invasion / selection gradient | FF16, TF24, TF24f |
| `stand_state_jacobian(scm, traits, species)` | per-cohort state × trait Jacobian (escape hatch for any downstream metric) | FF16, TF24 |
| `grow_individual_to_size_gradient(individual, sizes, size_name, env, traits)` | `d(t*)/dθ` and total `d(state at t*)/dθ` for one plant in a fixed env | FF16, TF24f |
| `birth_rate_gradient(scm, metrics, species)` | the birth-rate axis: resident `d(census)/d(birth_rate)` and `d(net_reproduction_ratio)/d(birth_rate)` = `dR0/db` (demographic equilibrium) | FF16 |

`feedback = "frozen"` (default) is the rare-mutant invasion gradient (canopy held fixed);
`feedback = "resident"` is the resident-total gradient via a coupled whole-stand replay
(every trait re-shades the canopy; on a multi-species stand, the cross-species total).

## Validation

- **Whole test suite green**, unchanged reference behaviour: FAIL 0, PASS 2610 (`make
  test` / plain-R `load_all`); the AD-specific tests run against the installed DLL via
  `make test-ad` (FAIL 0, PASS 362).
- **AD-vs-AD regression fixture** (`tests/testthat/fixtures/gradient-baseline.rds`,
  `scripts/gradient_fixture.R`) pins every compiled gradient engine to its validated value
  — bit-identical for the relocation/frozen paths, noise-floor for the coupled paths.
- **AD-vs-FD** throughout: each engine is validated against a finite difference over the
  *same* reconstruction (the honest contract), plus the physics where available.
- FF16 emergent gradients are faithful to ~1e-13…1e-8; TF24/TF24f to the leaf-optimiser
  floor (~5e-7), as expected for a re-solved optimum.

## Honest scope & known limits

Stated plainly in the guide's "Where it works, and where it breaks" section (swept
empirically across patch lifetime and species count):

- **FF16 has no hard break** in the tested envelope (patch lifetime → 200, → 5 species).
  Soft limits only: cost grows with cohort count (trait-count-free by design), and the
  frozen-*grid* reconstruction drifts gently (~1e-6 relative at long horizon).
- **The coupled/resident gradients are the frozen-*grid* total** — they freeze the ODE
  schedule + knot positions and re-evolve the stand. Against a *full* `run_scm` FD (which
  lets the adaptive grid re-respond) they differ ~16–30 % for geometry traits (~half is
  genuine grid response). This is the invasion/selection gradient's exact object; consumers
  wanting the fully-adaptive derivative should know the gap.
- **TF24f resident census single-species gates at patch lifetime ≳ 5**, and multi-species
  resident gates on a refined (clustered) schedule — both raise a *clear error*, never a
  NaN. Root cause is the frozen-step replay meeting a stiff `log_density ↔ canopy`
  feedback the live SCM only tames by adaptive stepping. **Plausibly the same family as
  #550** (the SCM size-density blow-up under drought) — same `d(log_density)/dt = −dg/dh −
  mortality` equation with a large `dg/dh` from the TF24 leaf — but a distinct
  manifestation: here `run_scm` stays healthy and the true gradient exists; only the replay
  diverges. #550's root fixes (NSC #517 / hydraulic-optimum continuity) would help both.
- **TF24 census/resident metrics are not available** (only `offspring_production`); the
  census number density needs a leaf-optimiser cross-sensitivity the linearised harvest
  does not capture. TF24f (analytic tracked collar) gets census instead.

Every access point is **finite-and-validated or gated-with-an-error — never
plausible-but-wrong**, which is the property a downstream consumer needs.

## Deferred (downstream / follow-ups)

Calibration, likelihoods, and adaptive-dynamics / equilibrium solvers live in *downstream*
packages (e.g. `regnans`) — `plant`'s job is to expose the gradient cleanly. Follow-ups
noted in `notes/ad-refactor-optimize-roadmap.md`: TF24 census; the `ResidentHarvest` seam
for multi-patch/stochastic variants; standalone `Leaf` calibration; TF24f resident
birth-rate; cross-species `net_reproduction_ratio`; and (the one change that would retire
most gated cases) adaptive sub-stepping in the replay.

## Notes for review

- **Large but layered.** ~17.5k net lines / 143 commits, but additive and bit-identical to
  the reference model at every step (the reference suite never changed). The C++ AD engines
  live in `src/{ff16,tf24,tf24f}_emergent.cpp` + `inst/include/plant/gradient/` +
  `inst/include/plant/models/*_production_kernel.h`; the R API in `R/emergent_gradient.R`
  and `R/tf24*_emergent_gradient.R`.
- The foundation slice (`<T,E,S>` templating + exact AD growth-rate gradient) was PR #541,
  now **closed** and folded in here as branch ancestry.
- The guide is a staging copy destined for the Overstorey site; it builds no figures.

Issue #517, opened 2026-06-24
[TF24 nsc] Enable resource storage, presumably NSCs, to buffer short-term variation in productivity


dfalster:
Plants currently in the FF16 and TF24 model effects of lower productivity are instantaneous, translating into increased mortality. Realistically, plants have non-structural carbohydrates, which allow them to buffer the ups and downs of short-term variation. In a thesis, Fiona Robinson showed the shortcomings of not having it: very high spikes in mortality whenever productivity drops

We should attempt to capture this phenomenon by adding NSC storage to the TF24 model.

Issue #485, opened 2026-06-18
[TF24 hydraulics] Non-finite psi_soil (NaN) under low-rainfall drought in multi-layer root-water model


dfalster:
## Summary

Running the TF24 multi-layer root-water scenario in `vignettes/models/test-root.qmd`
under a **low-rainfall (drought) regime** aborts with:

```
Error: set_physiology received non-finite psi_soil at layer=0; psi_soil=nan
```

The run gets partway through the patch lifetime, then a soil layer's water
potential becomes non-finite and `Leaf::set_physiology` rejects it.

## Not a regression from the `develop` merge

This was first noticed right after merging `develop` (which brought in #471,
the FF16/TF24 hot-path optimisation), but the merge is **exonerated** — the
failure reproduces identically on both sides:

| Build | rainfall mean | result |
|---|---|---|
| post-merge `2f9d5337` | 1.0 | ✅ OK, `offspring_production = 0.394` (healthy cohort) |
| post-merge `2f9d5337` | 0.5 | ❌ NaN `psi_soil` |
| pre-merge `5cf6694e`  | 0.5 | ❌ NaN `psi_soil` (identical) |

(15 soil layers, initial moisture 0.2, fine node schedule, MPL = 10, `-O2 -g`
build via `make compile`.) The current build runs the scenario fine at the
original rainfall mean of 1.0.

## Actual trigger

The vignette's rainfall driver was changed at some point from

```r
y = 0.25*sin(2*pi*x) + 1     # mean 1.0  (ran fine, offspring ~0.39)
```
to
```r
y = 0.25*sin(2*pi*x) + 0.5   # mean 0.5  (drought -> NaN)
```

At mean 0.5 the input is roughly half the original; with 15 layers starting at
moisture 0.2 (sat 0.428) the column dries faster than rainfall replenishes it.
The cohort is driven to near-extinction (`offspring_production ~ 1e-10`, vs
~0.39 at mean 1.0) and a soil layer dries past the point where the retention
curve is well-defined. Short runs survive (MPL = 2–4 complete with tiny
offspring); by MPL ≈ 5–10 the cumulative drying tips a layer over.

## Mechanism (robustness gap)

The bucket model has **no residual-moisture floor and no positivity guard**.
The retention curve in `inst/include/plant/models/tf24_environment.h`:

```cpp
double psi_from_soil_moist(double soil_moist_) const {
[...]

PR #554, opened 2026-07-02
NSC storage pool for TF24 with reserve-gated growth (#517)

### LINKED ISSUE #549: [TF24 hydraulics] NaN rate bypasses residual-floor positivity guard in TF24_Environment::compute_rates
## Summary

When a TF24 soil layer sits at the residual moisture floor, `psi_soil`
becomes enormous (confirmed: 93,111,981 MPa at `theta = theta_r = 1e-2`).
Fed that value, the leaf hydraulic solve can emit a non-finite per-layer
`soil_consumption_`, which arrives at `TF24_Environment::compute_rates()` as
a NaN `resource_depletion[i]`. The issue #485 positivity guard added to
protect against exactly this kind of drought singularity does not catch it,
because `NaN < 0.0` is `false` in IEEE 754 — so the NaN rate is written
straight into the soil state, unguarded.

## Confirmed causal chain

1. A soil layer dries toward the residual floor
   (`theta <= soil_moist_residual` in
   [tf24_environment.h](inst/include/plant/models/tf24_environment.h)).
   `psi_from_soil_moist()`
   ([tf24_environment.h:272](inst/include/plant/models/tf24_environment.h#L272))
   floors `theta` at `soil_moist_residual` but still evaluates the retention
   curve there, producing an enormous (not infinite) `psi_soil` — reproduced
   below at 9.3e7 MPa.
2. That extreme `psi_soil` is fed into the leaf hydraulic solve
   ([leaf_model.h](inst/include/plant/leaf_model.h)). Once a layer's
   `psi >= ~1e5 MPa` (theta <= ~0.02), the per-layer `soil_consumption_`
   (used via `evapotranspiration_dt`) goes non-finite.
3. That NaN becomes a NaN `resource_depletion[i]` passed into
   `TF24_Environment::compute_rates()`
   ([tf24_environment.h:213](inst/include/plant/models/tf24_environment.h#L213)).
4. `rate = (water_input - water_flux[i] - resource_depletion[i]) / dz[i]`
   ([tf24_environment.h:247](inst/include/plant/models/tf24_environment.h#L247))
   becomes NaN. The issue #485 positivity guard just below it —
   `if (theta <= soil_moist_residual && rate < 0.0) { rate = 0.0; }`
   ([tf24_environment.h:248](inst/include/plant/models/tf24_environment.h#L248))
   — does **not** catch it, since `NaN < 0.0` evaluates to `false`. The NaN
   rate is set via `vars.set_rate(i, rate)` regardless.
5. The integrator then drives `theta -> NaN` for that layer, so
   `psi_from_soil_moist(NaN)` returns NaN, and several steps later
   `Leaf::set_physiology()` aborts with `"non-finite psi_soil at layer=N"`
   ([src/leaf_model.cpp:190](src/leaf_model.cpp#L190)) — reported at the
   deepest layer in the run that surfaced this, consistent with drought
   progressing downward through the soil column.

The plant's own rates stay finite throughout this (e.g. `height_dt` floors
to 0); the only material corruption path is via the soil feedback above.
The eventual abort names the *symptom* (a NaN already in `theta`), not the
*cause* (the non-finite `soil_consumption_`/`resource_depletion[i]` several
steps earlier), which makes this very hard to debug from the crash alone.

## Reproduction (confirmed working)

`test_soil_singularity_repro.R` at repo root isolates steps 1 and 4 directly
(no need to drive the full leaf hydraulic solve into NaN):

```r
pkgload::load_all(".")
ndepths <- 5
theta_r <- 1e-2 # soil_moist_residual in tf24_environment.h

env <- Environment("TF24")
env$set_soil_number_of_depths(ndepths)

psi <- env$psi_from_soil_moist(theta_r)
# psi = 93111981 MPa -- confirms step 1

state <- rep(0.2, ndepths)
state[ndepths] <- theta_r
env$set_soil_water_state(state)
env$extrinsic_drivers_set_constant("rainfall", 0.5)

resource_depletion <- rep(0, ndepths)
resource_depletion[ndepths] <- NaN
env$compute_rates(resource_depletion)

env$soil$rates[ndepths]  # NaN -- guard did not zero it, confirms steps 3-4
```

Output:

```
psi_soil at residual floor: 93111981 MPa
rate at layer 5 = NaN
NaN < 0.0 = NA
BUG CONFIRMED: NaN rate not caught by positivity guard
```

## Suggested fix directions

- Harden the issue #485 positivity guard in `compute_rates()` to also catch
  non-finite rates, not just negative ones, e.g.
  `if (theta <= soil_moist_residual && !(rate >= 0.0)) { rate = 0.0; }`.
- Investigate/guard the leaf hydraulic solve so `soil_consumption_` cannot go
  non-finite in the first place when fed an extreme (but finite) `psi_soil`
  near the residual floor — this is the actual point of corruption and the
  more robust fix.
- Consider validating `resource_depletion[i]` for finiteness where it enters
  `TF24_Environment::compute_rates()`, so a future upstream NaN fails loudly
  at its source rather than silently propagating into the soil state.

dfalster:
## Summary

Adds a non-structural-carbohydrate (NSC) **storage pool** to the TF24 strategy so growth and mortality respond to *buffered* carbon rather than instantaneous net production. This is the principled fix for the SCM cohort-density blow-up (#550): under extreme seasonal drought the instantaneous growth-dependent mortality `a_dG1*exp(-a_dG2*productivity_area)` overflowed to ~1e32 and ran the coupled `(log_density, mortality)` ODE away. Local fixes (mortality caps, growth smoothing, density limiters) were all chaotic knife-edges (see #550); NSC buffering is the root-cause fix.

Design aligns with Stefaniak et al. 2026 (the FF16 "plantNSC" variant; Falster is a co-author) and was steered by the biological principle that **a plant should not grow unless it has ample carbon in storage**.

## Model (new ODE state `storage` S; capacity `S_max = a_st1·mass_sapwood`)

- **Reserve-gated growth**: growth/reproduction proceed at the *production* rate scaled by a smooth gate `G(r) = logistic((r − a_st2)/w)`, `r = S/S_max`. Full reserves → grow at production rate (healthy dynamics preserved); low reserves → redirect carbon to refill; drought → growth pauses and resumes only once reserves refill. Growth is gated at the production rate rather than metered out as `a_st2·S` — with sapwood-scaled capacity, storage is tiny relative to seedling productivity (a full pool funds only ~6% of production at birth), so a discharge-rate limit chokes establishment.
- **Storage dynamics** `dS/dt = P − growth_flux`, floored at 0 so `r ∈ [0,1]`.
- **Reserves-based mortality** `a_dG1·exp(−a_dG2·r)` — now **bounded** in `[a_dG1·e⁻²⁰, a_dG1]`, eliminating the 1e32 spike by construction; death is gradual as reserves deplete.
- New params `a_st1` (0.10), `a_st2` (growth threshold, 0.10), `a_st3` (birth fill, 0.8); seeded at birth via `set_initial_states`. TF24f inherits the state and seeding automatically.

## Also fixes #549 (companion soil-water bug)

The #550 sweep surfaced a second, independent failure mode: near the residual soil-moisture floor `psi_soil` explodes (~1e8 MPa) → the leaf hydraulic solve emits non-finite consumption → it cascades through the inter-layer drainage (`soil_K` of a blown-up `theta`). Fixed in `tf24_environment.h`:
- harden the #485 positivity guard to catch non-finite rates (`!(rate > 0)`, since `NaN < 0` is `false` in IEEE 754);
- cap `psi_soil` at `soil_psi_max_ = 1e3` MPa (root conductance is ~0 well below);
- clamp `soil_K` input `theta` to `[0, soil_moist_sat]`.

Kept in this branch since both are required for the #550 sweep to complete. **Closes #549.**

## Validation

- **#550 sweep completes robustly** across `mpl ∈ {20,30,50,70} × amp ∈ {0.30–0.40}` (all 20 cells OK) — the exact grid where the rejected mortality cap was chaotic.
- Single-plant drought probe confirms the intended behaviour: healthy plants grow with stable ~45% reserves; drought **pauses growth** and depletes reserves; mortality rises **gradually to a bounded ~5.5/yr**; growth **resumes after reserves refill**.
- `test-strategy-tf24`, `test-strategy-tf24f`, `test-environment-TF24` all pass. Regenerated RcppR6; updated tf24/tf24f `Defaults`; inverted the #550 regression test to assert completion; re-pinned offspring-arrival (227.9 → 25.4; the 2-species case now shows the slower species largely competitively excluded).

## Notes

- Healthy reproduction drops ~9× (227 → 25) — a defensible consequence of reserve-gated growth plus reserves-based mortality, but worth reviewer attention. The growth threshold `a_st2` is the knob (0.10).
- Params are **prototype defaults, not calibrated**. Calibration vs Fiona Robinson's thesis / Stefaniak et al. 2026 is the agreed follow-up.


inst/include/plant/models/tf24_environment.h
```diff
@@ -261,10 +275,15 @@ class TF24_Environment : public Environment {
...
-    // drains nothing (K = 0). See issue #485.
-  return K_sat * std::pow(std::max(theta, 0.0)/soil_moist_sat, 2*n_psi + 3);
+    // Clamp theta to [0, soil_moist_sat] (issue #485/#549): an intermediate
+    // explicit-RK stage can probe theta < 0 (std::pow(negative, non-integer) is
+    // NaN -> a non-positive layer drains nothing, K = 0) or, once the soil
+    // feedback is perturbed, a theta far above saturation, whose large positive
+    // exponent would otherwise make the inter-layer water_flux astronomically
+    // large and cascade the blow-up across layers. Physically K saturates at
+    // K_sat, so clamp there.
+  const double t = std::min(std::max(theta, 0.0), soil_moist_sat);
```
itowers1, line 285: Agreed that this will fix problems, wondering about issues to do with conservation of total moisture though. For example, capping theta to soil_moist_sat would effectively discard some water without being tracked. Not a strict blocker to this PR, but something we may need to think about.

itowers1 (approved):
Looks great! I like the implementation, in the sense that NSCs are not just "leftovers" as per passive storage. It does place greater emphasis on the relationship between prior productivity (now described by reserves) and mortality, so will need to investigate this behaviour closely. Added an example to the vignette of a smooth decline in rainfall as the sudden drop off in soil moisture seemingly immediately used up reserves.

dfalster:
### Updated from `develop` + scientific version applied (via #560)

Merged the latest `develop` into this branch (`69eba1c4..d06a729f`) — it was previously conflicting. Now `MERGEABLE`.

**Conflict resolution**
- **Drought regression test** (`test-strategy-tf24.R`): kept this branch's version, which asserts the SCM run **completes** (the NSC pool bounds mortality and fixes the #550 blow-up at the root cause). Dropped develop's graceful-failure variant and its `skip_on_os("windows")` — completion is now platform-independent, so there's nothing to skip.
- **`offspring_production` pins**: re-measured against the *merged* science. develop's #548 hyperparameterisation shifted the baseline, so neither the old branch value nor develop's value survived:
  - single species: `25.4 → 82.09`
  - two-species dominant: `16.9 → 67.54`; the slower species stays competitively excluded (`< 0.5`, now ~5 orders down).
- `R/RcppR6.R` regenerated.

**Applied #560's versioning to this science update**
Per the scheme, an equation/default change that alters output for identical inputs bumps the model's scientific version:
- `TF24_Strategy::scientific_version` **2 → 3**
- `TF24f` compound version auto-tracks the major component → **`TF24f@v3.1`** (approximation revision unchanged)
- Re-blessed the `test-model-version` drift-guard snapshot; added a NEWS entry. **This invalidates the `logpile` cache for TF24/TF24f**, which is the intended signal.

**Local verification**: full suite `2413 pass, 0 fail` (4 pre-existing/opt-in skips).

---

@itowers1 — your moisture-conservation point on the `theta` cap (`tf24_environment.h`, capping to `soil_moist_sat` discards water without tracking it) is **not addressed here** — this update is scope-limited to the develop merge + version bump. Leaving it open as you suggested; worth a follow-up issue so it isn't lost.

PR #556, opened 2026-07-03
TF24 hydraulic scenario evaluation framework (gateway checks)

dfalster:
## Summary

Adds a rerunnable, branch-agnostic **scenario evaluation framework** that scores
how well the current TF24 build reproduces the expected hydraulic behaviours in
the scenario table from #555. For each scenario it translates the qualitative
descriptors into concrete parameters/environment/rainfall, runs the SCM,
classifies the run (success/failure, with a persisted/extinct/crashed
breakdown), and scores it against the expected outcome — producing a
provenance-stamped scorecard. Many scenarios are expected to fail today; those
are the improvement targets (e.g. the NSC work in #554).

It doubles as long-term **gateway-check** infrastructure: re-run across branches
to see the scorecard move as features land.

Part of #557 (tracking issue). This is the evaluation half; the expectations
table it consumes is #555.

## What's included

- **Engine** — `R/scenario_eval.R`: `read_scenario_table` / `read_scenario_mapping`
  → `scenario_to_config` → `build_scenario` → `classify_scm_run` →
  `evaluate_scenario` → `run_scenarios`, plus `scenario_summary`,
  `scenario_run_metadata`, `scenario_model_fingerprint` and
  `scenario_generate_report`.
- **Data** — `inst/scenarios/scenario_mapping.csv` (editable qualitative→quantitative
  lookup) and `inst/scenarios/model_scenarios_hydraulic.csv` (the #555 table,
  corrected in place — see below).
- **Report / runner** — `inst/reports/scenario_scorecard.Rmd`,
  `scripts/run_scenario_gateway.R`, and a `make scenarios` target.
- **Tests** — `tests/testthat/test-scenario-eval.R` (fast unit tests) and an
  opt-in baseline-diff gateway `tests/testthat/test-scenario-gateway.R`
  (`PLANT_RUN_SCENARIOS=1`), with a recorded baseline scorecard.
- **Parallelism** — `run_scenarios(workers = N)` uses fork-based `mclapply` so it
  works under `load_all`; a PSOCK cluster would run the installed (stale) build.
- **Rerun cache** — `run_scenarios(cache = path)` content-addresses each result
  by resolved config + `max_patch_lifetime` + all `ctrl` fields + a broad model
  fingerprint (package version + compiled `.so` md5 + md5 of all `R/*.R` + the
  scenario CSVs). A model recompile / R-source edit reruns everything; editing
  one mapping cell reruns only affected scenarios; the cache errs toward
  rerunning.

## Model change

Exposes `g1_TF24` as a settable `TF24_Pars` field. This overlaps with #548,
which is merged in here and is the authoritative version (it also derives the
hydraulic traits — see dependencies).

## CSV review (findings addressed)

The #555 table had: two exact duplicate rows, a fully confounded trait/environment
design, only one expected success, misleading scenario names, a `Psi_crit,b,c`
column that collapses to a single free trait, a filename typo and a BOM. The CSV
is corrected in place (single source of truth): duplicates removed, matched
controls added, names clarified, typo/BOM fixed. Confounding and binary
expectations are documented as interpretation caveats in the scorecard rather
than mechanically altered.

## How to run

```r
# one-off
scorecard <- run_scenarios(workers = 4)
scenario_summary(scorecard)
scenario_generate_report(scorecard, "scorecard.html")

# or the gateway runner (fork-parallel, writes an RDS + prints the summary)
# make scenarios         # SCENARIO_MPL / SCENARIO_WORKERS / SCENARIO_OUT to tune
```

Opt-in gateway test: `NOT_CRAN=true PLANT_RUN_SCENARIOS=1 devtools::test(filter = "scenario")`.

## Baseline results (post-#548, `max_patch_lifetime = 100`): 5/8

- #548 **fixes S03** (previously crashed in an expected-success scenario).
- The two matched controls **S07/S08 now crash**, and **S04** runs where failure
  was expected — the clearest remaining targets.

## Dependencies / base

- Targets **`develop`** and is up to date with it (**#548 is now merged into
  develop**). #548's `TF24_hyperpar` derives `p_50/c/b/psi_crit` from `K_s` and
  `g1_TF24` from `rho`, so those columns are set via `K_s`/`rho` and left to the
  hyperpar rather than passed as input traits.
- Carries the scenario CSV from **#555** (vendored into `inst/scenarios/` and
  corrected in place), so the PR is self-contained even though #555 itself is not
  yet in develop.

The diff is just the framework — no unrelated model changes.

## Caveats

- The qualitative→quantitative mapping uses a first-pass `k = 2` multiplier off
  the *E. saligna* defaults; recalibrate in `scenario_mapping.csv` (one file, no
  code).
- The `root_depth_shape_eta` top/bottom-heavy sign is a documented assumption to
  verify against `Q()`.
- Under #548's default, `g1_TF24` does not scale with `rho`, so that CSV column
  has no effect unless recalibrated.

See `notes/plan-tf24-scenario-framework.md` for the full design.


tests/testthat/test-scenario-gateway.R
```diff
@@ -0,0 +1,40 @@
...
+## scripts/run_scenario_gateway.R.
+##
+## Slow (runs the full SCM for every scenario); gated behind an env var.
+
+test_that("scenario outcomes match the recorded baseline", {
+  skip_on_cran()
+  skip_if_not(nzchar(Sys.getenv("PLANT_RUN_SCENARIOS")),
+              "Set PLANT_RUN_SCENARIOS=1 to run the (slow) scenario gateway.")
+
+  baseline_path <- test_path("test_data", "scenario_baseline.rds")
```
itowers1, line 15: Nothing currently generates this

itowers1 (commented):
Created new folder called scenarios under inst. Right place/right folder? Otherwise, happy to go ahead.

tests/testthat/test-scenario-gateway.R
dfalster: Good catch — you were right that nothing wrote to this path. `run_scenario_gateway.R` defaulted to `scenario_scorecard.rds`, so re-blessing the baseline meant a manual copy. Fixed in 0544aa03: added a `make bless-scenarios` target that regenerates `tests/testthat/test_data/scenario_baseline.rds` in place (via `SCENARIO_OUT`), and updated the test comment to point at it. The baseline `.rds` itself is committed, so the test still runs as-is.

dfalster:
On the `inst/scenarios/` folder — yes, that is the right place, so happy for you to go ahead. The CSVs are resolved with `system.file("scenarios", name, package = "plant")` (with an `inst/scenarios/` fallback for `load_all`), which is the standard R-package pattern: anything under `inst/foo/` installs to the package root and is found via `system.file("foo", ...)`. It also parallels the `inst/reports/` folder this PR adds. The only other convention would be the catch-all `inst/extdata/`, but a named `inst/scenarios/` is clearer and consistent with `inst/reports/`.

Separately, addressed your baseline comment in 0544aa03 (`make bless-scenarios`).

PR #558, opened 2026-07-03
feat(tf24): add layered soil parameter API with strict depth validation

itowers1:
add set_soil_parameters(depths, soil_moist_sat, K_sat, a_psi, n_psi) to TF24 environment support per-layer parameter vectors for soil hydraulics while preserving scalar fallback behavior enforce that provided vectors must match soil layer count; error on mismatch reset layered parameter state when soil depth structure is reset wire layered parameters into infiltration, conductivity, and psi conversions expose new method through generated bindings
expand TF24 environment tests for:
per-layer setup path
NULL/default replication path
per-parameter length validation errors


dfalster:
### Updated from `develop` — CI green ✅

Merged the latest `develop` (`9b5a98ff..5a891e2c`); was previously conflicting, now `MERGEABLE` with all checks passing (lint, macOS, **windows**).

**Conflict resolution** (`inst/include/plant/models/tf24_environment.h`) — the layered soil-parameter work and develop's #549 soil-numerics fixes edit the same retention/conductivity functions, so I **combined** them rather than choosing a side:
- `soil_K_from_soil_theta`: keeps this PR's per-layer parameters (via `soil_parameter_value`, scalar fallback) **and** applies #549's clamp of `theta` to `[0, soil_moist_sat_layer]`.
- `psi_from_soil_moist`: keeps the per-layer parameters **and** applies #549's cap of `psi` at `soil_psi_max_`.

Each guard now reads the per-layer value where develop used the scalar, so the layered API and the blow-up guards coexist.

**Windows CI failure cleared**: it was the non-portable `blow-up fails gracefully (#550)` test expecting `run_scm()` to throw; develop (via #554) replaces it with a completion assertion that holds on all platforms.

No `scientific_version` bump — this is a scalar-fallback interface addition, so default simulation output is unchanged. Local full suite: 2430 pass, 0 fail.

Issue #523, opened 2026-06-24
[TF24 hydraulics] Penman-Monteith transpiration and leaf temperature


dfalster:
## Summary

Replace the Fick's-law transpiration approximation in the `tf24` model with a
**Penman-Monteith (PM)** formulation that solves the coupled leaf energy
balance and vapour diffusion algebraically, yielding physically consistent
transpiration `E` **and** leaf temperature `Tleaf` with no iteration. `Tleaf` is
then passed to the Farquhar photosynthesis model in place of `Tair`.

## Motivation

`tf24` currently computes `E = gs · VPD`. This assumes:

1. **Tleaf = Tair** — but latent-heat cooling and radiation loading can push
   Tleaf 5–10 °C from Tair under hot, high-radiation conditions (Australian
   summer).
2. **VPD evaluated at Tair** — the true driving gradient uses saturation vapour
   pressure at `Tleaf`.

Both errors propagate into the stomata optimisation (wrong water cost `E` and
wrong carbon benefit `A`, since Farquhar runs at the wrong temperature). The
`Tleaf → Farquhar` link is a *structural* bias, not noise: underestimating
Tleaf under high radiation systematically overestimates carbon gain.

## Approach

PM at the leaf scale:

```
λE = (Δ·Rn + ρ·cp·VPD/ra) / (Δ + γ·(1 + rs/ra))
E  = λE / λ
Tleaf = Tair + (Rn − λE)·ra / (ρ·cp)
```

The only change to the optimisation loop is the `E` submodel: replace
`E = gs · VPD` with the PM call (`rs = 1/gs`) and feed `Tleaf` to Farquhar. No
architectural change to the optimisation structure.

Three new, weakly-sensitive parameters are introduced (`d` leaf dimension,
`α_w` wind extinction, `U₀` above-canopy wind speed), plus a net-radiation
submodel (shortwave from absorbed PAR, longwave as a cooling offset) and an
optional exponential wind/`ra` profile through the canopy.

## Proposed tasks

- [ ] **PM core** — `transpiration_penman_monteith(Rn, VPD, ra, gs, Tair, P)`
      returning `λE`, `E`, `Tleaf`; unit-tested against known solutions.
- [ ] **Net radiation submodel** — `net_radiation_leaf(...)` using the PAR×2
      shortwave approximation and a fixed/humidity-derived longwave term.
- [ ] **Wind profile submodel** — `wind_speed_canopy(U0, L, α_w)` and
[...]

PR #567, opened 2026-07-22
[TF24 hydraulics] Penman-Monteith leaf energy balance (opt-in, default off)

dfalster:
## Summary

Adds a **Penman-Monteith leaf energy balance** to the shared TF24/TF24f `Leaf`
submodel (issue #523), **opt-in and default off** so it is fully
backward-compatible. Leaf temperature is solved from the hydraulically-pinned
transpiration and fed into the Farquhar temperature scaling, correcting the
`Tleaf = Tair` assumption that biases both carbon gain `A` and water cost `E`
under hot, high-radiation conditions.

Key design point (see `notes/penman-monteith/implementation-plan.md`): `E` stays
the Sperry hydraulic supply — PM is **not** used to re-predict `E`. Because
`A(Tleaf)` doesn't feed back into `E`, `Tleaf` is an explicit algebraic function
of `E`: one non-iterative forward pass per operating point, no new optimisation
dimension.

## What's here (Steps 1–6 of the plan)

- **Core (1–3):** Arrhenius block extracted to `update_temperature_dependent_params`; `leaf_temp_from_E()` + `es`/`Δ` helpers; `use_energy_balance_` gate wired into `set_leaf_states_rates_from_psi_stem`. TF24f uses its finite-difference gradient when PM is on (the exact AD gradient assumes temperature-independent kinetics — deferred).
- **Interface (4):** `TF24_Pars` gains `use_energy_balance` + `d`; new `wind_speed` driver; `ra = C_ra·√(d/U₀)`; the flag is R-settable. RcppR6/Rcpp bindings regenerated; NEWS breaking-changes entry; model-version snapshot re-accepted (no `scientific_version` bump — PM off is unchanged).
- **Robustness (PM-only, gated):** `psi_stem_to_ci` falls back to the compensation point when extreme heating leaves no feasible `ci`; `Tleaf` clamped to a physical range so a non-equilibrium probe can't make the Arrhenius block non-finite.
- **Evaluation (5):** factorial Fick-vs-PM (`notes/penman-monteith/`). **Verdict: keep the PM core** — ΔTleaf up to 22 °C (median ~5 °C), A collapses 90–100 % in the hot/high-radiation/high-VPD corner, negligible (<2 °C, <2 % A) in cool/low-light cells.
- **Demo (6):** staging leaf-level demo in `overstorey_staging/` (`.Rbuildignore`'d) with a fast smoke test so it can't rot.

## Backward compatibility / testing

- With the gate **off**, output is **bit-identical** to before (verified against a leaf-level grid and a full SCM run — offspring to 17 digits).
- Full `devtools::test()` suite green (2377 pass, 4 skip). New/updated: `test-pm-leaf-demo.R`, the two TF24 `Defaults` tests, and the model-version snapshot.

## Not yet done (before PM outputs are used in publications)

Doc §7.5.2–7.5.4: fixed-ra vs wind-profile sensitivity, Rn ±30 % sensitivity, and the annual SCM integration. The wind profile and longwave term remain candidates to drop if they fail their own gates.

## Related

A companion PR re-blesses the drifted `scenario_baseline.rds` and adds a regenerable reference-run generator on `chore/refresh-pre-pm-baselines` (→ develop); that drift is from earlier merged work (NSC #517, soil #558), not this PR.


inst/include/plant/leaf_model.h
```diff
@@ -82,6 +82,33 @@ const double gravity_head = 9.8e-3; // MPa / m
...
+// latent heat of vaporisation of water, J kg^-1 (fixed at 25 deg C)
+static const double latent_heat_vap = 2.45e6;
+// volumetric heat capacity of air, J m^-3 K^-1
+static const double vol_heat_cap_air = 1200.0;
+// PAR energy conversion: ~4.57 umol photons per J of PAR (shortwave)
+static const double umol_par_per_joule = 4.57;
+// shortwave absorbed ~= 2 * absorbed PAR (PAR ~= 50% of shortwave; doc 3.3)
+static const double sw_abs_per_par = 2.0;
+// fixed net longwave (cooling) offset, W m^-2 (clear-sky approximation; doc 3.3)
+static const double longwave_net_offset = -40.0;
```
itowers1, line 99: Need to investigate this further

src/leaf_model.cpp
```diff
@@ -237,25 +240,43 @@ void Leaf::set_physiology(double area_leaf, const std::vector<double>& mass_root
...
+   // PAR (PPFD_, umol m^-2 s^-1): shortwave ~= 2*PAR converted to W m^-2, plus a
+   // fixed clear-sky longwave cooling offset (doc 3.3). ra fixed for the minimal
+   // cut (doc fallback). Only read on the use_energy_balance_ path.
+   Tair_ = leaf_temp_;
+   Rn_ = sw_abs_per_par * PPFD_ / umol_par_per_joule + longwave_net_offset;
+   // Aerodynamic resistance from leaf boundary-layer theory: ra = C_ra*sqrt(d/U)
+   // (doc 4.1), with the per-strategy leaf dimension d_ and the above-canopy wind
+   // wind_speed_. Fall back to the fixed value when the wind model is unusable
+   // (non-finite / non-positive inputs), e.g. a bare Leaf that set neither.
+   ra_ = (std::isfinite(d_) && std::isfinite(wind_speed_) &&
```
itowers1, line 262: This will allow the model to work when wind speed is 0 OR NA, I understand the 0 case, but possibly just want it to fail if wind speed is NA.

inst/include/plant/leaf_model.h
itowers1: Outgoing longwave radiation is actually a function of temperature, but this would cause another root-finding operation (I believe)

notes/penman-monteith/penman_monteith_implementation.md
```diff
@@ -0,0 +1,288 @@
...
+**Longwave (thermal):**
+```
+LW_net = εa·σ·Ta⁴ - εl·σ·Tleaf⁴
+```
+
+This depends on Tleaf — which is what we are solving for. PM's linearisation handles this implicitly: it approximates LW_net ≈ f(Ta) using the slope Δ, avoiding the need to know Tleaf when computing Rn. For consistency, compute Rn using Tleaf = Tair (small error, typically < 5 W m⁻²) and rely on PM's linearisation for the remainder.
+
+In practice:
+```
+LW_net ≈ ε·σ·(εa·Ta⁴ - Ta⁴) = ε·σ·Ta⁴·(εa - 1)
```
itowers1, line 111: This could potentially be implemented in this PR or raised as an issue.

itowers1 (commented):
Couple of comments, but ready to go

src/leaf_model.cpp
dfalster: Good catch — fixed in 76df7169. `set_physiology` now errors on a non-finite `wind_speed_` (or `d_`) on the energy-balance path rather than silently using the fixed `ra`; zero wind/d (physically `ra → ∞`) stays a legitimate fallback. Added a test covering all three cases (NA → error, PM-off NA → ok, zero-wind → fallback).

inst/include/plant/leaf_model.h
dfalster: Agreed — the fixed −40 W/m² offset is a Tleaf-independent approximation, and a faithful outgoing-longwave term makes Rn implicit in Tleaf (root-find), which would break this PR's single non-iterative forward pass. Raised as #581 with a closed-form Stefan-Boltzmann linearisation option (adds a radiative-conductance term to the balance, no iteration), to be gated by the Step 5 longwave sensitivity check per the plan rather than expanding this PR. Noted at the constant in 76df7169.

dfalster: Agreed — the fixed −40 W/m² offset is a Tleaf-independent approximation, and a faithful outgoing-longwave term makes Rn implicit in Tleaf (root-find), which would break this PR's single non-iterative forward pass. Raised as #581 with a closed-form Stefan-Boltzmann linearisation option (adds a radiative-conductance term to the balance, no iteration), to be gated by the Step 5 longwave sensitivity check per the plan rather than expanding this PR. Noted at the constant in 76df7169.

notes/penman-monteith/penman_monteith_implementation.md
dfalster: Agreed — the fixed −40 W/m² offset is a Tleaf-independent approximation, and a faithful outgoing-longwave term makes Rn implicit in Tleaf (root-find), which would break this PR's single non-iterative forward pass. Raised as #581 with a closed-form Stefan-Boltzmann linearisation option (adds a radiative-conductance term to the balance, no iteration), to be gated by the Step 5 longwave sensitivity check per the plan rather than expanding this PR. Noted at the constant in 76df7169.

dfalster:
**Closing unmerged. The implementation is not being dropped — it moved.** Every commit on
this branch is in #591 (`feature/consume-leaf-package`), which was rebased onto `develop`
and carries the whole PM stack; the model code itself now lives in
[traitecoevo/leaf_cpp](https://github.com/traitecoevo/leaf_cpp), which plant consumes via
`LinkingTo`.

@itowers1 — thank you, both your comments landed. Since your review the code changed
address, so here is where each one ended up, checked rather than assumed:

**1. The `ra` fallback accepted a non-finite wind speed.** Your point: zero wind you
understood, NA should fail. Fixed in `76df7169`, and the fix **survived the move** — it is
now at `leaf_model.hpp:831` in the leaf package, still failing fast on the energy-balance
path while leaving zero wind (physically `ra → ∞`) as a legitimate fallback.

The *test* did not survive as well, which is worth flagging. It lived in
`tests/testthat/test-pm-leaf-demo.R` — a demo smoke test that sources a `.Rbuildignore`'d
helper and skips whenever `overstorey_staging/` is absent — in the repo that no longer
contains the code. It is now a C++ test in the package that owns the contract, covering
your three cases plus a fourth you did not have to ask for (`d_`, which enters the same
formula and was also unguarded).

Porting it also turned up that the *existing* PM test there was vacuous in the way that
mattered: it set `use_energy_balance_`, `wind_speed_` and `d_` on a leaf and then
immediately reassigned that leaf, discarding all three. Its own comment said "ra and Rn
are derived in `set_physiology`, so re-run it with the gate on" — and it did not re-run
`set_physiology`. It passed because 2.0 and 0.05 are also the member defaults, so nothing
ever checked that the wind model, rather than the fixed fallback, was what set `ra`. It
does now.

**2. The fixed −40 W m⁻² longwave offset.** Your two observations — that outgoing longwave
is a function of temperature, and that treating it faithfully forces another root-find —
were both right, and are why this stayed out of the PR. Tracked as
[leaf_cpp#28](https://github.com/traitecoevo/leaf_cpp/issues/28), with the
Stefan-Boltzmann linearisation (a radiative conductance in the denominator, closed-form,
no iteration) as the preferred option, still gated behind the doc §7.4/§7.5.3 sensitivity
check. plant#581 is closed and redirected there, because it cited two files this branch
deletes.

The constraint is tighter in its new home, which strengthens your instinct:
`find_root_collar_psi` runs ~10³ inner evaluations per solve and plant calls it millions
of times, so an implicit solve is a hot-path change and not merely inelegant.

---

**What "Not yet done" still means.** Doc §7.5.2–7.5.4 (fixed-`ra` vs wind-profile
sensitivity, `Rn` ±30 %, annual SCM integration) are still outstanding, and the longwave
gate above is the first of them. PM remains **default-off**, so none of it blocks #591:
with the gate off, the leaf package's golden file is blind to the PM path and plant's SCM
values are unaffected by it.

One thing that did change since this PR was written, and is worth knowing before anyone
turns PM on: `TF24_Environment`'s `atm_kpa` driver defaulted to **100.5** while the leaf's
ppm→Pa conversion was hard-coded at `0.1013`, i.e. 101.3 kPa. Those disagreed, invisibly.
#591 pins the driver to 101.3 and the leaf derives the conversion from it, so the
photosynthesis and conductance sides finally assume the same pressure — which matters for
a temperature/energy-balance model rather more than for the Fick one.

Issue #557, opened 2026-07-03
Establish and quantify expected hydraulic behaviour for TF24/TF24F (scenario gateway)


dfalster:
## Motivation

The TF24 / TF24F hydraulic strategy models (epic #424) are progressing well but
**fail under certain drought/soil conditions** — e.g. #549 (non-finite
`psi_soil` at residual soil moisture) and the growth-mortality blow-up addressed
by NSC storage (#554). To steer that work we need to (a) write down the
**expected** behaviour of the models across a range of environments and trait
syndromes, and (b) **quantify how well the current model achieves it**, so we
have concrete, re-checkable targets rather than anecdotes.

## Goal

Establish expected hydraulic behaviour for a set of scenarios and build a
rerunnable framework that scores the model against those expectations — a
**gateway check** we can re-run as features land (NSC, hyperparameterisation,
future fixes) to see the score move.

This issue tracks three pieces:

- **#572 — scoring the gateway on persistence.** The framework below is built and
  reruns, but returns no signal: it classifies on `total > 0`, so an R0 of 2e-15
  counts as persistence and only 1 of 8 scenarios actually replaces itself.
  Decides how to separate numerical viability from ecological outcome.
- **#555 — the expectations table.** A CSV of hydraulic scenarios (trait
  syndromes × environment descriptors) each labelled with an expected outcome
  (model runs / model fails). This is the specification.
- **#556 — the evaluation framework.** Translates each qualitative scenario into
  concrete parameters/environment/rainfall, runs the SCM, classifies the run
  (success/failure, with a persisted/extinct/crashed breakdown), and scores it
  against the expected outcome — producing a provenance-stamped scorecard. Adds
  an editable qualitative→quantitative mapping, a scorecard report, a runner +
  `make scenarios` target, fork-based parallelism, a content-addressed rerun
  cache, and an opt-in baseline-diff gateway test.

## Current status (baseline, pre-NSC, `max_patch_lifetime = 100`) — ⚠️ superseded

> **The 5/8 below is stale.** Re-measured 2026-07-28 at `b2ad125c` and again at the
> head of #570: the match rate is **3/8**, all 8 scenarios "succeed", none of the 5
> expected failures fails, and **only 1 of 8 persists** (R0 ≥ 1). The crashes noted
> below are gone — good news that the metric reports as a drop. See
> [the full table](https://github.com/traitecoevo/plant/issues/557#issuecomment-5103704756)
> and #572. Original text kept below for the record.
>
[...]

PR #570, opened 2026-07-28
[TF24 hydraulics] [Env drivers] Unblock site-level prediction: six TF24 fixes

dfalster:
Six fixes needed before `plant` can be run against a site and scored on observations, found by working through what it would take to reproduce a Dryland simulation from [Verbruggen et al. (2025) at the Dahra flux tower (Sahel, 416 mm MAP)](https://doi.org/10.5194/gmd-18-6623-2025) Grouped because they are all small, all in the same corner, and each was found by the previous one.

## Bit-identical (no `scientific_version` bump)

1. **`assimilation` is now reported.** The aux was declared in `aux_names()` but never written — no index resolved, no `set_aux` call — so the slot reported whatever `Internals` happened to hold and carbon uptake was unavailable as a model output. Now carries net CO₂ assimilation at the optimal operating point (µmol m⁻² s⁻¹), integrated over the crown, and is set explicitly to −R_d on both hydraulic shut-down exits. **Net, not gross** — `Leaf::assim_colimited()` subtracts R_d, and its comment claimed otherwise (also corrected).
2. **Rainfall driver floored at zero, plus `check_driver_interpolation()`.** Extrinsic drivers interpolate with a cubic spline, which is a poor fit for intermittent forcing: a realistic daily series (~10% wet days) evaluates negative at **45% of points, reaching −5.7 m yr⁻¹**. The spline conserves the integral exactly, so the problem is invisible in a total-rainfall check — hence the diagnostic. *Partial:* the floor bounds the sign only; a step/piecewise-constant interpolation mode is the real fix, since daily forcing *is* a step function.
3. **Multi-layer water-budget conservation tests** at 1, 5 and 15 layers with uptake included, saturated to dry, and at and below the residual floor. Closure holds to round-off (relative residual < 1e-12) in every case. Coverage was previously single-layer and without uptake (it ran 0.01 yr). The hypothesised `soil_moist_residual` / ψ-clamp leak **does not exist**: with `n_psi = 6.57` the conductivity exponent is ~16, so K(0.05) < 1e-10 against K_sat = 163 and the floor is effectively unreachable dynamically.
4. **`root_b`, `root_c`, `root_psi_crit`, `rooting_depth_max` promoted to `TF24_Pars`** (defaults unchanged). Root shutoff was pinned at ψ ≈ 5.87 MPa — too conservative for taxa operating below it, e.g. *Acacia aneura* — and rooting depth at 1.5 m. Verified live: with dry shallow over wet deep layers, a 0.2 m cap settles at ψ_collar −7.09 MPa against −5.87 for 1.5 m.

All four leave `offspring_production` identical to the bit on all 8 gateway scenarios.

## Changes water-limited output (`TF24@v4`, `TF24f@v4.1`)

5. **A hydraulically shut-down plant no longer draws water from the soil.** Both shut-down exits in `Leaf::find_root_collar_psi` set `profit_` directly and bypass `profit_psi_stem_TF`, and the first returns before any `E_from_Soil_to_Root_Collar` call. Since `Leaf` is a value member reused across every `compute_rates` call, everything they did not assign kept the **previous step's** value.

   This was not a reporting problem. `soil_consumption_` feeds `evapotranspiration_dt` and hence the patch water balance, so **a plant with closed stomata carried on drawing its last wet-step uptake out of the soil.** Measured: an individual moved from θ = 0.30 to θ = 0.02 (past ψ_crit) reported `E_up_` and `transpiration` identical to its wet step (2.7145e-05) while `assimilation` correctly collapsed to −R_d. The budget still *closed* in that state — what was recorded as depleted was what was removed — so the conservation tests in (3) could not catch it. Only the physics was wrong.

   Both exits now zero the transport chain, with recovery on rewetting tested. Gateway: R0 moves ≤ 5e-3 relative on 5 of 8 scenarios, every classification unchanged.

## Robustness and honest reporting (only affects states that previously threw)

6. **A dry TF24f patch no longer aborts the whole run on the ci root-find.** `Leaf::dprofit_droot_collar_psi` called `psi_stem_to_ci()` *before* testing for shut-down. In shut-down `psi_upstream >= psi_stem` (both positive magnitudes), so `gc = const · transpiration` goes negative, the residual stops crossing zero over (Γ*, c_a], and TOMS748 **throws** `a and b do not bracket the root` rather than returning non-finite — which is exactly what the `isfinite` guard on the next line was written to catch and cannot see. One dry patch killed the run. The condition is now tested before the solve, as `set_leaf_states_rates_from_psi_stem()` has always done. Reproduced at 5 layers, θ = 0.005–0.03, 1 m yr⁻¹ (ψ_stem = 1.23 vs ψ_upstream = 5.92 MPa).

   **Adaptive interpolation no longer blames resolution for a NaN.** `check_err()` compares against NaN, and every NaN comparison is false, so one non-finite target value made its interval permanently unacceptable: refinement ran to `max_depth` and reported *"Interpolated function as refined as currently possible"*. That message cost real debugging time here. Non-finite now fails immediately naming the point and value; the resolution message reports the spacing reached, the limit, and the tolerances missed. New `test_adaptive_interpolator()` hook drives the refiner from R (the production caller passes a C++ lambda).

   **The scorecard now reports `persists` (R0 ≥ 1).** `classify_scm_run()` tests `total > 0`, which is far weaker than it looks: five of eight scenarios return R0 between 2e-15 and 6e-14 — numerically extinct — and all were recorded as `"persisted"`. **Only 1 of 8 scenarios persists.** That is the main reason the gateway discriminates so little (3/8, with no expected failure failing, and #557 still quoting a stale 5/8). Reported *alongside* `status`/`outcome` rather than replacing them: the CSV's "Model failure" means the model breaks numerically, a different axis, and the blessed baseline diff is defined on the existing columns.

   **Gateway seasonal driver knots are placed per year (48 yr⁻¹)** rather than per run, so realised seasonality no longer depends on `max_patch_lifetime`. Measured before changing: 6 knots/cycle was accurate to 5.4e-3 on a peak of 3.0 (0.18%) and conserved the annual total to 1e-5 %, so this is robustness rather than the bug it was suspected to be. It does move R0 by up to **6.2%** relative — demography amplifies driver error ~30× here — with every classification unchanged.

## What this does *not* fix

Once (6) is guarded, TF24f reaches the same wall TF24 already hit. When water limitation stalls growth, every cohort sits at the introduction height (0.392 m, `transpiration` = 0, `assimilation` = −R_d, `opt_psi_stem` pinned at the 5.92 MPa shutoff, `mortality` = Inf) and the light-availability profile develops a step narrower than the refiner's depth limit (~1.5e-5 m over a 15.7 m domain). It is non-monotone in every axis — fails at θ₀ = 0.005–0.03 but not 0.001 or ≥ 0.05, at 3 and 5 layers but not 1, 2, 8, 10, 15, 20, at 0.8/1.0/2.0 m yr⁻¹ but not 1.5 — so a knife-edge, not a threshold. This is the #550 family (the size distribution collapsing to a delta), and it is a hard blocker on any dryland site run. Filed separately.

Also found and not fixed: the `Leaf` API throws in the dry states the SCM reaches routinely — `find_root_collar_psi()` gives `invalid f_ri` at ψ_soil = 4 MPa and `find_psi_stem_from_psi_root()` leaves the transport spline's domain above ~3.7 MPa — so the hydraulics cannot be unit-tested where it matters most. That is why (6)'s regression test goes through the SCM rather than a bare `Leaf`.

## Verification

- Full suite **2540 passing, zero failures** (2430 before this branch).
- Scenario gateway re-run: **all 8 classifications unchanged**, so the blessed baseline needs no re-blessing. Score 3/8, `n_persists` 1/8.
- New test files: `test-tf24-water-budget.R`, `test-tf24-shutdown-state.R`, `test-tf24-assimilation-aux.R`, `test-tf24-root-pars.R`, `test-tf24-arid-corner.R`.

Refs #549, #550, #557.


dfalster:
Hi @aornugent @itowers1  - FYI - this is first of three stacked PRs (#570, #573 and #574) all linked above, which go most of the way towards removing issues running stands at low soil water (and then associated density blowups). 

they should be merged in order. I've requested reviews from Isaac

src/leaf_model.cpp
```diff
@@ -894,10 +924,24 @@ double Leaf::dprofit_droot_collar_psi(double opt_root_psi) {
...
+  // transpiration. It has to be caught *here* because psi_stem_to_ci() does not
+  // return non-finite in that state, it throws: gc = const * transpiration goes
+  // negative, which flips the sign of the supply term so the residual no longer
+  // crosses zero over (gamma*, ca] and TOMS748 reports "a and b do not bracket
+  // the root". The isfinite check below was written to cover shut-down but
+  // cannot see a thrown exception, so a dry patch killed the whole run:
+  // reproduced on TF24f at 5 layers, theta = 0.005-0.03 with 1 m/yr rainfall,
+  // at psi_stem = 1.23 against psi_upstream = 5.92 MPa.
+  if (!std::isfinite(psi_stem) || psi >= psi_stem) {
     return 0.0;  // shut-down / infeasible: no informative gradient
```
itowers1, line 939: These changes will allow it to run, but worried about whether we essentially get an undisclosed error when the model should really just break.

dfalster: Good instinct, and it turns out to be half right — but the undisclosed thing is
downstream, not here. Some measurements.

**This gradient is not consumed by a root-find.** `TF24f_Strategy::compute_rates`
does

```cpp
vars.set_rate(state_idx_opt_root_psi_state, k_acclim * dprofit_dpsi_);
```

so it is the *rate* of an ODE state — TF24f tracks rather than optimises. Returning 0
therefore freezes the tracked collar psi while there is no flow; it does not
fabricate a converged optimum, which is what would be genuinely undisclosed. The
finite-difference branch in the same function already does exactly this for the
analogous infeasible case (`dprofit_dpsi_ = 0.0; return;`, "no interior interval to
perturb in"), so the AD branch returning 0 makes the two gradient methods agree
rather than inventing a behaviour.

**How often it fires** (instrumented the branch, counted, then reverted):

| TF24f, 10 yr | calls | shutdown | non-finite | fraction |
|---|---|---|---|---|
| default env | 1,380,000 | **0** | 0 | 0.0000 |
| dry start, θ = 0.005 | 3,520,000 | 19,403 | 0 | 0.0055 |

So it is never reached in a well-watered run — nothing is being masked in normal
operation — and in a dryland run it is 0.55% of gradient evaluations: a transient
regime individual cohorts pass through, not the whole run. "The model should just
break" would mean aborting an entire dryland site because 0.5% of gradient
evaluations found a plant in hydraulic shutdown, which the rest of the model already
represents as zero transpiration (`set_leaf_states_rates_from_psi_stem`).

Also worth noting `non-finite = 0` in both: the pre-existing `!isfinite(psi_stem)`
guard never fires in practice. As the comment says, it was written to cover shut-down
but cannot see a thrown exception, so before this PR the intended guard was
effectively unreachable and the actual behaviour was a TOMS748 "a and b do not
bracket the root" abort from three layers down. That is not the model breaking
honestly either — it names nothing about hydraulic shutdown. So the choice was never
"silent 0 vs honest break", it was "defined 0 vs a misleading numerical abort".

**Where you are right.** There *is* an undisclosed failure, and we found it while
working #571 on top of this branch: hydraulic shutdown makes the height growth rate
`g <= 0` at introduction, and `Node::compute_initial_conditions` then sets
`log_density = log(0.0)` — density exactly zero, pinned there permanently. In the dry
start, 59 of 93 cohorts are born inert and then grow to ~15 m anyway, because height
growth does not depend on density. The run *completes* and looks plausible from the
outside. That is filed as #575 with the measurements. Making this line throw would
not fix it, it would only restore the crash.

**Concrete suggestion, if you want disclosure at this level.** Shutdown is currently
inferable from the aux outputs (`transpiration`, `stom_cond_CO2`, `E_up` all 0) but is
not flagged, so nobody looks. Exposing it as a first-class aux — a shutdown flag, or
the fraction of the step spent shut down — would make "was this plant shut down?"
directly answerable instead of an inference. Happy to add that here or as a follow-up;
it is a genuine gap and it is what "undisclosed" is really pointing at.

dfalster: Daniel framed the general principle behind this, and it wasn't written down anywhere in the
family, so it now is: **traitecoevo/plant-meta#2** (`governance/model-robustness.md`).

> Models should run across a wide range of plant strategies and environmental conditions and
> produce sensible biological outputs — including zero growth or reproduction — rather than
> crash. But where a state is physically unrealistic, the model should fail.

Which puts this thread's question on a rule rather than a case-by-case argument: hydraulic
shutdown is physically attainable, so it gets represented (this PR); a cohort at exactly zero
density growing to 15 m is not, so it should fail or be flagged rather than pass silently
(#575). Your "should really just break" instinct is the second half of that principle, pointed
at the right problem one layer downstream.

The doc also takes on the part of your comment I think is most durable — that returning a value
is only acceptable if the state is *observable*, since one that can only be inferred from
`transpiration == 0` is not actually disclosed.

dfalster: Correction to my two replies above — Daniel has pointed out I misread the zero-density case, so
strike the part where I said the "undisclosed error" you were sensing lives downstream in #575.

A cohort at exactly zero density whose trajectory reaches 15 m is the method of characteristics
working as intended: the cohort *is* a trajectory, and the density carried along it is a
separate quantity. It says a plant germinating at that moment would by now be 15 m, and none
did. `density = 0` is not a hidden failure, it is the disclosure that nothing established. I
have retracted #575 on that basis.

What survives from that exchange is narrower, and it is the part of your comment I think was
always the durable bit: **if a guard returns a value instead of failing, the state has to be
observable in the outputs.** Hydraulic shutdown currently satisfies that only by inference —
you have to notice that `transpiration`, `stom_cond_CO2` and `E_up` are all zero — which in
practice means nobody looks. That is a real gap and the offer stands to expose it as a
first-class aux (a shutdown flag, or the fraction of the step spent shut down), here or as a
follow-up.

The rest of my argument for `return 0.0` at this line is unaffected: it is an ODE rate rather
than a root-find target, it matches the finite-difference branch, it fires on 0 of 1.38M
gradient evaluations when well watered versus 0.55% when dry, and the prior behaviour was a
TOMS748 bracketing abort that disclosed nothing about shutdown.

Issue #571, opened 2026-07-28
[TF24 hydraulics] Dry-start runs die in the light spline once stalled growth piles cohorts at one height


dfalster:
Split out of #570, where the first-order cause was fixed and this one was left. **This is the hard blocker on running `plant` at any dryland site.**

## Symptom

```r
env <- Environment("TF24")
env$set_soil_number_of_depths(5)
env$set_soil_water_state(rep(0.005, 5))
env$extrinsic_drivers_set_constant("rainfall", 1.0)
p <- scm_base_parameters("TF24")
p$max_patch_lifetime <- 10
p <- add_strategies(p, trait_matrix(0.0825, "lma"))
run_scm(p, env, collect = TRUE)
#> Error: Interpolated function as refined as currently possible: spacing 0.000007
#>   is below the limit 0.000015 set by max_depth=16, and the target still misses
#>   atol=0.000100 / rtol=0.000100. The target has a feature narrower than that spacing.
```

Same on TF24f (after #570 — before it, TF24f failed earlier on the ci root-find, which was masking this).

## It is not a soil bug, and not a threshold

Originally suspected as "rewetting from below the residual floor". Measured, it is neither. The failure is **non-monotone in every axis**, at 5 layers / 10 yr / 1 m yr⁻¹ unless stated:

| swept | fails | runs |
|---|---|---|
| θ₀ | 0.005, 0.008, 0.0099, 0.010, 0.0101, 0.012, 0.015, 0.02, 0.03 | 0.001, 0.05, 0.10, 0.20, 0.25, 0.30, 0.428 |
| layers (θ₀ = 0.005) | 3, 5 | 1, 2, 8, 10, 15, 20 |
| rainfall (θ₀ = 0.005) | 0.8, 1.0, 2.0 | 0, 0.05, 0.1, 0.2, 0.4, 1.5 |
| lifetime (θ₀ = 0.005) | 10 | 2, 4, 5, 6, 7, 8, 9 |

It straddles `soil_moist_residual` (0.01) rather than sitting below it, and the *driest* start runs fine. A knife-edge, not a physical threshold.

## Cause

The failing interpolation is the **light-availability spline**, not anything in the soil column. At the last lifetime that completes (9 yr), the patch state is degenerate:

- every cohort is at **height 0.392 m** — the introduction height. Nothing has grown.
- `transpiration` = 0, `stom_cond_CO2` = 0, `E_up_` = 0
[...]

PR #574, opened 2026-07-28
Build the competition profile over a sorted node grid when heights invert (#571)

dfalster:
Now targets `develop` directly — #570, #573 and #580 have all merged, and this is rebased onto
them (two commits, two files). **This is the fix for the crash in #571**; #573 was the
diagnostic that found the real cause.

## The bug

`Species::compute_competition()` uses the node list itself as the quadrature grid for the
trapezium integral over the size axis, and `Species::height_max()` returned `nodes.front()`.
Both rely on nodes being ordered largest to smallest.

That ordering used to be free. With `dh/dt` a function of height and the shared environment,
growth trajectories cannot cross, so cohorts introduced in order stay in order — which is why it
was asserted in a comment rather than maintained.

**TF24 broke it.** Reserve-gated growth (#517) makes `dh/dt` depend on a cohort's own storage, so
two cohorts born moments apart into a rapidly wetting soil start from different operating points
and the later one can overtake. Measured on the #571 dry start: introduction times strictly
increasing (0.0195 → 0.0234) but final heights *not* (15.6202 → 15.6716). 18 inversions, and
`height_max()` reporting 14.9385 when the tallest — and only living — cohort is at 14.9691.

Consequences: a light spline truncated 0.03–0.10 m below the tallest plant, and an early exit
firing at the first node below the query height that dropped every node past it, including the
one cohort with appreciable density. That is the fictitious step the refiner was dying on.

## The fix

One fused `scan_heights()` pass per call gets both the tallest height and whether the heights are
ordered; when they are not, fall back to the same trapezium integral over a height-sorted view.

Zero-density nodes deliberately stay in the integral. `f = 0` at such a node is the
reconstruction saying density vanishes at that size; dropping those grid points would interpolate
live density straight across the band and overestimate it. So the grid gets sorted, not filtered.

The scan is cached and invalidated by the paths that can change a height (`set_ode_state` — now
shadowed rather than inherited — both `introduce_new_node` overloads, `clear()`,
`r_set_heights()`), because `compute_competition()` is called once per spline knot, i.e. hundreds
of calls between two changes to those heights.

Gating instead on *"a node was introduced at zero density"* was considered and rejected: density
never enters the height ODE, so a zero-density node is not what makes cohorts cross —
state-dependent growth is. The two are correlated in the #571 patch only because the inert
cohorts happen to be the ones bunched at the ceiling. Using density as the trigger would restore
the silent wrongness for the one case that cannot be ruled out, two *live* cohorts crossing.

## Verification

- **Equivalent, not a change of method.** Forcing the sorted path for *every* call in *every*
  model leaves the full suite and the TF24 scenario gateway passing. The early exit is a pure
  optimisation, so the ordered case is unchanged and only states that were previously wrong move.
- **No cache staleness.** With a temporary assertion that the cached scan always equals a freshly
  computed one, the full suite and the gateway pass with none detected — so the invalidation
  coverage is measured, not assumed. That mattered because a stale cache would silently
  reintroduce exactly the profile this PR fixes.
- **No baseline re-bless needed.** `PLANT_RUN_SCENARIOS=1` gateway passes against the blessed
  scorecard; TF24 bench `offspring_production` unchanged at `0.0001899296527`.
- **The crash is gone.** All nine θ₀ values recorded as failing in #571 run, and a 94-configuration
  sweep across all four axes of that issue's failure table (θ₀ 0–0.428, layers 1–20, rainfall 0–2,
  lifetimes 2–30, both TF24 and TF24f) reports no resolution failure anywhere.
- Suite green at 2556 pass / 0 fail after the rebase onto current `develop`.

Tests: `height_max()` equals the tallest cohort while the front node does not (the precise
regression), the dry start completes, and the whole non-monotone θ₀ failure set runs. Both new
tests `skip_if` the pathological state stops occurring, so they cannot silently become vacuous.

## Cost

The scan replaces an `O(1)` `nodes.front()`. Cached, it is within noise of baseline; uncached and
fused it was +2.5% / +2.7%, and as two separate passes twice that (interleaved A/B, sampling off,
per `profile-plant`).

| state | scm 20-rep (s) | build_schedule 20-rep (s) |
|---|---|---|
| before | 2.042 / 2.056 / 2.091 | 5.883 / 5.850 / 5.997 |
| uncached scan | 2.123 / 2.076 | 5.999 / 6.054 |
| **cached scan** | **2.023 / 2.028** | **5.796 / 5.814** |

## What this does *not* fix

- **#576** — one dry TF24f configuration (rainfall 0.04–0.05) still fails, with a *different*
  mechanism: a hydraulic spline evaluated outside its domain in `leaf_model.cpp`. So "TF24f cannot
  crash when dry" is not yet true.
- A latent numerics tension: characteristics legitimately converge to within ~2.6e-5 m near a
  water-limited ceiling, against the light refiner's 1.5e-5 m floor. This PR makes the integral
  correct regardless of ordering, so it is latent rather than removed.

Note #575 (zero-density cohorts) was **closed as invalid** — that behaviour is the method of
characteristics working as designed, per Daniel's correction; see
[plant-meta#3](https://github.com/traitecoevo/plant-meta/pull/3).

### DIFF
--- a/inst/include/plant/species.h
+++ b/inst/include/plant/species.h
@@ -35,7 +35,6 @@ public:
   // ODE plumbing and the per-element serialisers are inherited from SpeciesBase
   // and iterate all nodes (the deterministic model has no notion of "dead").
   using base_type::ode_size;
-  using base_type::set_ode_state;
   using base_type::ode_state;
   using base_type::ode_rates;
   using base_type::get_node_state;
@@ -62,7 +61,17 @@ public:
   // twice was measurably slower on FF16 (~5% on the SCM benchmark) than the
   // O(1) nodes.front() it replaced.
   struct HeightScan { double h_max; bool decreasing; };
+  // Cached: heights change only when the ODE state is set or a node is
+  // introduced/cleared, whereas compute_competition() is called once per spline
+  // knot, so this is hundreds of calls per change. Every mutator invalidates.
   HeightScan scan_heights() const;
+
+  // Setting the ODE state rewrites every node's height, so the cached scan goes
+  // with it. Shadows (rather than uses) the SpeciesBase version for that reason.
+  odelia::ode::const_iterator set_ode_state(odelia::ode::const_iterator it) {
+    invalidate_height_scan();
+    return base_type::set_ode_state(it);
+  }
   void compute_rates(const environment_type& environment, double pr_patch_survival, double birth_rate);
   std::vector<double> net_reproduction_ratio_by_node() const;
   // Per-node lifetime offspring, weighted by patch-age density and S_D.
@@ -133,6 +142,16 @@ private:
   // ordered, so the node list cannot be used directly as the quadrature grid.
   double compute_competition_unordered(double height) const;
 
+  // Cache for scan_heights(). Every path that can change a node height must call
+  // invalidate_height_scan(); a stale cache here would silently reintroduce the
+  // wrong competition profile of #571, so the coverage of these calls was checked
+  // by asserting cache == freshly-computed on every call across the whole suite
+  // and the scenario gateway.
+  HeightScan compute_height_scan() const;
+  void invalidate_height_scan() { height_scan_valid = false; }
+  mutable HeightScan height_scan_cache{0.0, true};
+  mutable bool height_scan_valid = false;
+
   // Storage (strategy, nodes) and control() live in SpeciesBase; the
   // using-declarations let the unqualified references below resolve through the
   // dependent base.
@@ -158,6 +177,7 @@ size_t Species<T,E>::size() const {
 
 template <typename T, typename E>
 void Species<T,E>::clear() {
+  invalidate_height_scan();
   nodes.clear();
   // Reset the new_node to a blank new_node, too.
   new_node = node_type(strategy);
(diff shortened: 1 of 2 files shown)


dfalster:
---

**Update: the ordering check is now free.**

`scan_heights()` is cached and invalidated from the places that can change a height
(`set_ode_state`, both `introduce_new_node` overloads, `clear()`, `r_set_heights()`),
since `compute_competition()` is called once per spline knot — hundreds of calls
between two changes to those heights.

Gating instead on *"a node was introduced at zero density"* was considered and
rejected. Density never enters the height ODE, so a zero-density node is not what
makes cohorts cross — state-dependent growth is (#517's reserve gate). The two are
correlated in the #571 patch only because the inert cohorts happen to be the ones
bunched at the ceiling. Using density as the trigger would restore the silent
wrongness for the one case that cannot be ruled out: two *live* cohorts crossing.

A stale cache would silently reintroduce the very profile this branch fixes, so the
invalidation coverage was measured, not assumed: with a temporary assertion that the
cache always equals a freshly computed scan, the full suite (2556) and the
`PLANT_RUN_SCENARIOS=1` gateway pass with no staleness detected.

| state | scm 20-rep (s) | build_schedule 20-rep (s) |
|---|---|---|
| before | 2.042 / 2.056 / 2.091 | 5.883 / 5.850 / 5.997 |
| uncached scan | 2.123 / 2.076 | 5.999 / 6.054 |
| **cached scan** | **2.023 / 2.028** | **5.796 / 5.814** |

So the +2.5% / +2.7% is gone — the cached version is within noise of baseline.
`offspring_production` unchanged at `0.0001899296527`.

itowers1:
Seems fine to me, and makes sense. I can picture the scenario where a seedling rapidly becomes taller than the previous, poorly growing cohort.

Issue #527, opened 2026-06-24
[AutoDiff] [acclimation] Smooth/analytic profit gradient for TF24f leaf-state tracking


dfalster:
Arising from #525 (TF24f, PR #526).

The TF24f prototype tracks the optimal root-collar ψ via gradient ascent,
`dψ/dt = k · ∂profit/∂ψ`, using a **finite-difference** gradient. The k-sweep
showed this gradient is the binding constraint:

- The optimal gain is *low* (`k ≈ 2`). Increasing `k` makes the system **stiffer**
  (run time climbs and the run becomes unintegrable beyond `k ≈ 20`) **and less
  accurate** — the finite-difference step sits at the scale of the leaf solve's
  own internal tolerances (GSS/Newton ≈ 1e-3), so its noise is amplified into the
  ODE right-hand side and into the dynamics.
- The result is an **accuracy floor** (median |Δheight| ≈ 5 mm, max ≈ 0.9 m,
  ~93 % within 1 % vs TF24) that cannot be removed by tuning `k`: "large k → TF24"
  does not hold.

The original TF24 optimiser sidesteps this because its golden-section argmax is a
*smooth* function of inputs (the demographic growth-rate gradient already relies
on that). Our FD gradient is not smooth.

### The challenge
Provide a **smooth / exact gradient** of leaf carbon profit w.r.t. the
root-collar potential, via automatic differentiation, so that:
- the ODE right-hand side is smooth → the adaptive solver is stable at higher `k`
  without a large FD step;
- the accuracy floor drops (closer to TF24, ideally to solver tolerance);
- the per-step cost stays low (one AD pass rather than ≥2 perturbed solves).

### Notes / dependencies
- Ties into the existing AutoDiff work (#491 / odelia) — the leaf profit path
  (`evaluate_root_collar_psi` → `find_psi_stem_from_psi_root` →
  `profit_psi_stem_TF` → `psi_stem_to_ci`, with spline lookups) would need to be
  differentiable end-to-end, including the inner root-find for ψ_stem.
- Open question: differentiate through the inner ψ_stem root-find (implicit-
  function theorem) vs. also tracking ψ_stem as a second state (the two-state
  variant noted in #525).
- Acceptance check: re-run the #525 k-sweep with the AD gradient and confirm a
  usable high-`k` window with a markedly lower height-difference floor.

dfalster:
**Speed benchmark (profile-plant, same machine, sampling off)**

[...]

PR #582, opened 2026-07-30
Don't ask for a gradient in hydraulic shutdown

### LINKED ISSUE #576: [TF24 hydraulics] TF24f: hydraulic spline evaluated outside its domain in a narrow dry rainfall window
Found while verifying that #574 clears the dry-corner crashes in #571. It does, for
93 of 94 configurations swept — this is the one that still fails, and it is a
**different mechanism**: a hydraulic spline evaluated outside its domain, not the
light-availability resolution failure of #571.

## Reproducer

```r
env <- Environment("TF24f")
env$set_soil_number_of_depths(5)
env$set_soil_water_state(rep(0.005, 5))
env$extrinsic_drivers_set_constant("rainfall", 0.05)
p <- scm_base_parameters("TF24f")
p$max_patch_lifetime <- 10
p <- add_strategies(p, trait_matrix(0.0825, "lma"))
run_scm(p, env, collect = FALSE)
#> Error: Extrapolation disabled and evaluation point outside of interpolated domain.
```

Deterministic, and TF24f-only — the same configuration on TF24 runs.

## It is a narrow window, again

| rainfall (m/yr) | 0.03 | 0.04 | 0.05 | 0.06 | 0.07 | 0.1 | 0.2 | 0.4 | 0.8 | 1.0 | 1.5 | 2.0 |
|---|---|---|---|---|---|---|---|---|---|---|---|---|
| TF24f | runs | **FAIL** | **FAIL** | runs | runs | runs | runs | runs | runs | runs | runs | runs |

Note rainfall **0** runs. So this is not "drier fails" — it is a knife-edge in the
middle, like the #571 sweep was. The rest of the sweep (θ₀ over 19 values including 0,
layers 1–20, lifetimes 2–30, both models = 94 runs) is clean.

## Where

Every `.eval()` that can raise this lives in the leaf hydraulics
(`src/leaf_model.cpp`): `root_vuln_from_psi`, `root_vuln_integral_from_psi`,
`transpiration_from_psi`, `psi_from_transpiration`. The most likely site is
`Leaf::transpiration_to_psi_stem` (leaf_model.cpp:1197), whose last act is
`psi_from_transpiration.eval(E_psi_stem)` with no domain guard on `E_psi_stem`.

This is the brittleness already flagged in a comment in
`tests/testthat/test-tf24-arid-corner.R`:

> `find_root_psi` throws "invalid f_ri" at psi_soil = 4 MPa, and
> `find_psi_stem_from_psi_root` leaves the transport spline's domain above
> ~3.7 MPa — brittleness worth its own issue, but not a place to anchor a
> regression test.

This is that issue.

## Worth deciding

- Should these lookups clamp to the spline domain, or should the domain be widened to
  cover the whole reachable psi range? Clamping is a physical statement (conductivity
  is ~0 out there), so it may be the honest choice rather than a fudge.
- Either way the error should name the spline, the evaluation point and the domain.
  #573 did this for the light spline; the same argument applies here, and would have
  turned this into a one-line diagnosis instead of a bare message with no location.

## Related

- #571 / #574 — the light-spline failure in the same corner, different mechanism
- #570 — fixed the `ci` root-find abort that used to mask failures here

dfalster:
TF24f's two gradient methods disagreed. The FD branch checks what
`prepare_collar_solve()` returned and takes a zero gradient when the operating
point was forced by feasibility handling. The AD branch — the default — went
through `evaluate_root_collar_psi()`, which hides that return value, and asked
for a gradient at a collar potential the feasibility analysis had already
rejected. In shutdown that amounts to asking for a ψ_stem wetter than
saturation, which does not exist.

The AD branch now mirrors the FD branch. The spline error stays, neither clamped
nor widened: hydraulic shutdown is attainable and is represented as a zero
gradient, but a ψ_stem wetter than saturation is not attainable and has to fail.
The four throwing transport lookups now name the spline, the point, how far
outside it fell, the domain and the caller.

Behaviour changes only in states that previously threw, so the TF24f scientific
version does not move.

Closes #576.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -309,6 +309,32 @@ were not previously recorded here:
 
 ### Minor changes & bug fixes
 
+* **TF24f no longer asks for an acclimation gradient in hydraulic shutdown**
+  (#576). In a narrow dry rainfall window (0.04–0.05 m/yr at θ = 0.005, five
+  layers) TF24f died with "Extrapolation disabled and evaluation point outside of
+  interpolated domain" while TF24 ran. The cause was an asymmetry between the two
+  gradient methods in `TF24f_Strategy::solve_leaf`, not a numerical limit: the
+  finite-difference branch checked `prepare_collar_solve()`'s return value and
+  took a zero gradient when the operating point had been *forced* by feasibility
+  handling, while the AD branch called `evaluate_root_collar_psi()` — which hides
+  that return value — and then asked for a gradient regardless. In shutdown the
+  collar potential is set to `-root_psi_crit`, where soil uptake is negative, so
+  the stem potential that would carry it is wetter than saturation and the
+  transport inverse genuinely has no solution. The out-of-domain error was
+  correct; the question was wrong. The AD branch now mirrors the FD branch, which
+  also drops a redundant re-derivation of the soil-side caches per step. Behaviour
+  changes only in states that previously threw, so the TF24f scientific version
+  does not move.
+* **The leaf hydraulic transport splines say which spline, point and domain
+  failed** (#576). `transpiration_from_psi` and `psi_from_transpiration` are the
+  only interpolators in `leaf_model.cpp` with extrapolation disabled, and odelia's
+  out-of-domain error names none of the three, so localising #576 meant bisecting
+  four call sites by hand. Lookups now report the spline, the evaluation point,
+  how far outside it fell and which call was asking; a non-finite point is
+  reported as such rather than as a too-narrow domain (the misdirection #573
+  removed from the adaptive interpolator). Failures inside the collar solve also
+  carry the bracket they were working within and which feasibility branch
+  `prepare_collar_solve()` exited through.
 * **The TF24 rainfall driver is floored at zero.** Because drivers are
   interpolated with a cubic spline, an intermittent series undershoots below
   every supplied value, and negative rainfall gave negative infiltration and an
--- a/inst/include/plant/leaf_model.h
+++ b/inst/include/plant/leaf_model.h
@@ -13,6 +13,9 @@
 #include <plant/uniroot.h>
 #include <plant/optimize.h>
 
+#include <limits>
+#include <string>
+
 namespace plant {
 
 // double check best namespace for constants (private vs global)
@@ -307,6 +310,20 @@ public:
   // false when the operating point is already fully determined (caller is done),
   // true when there is a real interval to choose a collar potential within.
   bool prepare_collar_solve(double& bound_a, double& bound_b);
+  // The two collar potentials prepare_collar_solve derived the feasible interval
+  // from (signed, <= 0), kept so a failure anywhere downstream of it can say
+  // which bracket it was working inside -- see collar_solve_context(). Only
+  // meaningful for the step whose prepare_collar_solve most recently ran.
+  double root_crit_ = std::numeric_limits<double>::quiet_NaN();
+  double root_zero_E_ = std::numeric_limits<double>::quiet_NaN();
+  // Which feasibility branch prepare_collar_solve exited through, or nullptr if
+  // it returned a real interval. Reported by collar_solve_context(), and the
+  // reason a caller can tell "the operating point was chosen by the optimiser"
+  // from "the operating point was forced by shutdown".
+  const char* collar_exit_ = nullptr;
+  // Human-readable dump of the collar-solve bracket and the soil state it came
+  // from, appended to errors raised inside the solve (#576).
+  std::string collar_solve_context() const;
   // Evaluate the leaf at a *given* root-collar potential (positive magnitude)
   // rather than optimising it: reuses prepare_collar_solve, clamps the target to
   // the feasible interval, and evaluates there (no golden-section search). Leaves
--- a/src/leaf_model.cpp
+++ b/src/leaf_model.cpp
@@ -24,6 +24,39 @@ template <typename T>
 T hydraulic_cost_ad(T psi_stem, double b, double c, double g1, double beta2) {
   return g1 * pow(1.0 - exp(-pow(psi_stem / b, c)), beta2);
 }
+
+// Domain-guarded lookup on the two xylem transport splines, which are the only
+// interpolators in this file built with extrapolation DISABLED (the root
+// vulnerability pair clamps instead -- see setup_root_vulnerability). odelia's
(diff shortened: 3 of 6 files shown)


dfalster:

---

Fixes #576.

## What it turned out to be

Not a too-narrow spline domain. The issue's guess (`Leaf::transpiration_to_psi_stem`) is where it *surfaces*, but both halves of the reasoning were wrong:

- it fails at the **lower** end of `psi_from_transpiration`, not the upper end of the vulnerability curve — the domain starts at ψ = 0 and the evaluation point is **negative**, so widening would not have helped;
- the lookup is fine. It is being asked a question with no answer, from a state the feasibility analysis had **already rejected**.

The cause is an asymmetry between TF24f's two gradient methods. The FD branch checks `prepare_collar_solve()`'s return value and takes a zero gradient when the operating point was *forced* by feasibility handling. The AD branch — the default — called `evaluate_root_collar_psi()`, which hides that return value, then asked for a gradient regardless:

```cpp
leaf.evaluate_root_collar_psi(tracked_root_psi_);
const double used = -leaf.root_collar_psi_;
dprofit_dpsi_ = leaf.dprofit_droot_collar_psi(used);   // at a rejected collar potential
```

In shutdown, `set_shutdown_state(-root_psi_crit)` makes `used` come back as `root_psi_crit`, where soil uptake is **negative** (three of four layers at the residual floor, ψ ≈ −1000 MPa, so water flows *out* of the roots). The ψ_stem that would carry a negative flux from a collar at 5.87 MPa is wetter than saturation. No such ψ_stem exists.

That also explains the knife-edge: the window is where the soil is dry enough to trip the `E_column(-psi_crit) < 0` shutdown test but not the earlier `-wettest >= psi_crit` one. Rainfall 0 misses it because the whole profile is at the floor and the *first* test fires.

## Which side of the model-robustness table

Both halves, in different places — per [`model-robustness.md`](https://github.com/traitecoevo/plant-meta/blob/main/governance/model-robustness.md), which names this issue:

- **Hydraulic shutdown is attainable → represent it.** Gradient zero; there is no interval to move within. Same call #570 already made *inside* `dprofit_droot_collar_psi`.
- **A ψ_stem wetter than saturation is not attainable → fail.** So the spline error was correct and **stays** — neither clamped nor widened. Clamping would have returned a plausible ψ_stem for an impossible state, which is the failure mode the doc says costs the most.

## Changes

- `TF24f_Strategy::solve_leaf`'s AD branch now mirrors the FD branch. Incidentally drops a redundant second `prepare_collar_solve` per step (the old form re-derived the soil-side caches for the restore call).
- The four throwing transport lookups name the spline, the point, how far outside it fell, the domain, and the calling function.
- Failures inside the collar solve carry the bracket they were working in, and which feasibility branch `prepare_collar_solve` exited through.

Deliberately **not** changed: the guard keeps odelia's exact comparison (`u < lo || u > hi`) rather than negating an in-range test, so a non-finite point still falls through to the spline. That is load-bearing — `profit_psi_stem_TF(NA, .)` is documented in `test-leaf.r` to return `NA`, and the value is caught downstream by the `isfinite(profit_)` guards, which report far more context than one lookup could. Negating the test reads like a tightening and is really an unrelated behaviour change.

## Verification

- Reproducer runs. Rainfall **0.03 / 0.04 / 0.05 / 0.06** all complete (0.04 and 0.05 were the failures).
- Re-ran this issue's sweep shape — rainfall × θ₀ × layers × lifetime, both models, **72 configurations: 72/72 run, no failures**.
- Full suite green. `test-strategy-tf24f.R` (57 expectations, including the reference comparison and the #527 AD-gradient tracking test) confirms healthy runs are unchanged.
- Three new tests in `test-tf24-arid-corner.R`: the dry window completes; **both** gradient methods survive shutdown (the asymmetry is what broke, so that is the property worth pinning); the diagnostic names spline/point/domain.
- Two assertions in `test-leaf.r` that pinned odelia's old bare message now assert the named diagnosis.

Behaviour changes only in states that previously threw, so the TF24f scientific version does not move.

## Not fixed here

`prepare_collar_solve`'s `bound_b = std::max(-root_crit, -root_psi_crit)` is dead code — the comment says the bracket is "clamped to `root_psi_crit`" but `-root_crit` is a positive magnitude while `-root_psi_crit` is negative, so the `max` always returns `-root_crit`. Reachable and measured; filed separately since it changes results for runs that currently complete.

dfalster:
Closing unmerged. Superseded by #605, which carries the diagnosis but not the diff.

**Why not rebase.** This branch predates #591 (leaf model out to phylloptim) and #600 (test trim). Merging `develop` in now conflicts in five of its six files: `NEWS.md`, `inst/include/plant/leaf_model.h`, `src/tf24f_strategy.cpp` and `tests/testthat/test-tf24-arid-corner.R` on content, plus `src/leaf_model.cpp` as a **modify/delete** — it would resurrect the file #591 deleted. Only `test-leaf.r` auto-merges.

**The diagnosis held.** Re-measured on `develop` 74b10ed7 (after #593, #583 and #600 merged), the failure is unchanged and the AD/FD asymmetry is exactly as described here. If anything the band is wider than this PR's body says: TF24f AD fails at rainfall **0.01 and 0.03–0.06**, not just 0.04–0.05. Rainfall 0 and 0.02 run, so it is still a knife-edge rather than monotone in dryness. FD is clean at all 13 values swept and TF24 (non-f) is clean across the full range.

Two things this PR got right that are worth not relearning:

- The out-of-domain throw is the *first statement* of what is now phylloptim's `dprofit_at_collar_psi`, so it precedes both of that function's guards — including the `feasible` out-parameter added since. Not asking is the only fix at that layer.
- Do not "tidy" `u < lo || u > hi` into `!(u >= lo && u <= hi)`. It is false for NaN by design; negating it breaks the documented `profit_psi_stem_TF(NA, .) -> NA` contract in `test-leaf.r`. Reads like a diagnostic fix, is a behaviour change.

**The other half is being split out.** This PR bundled the bug fix with making the out-of-domain error name the spline, the point and the domain. That half is not in #605, because after #591 the message comes from **odelia's** `interpolator.hpp` — which has `u`, `min()` and `max()` in hand and reports none of them. Doing it there fixes every interpolator in the family rather than the two leaf ones; phylloptim then adds the spline and caller identity odelia cannot know, and plant's `test-leaf.r:206`/`:214` (which still assert the bare string) get updated behind a version-floor bump. Separate PRs, bottom-up.

#584 stays open and latent.

PR #583, opened 2026-07-30
Score the gateway on two axes, in birth date

### LINKED ISSUE #572: [TF24 hydraulics] Score the scenario gateway on persistence, not on the model merely running
Split out of #557 so it does not stay buried in a comment thread ([the measurements are here](https://github.com/traitecoevo/plant/issues/557#issuecomment-5103704756)). #555 and #556 are merged, so the gateway exists and reruns — but it currently returns **no signal**, and the reason is a one-line classification choice rather than the confounded scenario design that has been blamed for it.

## The problem

`classify_scm_run()` ([R/scenario_eval.R](https://github.com/traitecoevo/plant/blob/develop/R/scenario_eval.R)) classifies a run as a success on `finite && total > 0`. With `birth_rate = 1` — which `build_scenario()` sets — `offspring_production` **is** the net reproduction ratio R0, so a strategy replaces itself only at R0 ≥ 1. Measured at the head of #570, `max_patch_lifetime = 100`:

| id | expected | observed | R0 | persists (R0 ≥ 1) |
|---|---|---|---|---|
| S01 | failure | success | 2.95e-01 | ✗ |
| S02 | failure | success | 4.98e-02 | ✗ |
| S03 | success | success | 6.49e-13 | ✗ |
| S04 | failure | success | 6.27e-14 | ✗ |
| S05 | failure | success | 2.33e-14 | ✗ |
| S06 | failure | success | 3.68e-14 | ✗ |
| S07 | success | success | **2.93e+01** | ✓ |
| S08 | success | success | 2.31e-15 | ✗ |

**All 8 scenarios "succeed", none of the 5 expected failures fails, and only 1 of 8 actually persists.** Five scenarios are 13–15 orders of magnitude short of replacement and every one is recorded as `"persisted"`.

Two consequences worth stating plainly:

- The match rate is **3/8**, not the 5/8 in #557's body. The drop is *good news* misreported — the matched-control crashes noted there are gone, presumably #546/#552/#554 landing — which is a sign the metric is pointing the wrong way round.
- Now that the model no longer crashes, the match rate mostly measures how well the CSV's crash predictions have aged, not model quality.

#570 adds a `persists` column and an `n_persists` summary reporting R0 ≥ 1, deliberately **alongside** `status`/`outcome` rather than replacing them, because changing the classifier in place would silently redefine what the blessed baseline means. That makes the problem *visible*; it does not decide what to do about it. This issue is that decision.

## The decision

The two axes have been conflated and need separating deliberately:

- **Numerical viability** — did the model run? This is what the CSV's "Model failure" means (see the motivation in #557: #549, #550), and what `status` should keep meaning.
- **Ecological outcome** — does the strategy persist? A different question, and the one the gateway is actually useful for now that the crashes are fixed.

Options, not mutually exclusive:

1. **Score the ecology on `persists`.** Under R0 ≥ 1: all five expected failures fail (including S01 at 0.295 and S02 at 0.0498), S07 succeeds, and S03/S08 fail against an expected success → 6/8, and measuring something real. The two new mismatches are informative rather than embarrassing: they say the CSV expects persistence where the model gives none.
2. **Rewrite the CSV's `Expectation` column** to be about persistence rather than crashing, and re-bless the baseline. More honest to the current model, but discards the crash-prediction record — so if it goes this way, keep the old expectations recoverable (git history is enough) and say so in `notes/plan-tf24-scenario-framework.md`.
3. **Report both axes on the scorecard** and stop quoting a single headline number. Cheapest, and probably wanted regardless of 1 vs 2.

My preference is 3 + 1: keep `status` as the numerical-viability gate, add a persistence score as the primary ecological metric, and retire "match rate" as the headline. That needs no CSV rewrite and no re-blessing of the existing classification.

## Note on the threshold

R0 ≥ 1 is the textbook criterion but is worth a sanity check before it is hard-coded, because `offspring_production` is scaled by the birth rate and evaluated under a **fixed** birth rate here — the same fixed-birth-rate density-dependence artefact that made TF24t-vs-TF24 R0 gaps look enormous in unrelated work. The scenarios all use `birth_rate = 1` so the comparison is internally consistent, but if the gateway ever varies birth rate, `persists` needs revisiting. `persists_at(total, finite, threshold = 1)` in #570 takes the threshold as an argument for exactly this reason.

## Done when

- [ ] Scorecard reports numerical viability and persistence as separate, clearly-labelled axes
- [ ] The headline reported by `scripts/run_scenario_gateway.R` and the scorecard report is no longer a bare match rate
- [ ] #557's stale 5/8 corrected, and `notes/plan-tf24-scenario-framework.md` updated with whichever option is taken
- [ ] Baseline re-blessed if and only if the change is deliberate, with the reason recorded

## Related

- #557 — umbrella; body still quotes the stale 5/8
- #570 — adds `persists` / `n_persists`; where this was measured
- #571 — the dry-start blocker found alongside this; unrelated cause
- #555, #556 — the CSV and the framework, both merged

dfalster:
Closes #572.

The gateway called a run a success on `finite && total > 0`. At `birth_rate = 1`
offspring production *is* R0, so five of eight scenarios were recorded as
persisting on an R0 near machine zero. `scenario_summary()` now reports
numerical viability and ecological persistence as separate axes, with CSV
agreement demoted.

Every scenario is TF24, whose reserve gate makes growth depend on more than
size, so the gateway also opts into birth-date integration (#590) via
`scenario_control()`. That moves R0 by 2.4-47x, takes S01 across R0 = 1, and
the baseline is re-blessed under it. Retracts the coordinate-dependent
two-species exclusion conclusion in `test-strategy-tf24.R`.


dfalster:

---

Closes #572. Implements **options 3 + 1** — your stated preference: keep `status` as the numerical-viability gate, add persistence as the primary ecological metric, retire the bare match rate as the headline. No CSV rewrite, no re-blessing.

## The problem, restated from the measurements

`classify_scm_run()` calls a run a success on `finite && total > 0`. With `birth_rate = 1` (what `build_scenario()` sets) `offspring_production` **is** R0, so that is far weaker than it looks: five of the eight hydraulic scenarios return R0 between 2e-15 and 6e-14 — numerically extinct — and every one was recorded as `"persisted"`. All eight "succeeded", none of the five expected failures failed, and only one replaced itself.

## What changed

`scenario_summary()` now reports two labelled axes, plus agreement demoted:

| Axis | Columns | Question |
|---|---|---|
| **Numerical viability** | `n_ran`, `n_crashed`, `viability_rate` | Did the model run? (what the CSV's "Model failure" means, #549/#550) |
| **Ecological persistence** | `n_persists`, `persistence_rate` | Does the strategy replace itself, R0 ≥ 1? (via `persists_at()`, #570) |
| CSV agreement | `n_match`, `match_rate`, expected-fail/success breakdown | Does the outcome match the CSV's expectation? |

Reflected in `scripts/run_scenario_gateway.R` and `inst/reports/scenario_scorecard.Rmd`. On the #570-head numbers the gateway now prints:

```
  Numerical viability : 8 / 8 ran (100%), 0 crashed
  Persistence (R0>=1) : 1 / 8 persist (12%)
  CSV agreement       : 3 / 8 (38%) — agreement with the crash-era
                        expectations, not a quality score
```

Viability is counted from the recorded `crashed` flag rather than from `match`, so an expected failure that crashes counts *against* viability while still counting *for* agreement — the two axes have to be able to disagree or there is nothing to separate.

## Why agreement is demoted rather than dropped

With the crashes fixed (#546/#552/#554) the match rate mostly measures how well the CSV's crash predictions have aged. The tell is that fixing the model **lowered** it, 5/8 → 3/8. It is still worth reporting; it is not a quality score.

## Not done, deliberately

Option 2 (rewriting the `Expectation` column to be about persistence). That discards the crash-prediction record and forces a re-blessing. `notes/plan-tf24-scenario-framework.md` records the decision, that option 2 was declined, and that the old expectations are recoverable from git history if it is ever taken.

`status` and `outcome` are untouched, so the blessed-baseline diff is still defined on the same columns and **the baseline is not re-blessed**. Scorecards predating the `persists` / `crashed` columns (including the blessed baseline) still summarise rather than error.

## Threshold caveat, recorded

R0 ≥ 1 is evaluated at a *fixed* birth rate, so it carries the same density-dependence artefact that inflated TF24t-vs-TF24 R0 gaps in unrelated work. All scenarios use `birth_rate = 1`, so the comparison is internally consistent; `persists_at(total, finite, threshold = 1)` takes the threshold as an argument for when that stops being true.

## Done-when

- [x] Scorecard reports viability and persistence as separate, clearly-labelled axes
- [x] Headline in `run_scenario_gateway.R` and the report is no longer a bare match rate
- [x] #557's stale 5/8 corrected (its banner now records the two-axis reading), and `notes/plan-tf24-scenario-framework.md` updated with the option taken
- [x] Baseline **not** re-blessed — the change is deliberately confined to reporting, and the reason is recorded in the note

## Tests

`test-scenario-eval.R`: 50 expectations, green. Two new blocks — the axes are reported separately and can disagree; viability is counted from `crashed`, not from the match.

dfalster:
Rebuilt on `develop` at d16d1357 — after the phylloptim port (#591), #585, #590, #574 and the odelia 0.2.1 pin (#597).

**The port does not touch this PR.** `R/scenario_eval.R`, `scripts/run_scenario_gateway.R`, the scorecard Rmd and `test-scenario-eval.R` contain no reference to `leaf` or `phylloptim`, and develop has not modified any of them since the branch point. The only conflict was in `NEWS.md`, where #585's `SpeciesBase::control()` bullet landed on the same lines; both bullets are kept.

**Re-measured rather than assumed.** Compiles clean against phylloptim 0.1.0 / odelia 0.2.1; `test-scenario-eval.R` is 50/50 green; and the gateway re-run at `max_patch_lifetime = 100` reproduces the headline unchanged:

```
  Numerical viability : 8 / 8 ran (100%), 0 crashed
  Persistence (R0>=1) : 1 / 8 persist (12%)
  CSV agreement       : 3 / 8 (38%)
```

#585's forward-model fixes did move the three scenarios that are not already at machine zero — S01 2.95e-01 → 2.59e-01, S02 4.98e-02 → 4.20e-02, S07 2.93e+01 → 3.02e+01 — by a few percent each. No `persists` verdict and no count changed. `notes/plan-tf24-scenario-framework.md` now carries both columns so the drift is on the record instead of the note quietly claiming numbers from a build that no longer exists.

Blessed baseline still not re-blessed: `status` and `outcome` are untouched.

dfalster:
Following up on the point that TF24 should be run with age integration — that was right, and it changes the result.

## The gateway was scoring TF24 through the wrong compression term

Every scenario here is TF24, and TF24 is precisely the model #590 identifies as the one the two density coordinates genuinely disagree on. The compression term is the total derivative of growth along a cohort's own trajectory, which equals `dg/dh` only when growth is a function of size; the reserve gate (#517) breaks that. The height coordinate is not a coarser approximation of the right derivative, it is an accurate derivative of a quantity the plant never experiences — and the diagnostic is that refining the node schedule does not close the gap for TF24, where for FF16 and K93 the two coordinates converge at ~2nd order.

So `scenario_control()` now supplies `control(node_density_in_birth_date = TRUE)` and is the default `ctrl` for `build_scenario`, `classify_scm_run`, `evaluate_scenario` and `run_scenarios`. The package default is left at `FALSE`; only the gateway opts in.

## Measured, at `max_patch_lifetime = 100`

| id | expected | R0 height | R0 birth date | ratio | persists |
|---|---|---|---|---|---|
| S01 | failure | 2.590e-01 | **1.489e+00** | 5.8x | ✗ → **✓** |
| S02 | failure | 4.203e-02 | 9.964e-02 | 2.4x | ✗ |
| S03 | success | 6.488e-13 | 5.499e-12 | 8.5x | ✗ |
| S04 | failure | 6.267e-14 | 4.746e-13 | 7.6x | ✗ |
| S05 | failure | 2.348e-14 | 5.888e-13 | 25x | ✗ |
| S06 | failure | 3.684e-14 | 3.261e-13 | 8.9x | ✗ |
| S07 | success | 3.023e+01 | 1.415e+03 | 47x | ✓ |
| S08 | success | 2.171e-15 | 3.230e-14 | 15x | ✗ |

**Persistence 1/8 → 2/8.** Viability stays 8/8, CSV agreement stays 3/8. Birth date is also about 2x faster here (36 s vs 70 s).

## What that exposed about the baseline guard

Across a 47x swing in R0 and a persistence flip, `observed` did not change on a **single** scenario. It tests `finite && total > 0`, which everything has satisfied since the crash fixes landed, so it was pinned at 8/8 and could not move. A guard that survives that is not guarding what the gateway now reports, so `test-scenario-gateway.R` diffs `persists` alongside `observed`.

Two related gaps closed while in here:

- **The baseline did not record its own coordinate.** It does now — `node_density_in_birth_date` and `max_patch_lifetime` go into the scorecard metadata, and `max_patch_lifetime` is set as an attribute rather than silently defaulting to 100 in the diff. A stored scorecard that does not say which coordinate produced it cannot honestly be compared to another.
- **The missing-`crashed` fallback in `scenario_summary()` had no real coverage** (the blessed baseline lacks `persists` but *has* `crashed`), so it is now pinned by a test.

## Re-blessed

This reverses the "not re-blessed" position earlier in the PR, which was correct for a pure reporting change and is not correct for a coordinate change that moves a verdict. The note records the reversal and the reason.

Tests: `test-scenario-eval.R` 53/53, `test-scenario-gateway.R` 2/2 with `PLANT_RUN_SCENARIOS=1`. CI was green on the previous push.

dfalster:
Bundled in the re-derivation of the two-species TF24 conclusion, which #590 flagged as coordinate dependent.

## The conclusion does not survive the coordinate change

`test-strategy-tf24.R` recorded "under the NSC reserve-gated growth of #517 the slower species is now largely competitively excluded" as a finding about #517. Re-derived at the test's own configuration (`lma` 0.0825 / 0.10, `max_patch_lifetime = 5`):

| coordinate | fast | slow | ratio |
|---|---|---|---|
| height | 67.32 | 2.775e-04 | 2.4e5 |
| birth date | 287.2 | 59.53 | **4.8** |

They coexist. At `max_patch_lifetime = 30` the ratio is 2.7 (2707 against 1004, reproducing #590 exactly) — so it *narrows* with patch lifetime, where progressive exclusion would widen it.

## Refinement is what settles it

This is the part that distinguishes a wrong derivative from a coarse one. Over two halvings of the node spacing:

| refine | nodes | height fast | height slow | ratio | birth fast | birth slow | ratio |
|---|---|---|---|---|---|---|---|
| 0 | 88 | 67.32 | 2.775e-04 | 2.43e5 | 287.2 | 59.53 | 4.824 |
| 1 | 175 | 73.18 | 2.939e-04 | 2.49e5 | 287.1 | 59.66 | 4.812 |
| 2 | 349 | 74.98 | 2.984e-04 | 2.51e5 | 287.2 | 59.69 | 4.811 |

Birth date is already converged at the default schedule — it moves in the third significant figure across a 4x increase in nodes. Height is still climbing at 349 nodes, and its exclusion ratio does not shrink toward the birth-date one, it **grows**. Quadrature error closes under refinement; a different derivative does not.

## What changed

- The comment in `test-strategy-tf24.R` no longer attributes the exclusion to #517. It attributes it to the coordinate, and says why birth date is the right one for a reserve-gated model.
- The height assertions are **kept** — they pin what `Control()` still does by default.
- A birth-date two-species case is **added**, pinning 287.16 / 59.53 at the same 2e-2 tolerance, so the correct-coordinate result is covered by a test rather than only described in prose.
- #590's own NEWS entry said the conclusion "needs re-deriving before it is relied on"; that now points at the resolution instead of leaving an open TODO.

## What I did not claim

What survives is the weaker statement: the faster species still leads by roughly 3-5x, so reserve-gated growth disadvantages the slower species without excluding it. How much of even that gap is attributable to #517 as against the trait difference itself is **not** measured here — isolating that needs a gate-disabled comparison, which I have not run. NEWS says so explicitly rather than letting the surviving gap be read as #517's effect.

Tests: `test-strategy-tf24.R` 50/50, `test-scenario-eval.R` 53/53, `test-density-coordinate.R` 247/247.

dfalster:
Merged `develop` at 835af08b (#593, environment integration in the stochastic solver).

**No numbers moved — the deterministic path is bit-identical.** #593 split `clear_environment()` (competition profile) from a new `clear_state()` (soil states and flux accumulators), which sounds like it could reach the SCM, since both headers are shared. It does not: `Patch` calls `environment.clear()`, which still invokes both, and only `StochasticPatch` calls the bare `clear_environment()`. Verified rather than assumed — every scenario reproduces the blessed baseline exactly:

```
  id     blessed         now         reldiff
 S01 1.48872e+00 1.48872e+00               0
 S02 9.96356e-02 9.96356e-02               0
 S03 5.49856e-12 5.49856e-12               0
 S04 4.74591e-13 4.74591e-13               0
 S05 5.88810e-13 5.88810e-13               0
 S06 3.26147e-13 3.26147e-13               0
 S07 1.41520e+03 1.41520e+03               0
 S08 3.23026e-14 3.23026e-14               0

observed identical: TRUE | persists identical: TRUE | R0 bit-identical: TRUE
```

So the baseline is **not** re-blessed and the two-species pins are unchanged. The only conflict was in `NEWS.md`, where both sides appended to the same section; both sets of bullets are kept.

Tests on the merge: `test-strategy-tf24.R` 50/50, `test-scenario-eval.R` 53/53, `test-stochastic-patch-runner.R` 120/120.

Issue #578, opened 2026-07-29
[TF24] Shut-down exits leave the leaf's uptake untouched


aornugent:
`set_shutdown_state(root_collar)` sets `root_collar_psi_`,
`opt_psi_stem_` and `profit_`, and neither `soil_consumption_` nor `E_up_`. Three of
`prepare_collar_solve`'s early exits call it, leaving three different wrong states:

| exit | what `soil_consumption_` holds |
|---|---|
| `-wettest_soil_layer >= psi_crit` | the previous cohort's values — `E_from_Soil_to_Root_Collar` is not called at all |
| `E_column(-psi_crit, ...) < 0` | uptake at `-psi_crit`, not at the shut-down point |
| `-root_crit >= psi_crit` | the **last trial iterate of the TOMS748 root-find** inside `find_root_psi` |

A shut-down plant transpires nothing; the correct value is zero in all three.

**Measured.** A wet 20 m tree, then the identical tree with every layer drier than
`psi_crit`: `E_up_` and `soil_consumption_` are unchanged from the wet solve. The same
dry solve on a fresh leaf reports `NA`, so the value is entirely a function of what ran
before it.

The `assim_max_ < 0` exit does call `E_from_Soil_to_Root_Collar` at its own operating
point and is already correct — it shows the intended shape.

dfalster:
Hit this independently while extracting the leaf model into a standalone package, and the reproduction agrees with yours. Adding three things this issue doesn't currently cover.

**The carbon-side fields go stale too, not just `soil_consumption_` and `E_up_`.** `set_shutdown_state` also leaves `transpiration_`, `stom_cond_CO2_`, `assim_colimited_` and `ci_` untouched, and `set_physiology` doesn't reset them either — `setup_clean_leaf()` is called from the two constructors and nowhere else. Same wet-then-dry protocol, one soil layer, `psi_crit` = 5.87 MPa:

```
psi_soil=  4.0  E=3.1769e-06  A=1.7432  gc=0.005347  soil_cons=0.000176347  profit=-3.7056
psi_soil= 12.0  E=3.1769e-06  A=1.7432  gc=0.005347  soil_cons=0.000176347  profit=-8.3846
psi_soil= 20.0  E=3.1769e-06  A=1.7432  gc=0.005347  soil_cons=0.000176347  profit=-8.3846
```

`profit_` tracks correctly; everything else is the 4 MPa solve.

[...]

Issue #577, opened 2026-07-29
[TF24] `soil_consumption_` carries over between cohorts


aornugent:
`Leaf::set_physiology` calls

```cpp
soil_consumption_.resize(soil_number_of_depths_, 0.0);
```

where it needs `.assign`. `resize`'s fill argument applies only to newly added
elements. `E_from_Soil_to_Root_Collar`'s loop then runs to `max_soil_layer`, not to
`soil_number_of_depths`, so every layer a plant has no roots in is neither written this
solve nor cleared — and `TF24_Strategy::compute_rates` bills all of them to the patch
water balance. The `Leaf` is shared across every cohort of a species through the
`Strategy` `shared_ptr`, so the inherited values are the previous cohort's.

Two lines above, the sibling buffers use `.assign(max_soil_layer, 0.0)`. The file is
inconsistent with itself by one call.

**Measured.** A 20 m tree then a 0.4 m seedling on the same leaf: the seedling's
layers 3–5 are bit-identical to the tree's. Solve seedling → tree → seedling and the
two seedling records differ. On a production run (`max_patch_lifetime = 105.32`,
10 153 cohort-time records) **33.78%** carry at least one stale layer and **25.73%**
carry three of five.

itowers1:
Recommending this is a high priority change, but aware that we are currently undergoing a number of changes.

dfalster:
**Fixed.** In plant by #585 (`soil_consumption_.assign(...)` rather than `.resize`), and in the
leaf package by traitecoevo/leaf_cpp#26, which is where this code now lives.

Thanks for finding it — the reproduction was exact and the mechanism was subtle: `resize`'s fill
reaches only newly-added elements, and the uptake loop writes only up to `max_soil_layer`, the
deepest *rooted* layer rather than the deepest soil layer. So a shallow-rooted plant solving on a
`Leaf` a deep-rooted one used before it inherited the deep layers' uptake, and
`evapotranspiration_dt` billed them to the patch water balance.

Two notes for the record:

- **Reproduced independently in the leaf package**, with the soil layer count held constant so
  that only the rooted depth shrinks — which is why resizing never noticed. Reverting the one word
  gives 1.31e-04 and 4.75e-05 kg H2O m⁻² s⁻¹ of phantom uptake in layers 2 and 3, and disagreement
  with a fresh leaf. That is the ~1e-4 "real difference" scale, four orders above rounding.
[...]

Issue #564, opened 2026-07-11
Mutant-fitness replay broken: odelia migration (#456) dropped the RK environment-cache hooks


dfalster:
## Summary

The mutant-fitness method (`SCM$run_mutant`, #362) is broken on `develop`. Any mutant run fails with:

```
Error: Run a resident first to generate a competitve landscape
```

even when the resident was run with `save_RK45_cache = TRUE`. `plant`'s own `tests/testthat/test-mutant.R` currently fails (`FAIL 2 | PASS 0`); it is not skipped.

## Reproduction (canonical pattern from test-mutant.R)

```r
library(plant)
p0 <- scm_base_parameters("FF16"); p0$max_patch_lifetime <- 50
e  <- Environment("FF16")
ctrl <- Control(); ctrl$save_RK45_cache <- TRUE          # flag persists (verified)
pr1   <- add_strategies(p0,  trait_matrix(0.1, "lma"), birth_rate = 1)
pr1m1 <- add_strategies(pr1, trait_matrix(0.2, "lma"), birth_rate = 1)

scm <- run_scm(pr1, e, ctrl)      # resident runs; net_reproduction_ratios OK
scm$run_mutant(pr1m1)             # Error: Run a resident first ...
```

## Root cause

The mutant replay depends on three hooks the ODE solver was supposed to call on the `Patch` (the ODE system):

- `Patch::cache_RK45_step(int)` — snapshot the environment at each RK stage into `environment_cache`
- `Patch::cache_ode_step()` — push `environment_cache` + time into `environment_history` / `step_history` after each accepted step
- `Patch::load_ode_step()` — load a cached environment during the mutant replay

All three still exist in `inst/include/plant/patch.h` but have **zero call sites** in plant or odelia — they are dead code. `Patch::set_mutant()` therefore throws because `environment_history` is always empty.

Before #456 these were wired in plant's own solver:

```
inst/include/plant/ode_solver/ode_solver.h:83  system.cache_ode_step();
inst/include/plant/ode_solver/ode_solver.h:96  system.load_ode_step();
inst/include/plant/ode_solver/ode_step.h:75    system.cache_RK45_step(rk_step);
```

[...]

PR #585, opened 2026-07-31
Forward-model correctness fixes in TF24's leaf, soil and demographic paths

aornugent:
Ten fixes to TF24's forward model, plus one that reaches FF16 and K93. Six are bit-identical corrections of latent defects; four move simulated numbers and need baselines re-blessed.

The common thread is **state leaking between plants that should be independent.** Every cohort of a species shares one `Strategy`, and so one `Leaf`. Several of these are the same defect at different depths: a buffer sized but not cleared, a member refreshed by one caller and read by another, a rate computed once and read after the state it described changed. Individually small; together they were enough that re-running one plant's rates from its own inputs did not reproduce what the solver had just computed. Solving this now helps with reverse mode where we want to compute from rates backwards.

## The changes

1. **Uptake charged to the wrong plant.** `Leaf::set_physiology` sized its per-layer uptake buffer with `resize`, whose fill reaches only new elements, and the solve writes only as deep as the plant has roots — so layers below rooting depth kept the previously solved plant's values, billed to the water balance. (`assign`, matching the sibling buffers two lines above.) **33.8% of plant-time records affected, 7.3% holding a stale non-zero value**; the gap is the early run, separable only by solving the same states in different orders. The shut-down exits leave the same fields stale (zero incidence at the default driver, margin 27x tolerance); and the profit derivative reads a soil vector only the solve refreshes, differing by 100% when called stale. Addresses #577 and #578.
2. **Uptake integrated from the wrong lower endpoint.** Water started at the smallest *existing* plant and returned zero for a lone plant, so a transpiring recruit drew nothing — at **0.70% of output times, the establishment window**. It now starts at the newborn, as the light field already does, and the one-plant case cannot arise. Reaches the recruitment channel, since establishment reads the soil this depletes.
3. **Rates not recomputed at introduction.** Introduction widens the state and rebuilds the field but left the rates describing the *pre*-introduction state, which the integrator then used as the derivative of the *post*-introduction state. **Wrong by more than a plant's own magnitude at 51 of 141 introductions** (the transport term amplifies). Family-wide, and the only change moving FF16 and K93.
4. **Three canopy profiles collapsed to one.** TF24 wrote its own cumulative form, density, inverse and a second inlined copy of the cumulative form; it now uses the shared `CanopyShape` FF16 and K93 already do, and the crown-centre coordinate (three copies) becomes one static function. Compared before switching over 15 087 real pairs: **every difference bounded by 2 units in the last place**, so the forward move is the step controller, not the profile. develop's exponent-specialised multiplication chains are untouched — an earlier draft rewrote them for readability and moved two models for nothing; they supply values, carry no exponent term, and were never the route to a derivative. One `pow` stays because the root-mass distribution calls the cumulative form at a different exponent.
5. **Leaf-area density finite at the crown base.** The density divided by height, giving `0/0` at the base for every height — the field's lowest sample point. It now branches to the limit there: **zero above exponent 1, `2/height` at 1**. The density and cumulative form are an exact derivative pair, verified to hold to 4e-12–6e-11 wherever a difference can check it.
6. **Soil moisture inverse converts its units.** Missing the MPa→Pa factor, so it returned values scaled by 8.19, above saturation. R-facing utility only, now with a round-trip test.
7. **Resource vector sized by resource count.** Sized by ODE width (nine) and filled with `NA`, so four `NA`s per plant per stage were integrated to `NaN` and discarded — **1 979 of 2 000 calls, none after**. Now an explicit `n_resources()` on the environment, one source of truth.
8. **Newborn leaf solved once per stage, not twice.** Setup computed the rates, then establishment recomputed the same quantity at identical arguments (a newborn is at birth size already). The stored value is passed in instead — **bit-identical, one fewer solve per stage per species**, the total difference equal to the stage count. FF16 fixed the same way; K93 reads no carbon budget.

## Measurements, for the whole change

One species, five soil layers, default driver, full patch lifetime, `refine_schedule = FALSE`, built `-O2 -DNDEBUG -g0`. Both numbers from one session on one machine — **a value gate that does not name its flags measures the compiler**, since this tree at `-O0` differs by 0.145% in offspring from the step controller alone.

| | offspring production | accepted steps |
|---|---|---|
| `develop` | `42.140173575095666` | 5 055 |
| this branch | `42.176246845059751` | 5 105 |

**+0.086% in offspring, +50 accepted steps.** The four value-moving fixes measured individually shift offspring by +0.14% to +0.79%; composed, +0.086%. They do not add — most of each individual figure is the adaptive controller landing on a different accepted sequence, not a systematic change. Treat the composite as the number, and distrust any per-fix figure below about 0.15%.

FF16 and K93 move only through change 3: FF16 **+0.0034%** (two more steps), K93 **+0.000038%** (same step count), both at full lifetime.

Runtime unchanged. Change 3 adds 141 rate evaluations against ~30 000; change 8 removes one solve per stage per species.

**Tests: 1 007 assertions pass across 15 files. Three fail, and they are the re-blessing this asks for** — all in FF16's "offspring arrival", on offspring production, an ODE time and a step count, against `testthat`'s default ~1.5e-08 tolerance given a 3.4e-05 movement. Nothing under `tests/testthat/FF16_reference/` was regenerated, and FF16's reference-comparison file still passes because its own tolerance absorbs the shift — so the hard-coded assertions are the tripwire, not the reference files. Two pre-existing failures (mutant-fixture tests) are unrelated and identical on `develop`.


aornugent:
Flagging an alternative: `49b03bb` adds a `compute_rates()` call to `introduce_new_nodes`, which is correct and unblocked. But it increases the number of state-mutating paths in `Patch` that must remember to recompute — `set_ode_state` (both overloads) and now `introduce_new_nodes` — because `ode_rates()` is `const` and only copies out whatever the last `set_ode_state` computed. Any future path that reaches a state by another route reintroduces the same bug.

The structural fix is to move `compute_rates()` out of both `set_ode_state` overloads and into `ode_rates()`, which is then no longer `const`. `ode_rates` today returns whatever the last `set_ode_state` computed, so any caller that arrives at a state by another route reads rates belonging to a different one. `Patch::introduce_new_nodes` widens the state and rebuilds the light field without recomputing rates, and `set_state_from_system` — the second blocked signature below — then seeds `dydt_in` from those stale rates and marks them clean. Under first-same-as-last that vector becomes `k1`. With ode_rates computing, that cannot arise — the rates are always those of the state just set.

No arithmetic moves and no reordering: the forward path calls `set_ode_state` then `ode_rates`, so the same work happens in the same sequence. The invariant would hold by construction — rates are always those of the state just set — and the `compute_rates()` added here becomes unnecessary rather than correct. 

It also separates loading a state from evaluating at it, which is what a reverse pass wants: re-establish a stage's state and environment without paying for a rate evaluation the recordings will redo. See https://github.com/aornugent/plant/pull/65 as an example.

It is **blocked in odelia**, which is why it isn't proposed as part of this PR. 
Two signatures take the System by const reference and would need mutable ones:

| site | what it does |
|---|---|
| `odelia/ode_interface.hpp` — `r_ode_rates(const T& obj)` | the R accessor |
| `odelia/ode_solver_internal.hpp` — `set_state_from_system(const System& system)` | seeds `dydt_in` from `ode_rates` and marks it clean |

Happy to raise a paired PR if this is preferred.

---
_Generated by [Claude Code](https://claude.ai/code)_

aornugent:
One more in `903b8292`.

**The mechanism.** `Patch::reset()` returns the patch to `t = 0` and calls `environment.clear()`. For TF24 that reaches `clear_environment()`, which clears the light profile and never touches `vars.states` — the five soil moisture states and the four cumulative flux accumulators. So a reset patch is at time zero with the previous run's soil.

**It is reachable, and by the one caller that matters.** `run()` begins with `reset()`, and `refine_schedule()` calls `run()` in a loop. So every refinement iteration after the first computes its error signal against a run that started from the previous iteration's depleted soil, and the schedule is then chosen on that signal. 

**Measured, on the branch before the fix**, one species, five layers, default driver, full lifetime:

| | offspring production | accepted steps |
|---|---|---|
| first `run()` | `42.176246845059751` | 5 105 |
| second `run()` on the same SCM | `42.296696440471194` | 5 062 |

and the state after `run()` then `reset()` begins `0.284128746037276 …` against `0.214` for a freshly built environment, with the flux accumulators at `81.2` rather than zero. After the fix the second run reproduces the first exactly and the post-reset state is bitwise equal to a fresh one.

**The forward run is unmoved** — `42.176246845059751` at 5 105 accepted steps, the same figure this branch already reports. That is the safety property rather than a null result: the fix restores an initial condition, it does not change one. Nothing to re-bless.

**The starting state is snapshotted, not recomputed, and that is the one design choice here.** The constructor's last line is the only initialiser, so recomputing it *would* reproduce the default — but `set_soil_water_state` is part of the R interface and is exercised (a test hands `run_scm` a one-layer soil at `0.1`; `scenario_eval` sets it too). Recomputing the default at `reset()` would overwrite the caller's choice and move the **first** run's initial condition, which is exactly what this must not do. So the state is recorded where it is set and restored where it is cleared, and it cannot drift from either.

_Generated by [Claude Code](https://claude.ai/code)_

inst/include/plant/models/tf24_environment.h
```diff
@@ -409,7 +421,8 @@ class TF24_Environment : public Environment {
     const double n_psi_layer = soil_parameter_value(n_psi_layers, n_psi, layer);
     const double soil_moist_sat_layer =
       soil_parameter_value(soil_moist_sat_layers, soil_moist_sat, layer);
-    return pow((psi_soil_ / a_psi_layer), (-1 / n_psi_layer)) * soil_moist_sat_layer;
+    // psi_soil_ is in MPa; a_psi is in Pa.
+    return pow((psi_soil_ * 1e6 / a_psi_layer), (-1 / n_psi_layer)) * soil_moist_sat_layer;
```
dfalster, line 425: @itowers1 Can you check this?

src/tf24_strategy.cpp
```diff
@@ -38,7 +38,8 @@ double TF24_Strategy::compute_average_light_environment(
 // rather than allowed to reach 0 (original rationale was never recorded;
 // preserved as-is).
 
-     return std::max(environment.get_environment_at_height(z), 0.0001) * q(z, height);
```
dfalster, line 41: wow, good catch

src/tf24_strategy.cpp
```diff
@@ -720,31 +729,21 @@ double TF24_Strategy::establishment_probability(const TF24_Environment& environm
...
-  return 2 * pars.eta * (1 - tmp) * tmp / z;
+  return pars.k_I * area_leaf_ * canopy_shape.Q(z * height_inverse);
 }
 
-// [eqn 10] ... Fraction of leaf area above height 'z' for an
-//              individual of height 'height'
+// [eqn 10] Cumulative fraction of a quantity distributed over an extent with
+//          shape exponent 'eta_x', above coordinate 'z' of a total 'height'.
+//          Serves the root mass distribution over soil depth.
 double TF24_Strategy::Q(double z, double height, double eta_x) const {
```
dfalster: Presumably this can go too, now that using the canopy_shape object

tests/testthat/test-scm.R
```diff
@@ -295,3 +295,29 @@ test_that("Can create empty SCM", {
   }
 })
 
+test_that("A second run on one SCM reproduces the first", {
```
dfalster, line 298: smart

dfalster (approved):
Excellent work catching these @aornugent .  Have flagged on for @itowers1 to comment on

tests/testthat/test-scm.R
itowers1: Great idea, like this!

inst/include/plant/models/tf24_environment.h
itowers1: Yep, technically true if psi_soil was passed in as MPa, which it emerges from the soil moisture model as. The soil_moist_from_psi functions are not actually in use at the moment from what I can see. Might be worth raising a seperate issue that we need to swap all potential values and parameters to MPa for consistency.

dfalster:
Approving on the substance, but three things before this merges — the first is blocking.

**CI is red and needs the re-blessing committed.** The three FF16 "offspring arrival" assertions fail identically on macOS and Windows. I reproduced them locally and bisected: deleting the `compute_rates()` added to `introduce_new_nodes` restores develop's numbers to the last digit, so the attribution in the description holds and everything else in the branch really is bit-identical for FF16. I also traced why that fix is needed rather than merely tidy — `set_state_from_system` copies `ode_rates()` into `dydt_in` and sets `dydt_in_is_clean = true`, so `setup_dydt_in` never re-evaluates and that vector becomes `k1` under first-same-as-last. So: happy for the numbers to move, but please commit them (`4.215899`, `c(11.995204, 16.474988)`, `293`) rather than merging red. The assertion at line 215 still passes on its `1e-4` tolerance while having drifted to `16.8895016065` — worth recording at the same time.

**The FF16 figure in the description has the wrong sign.** It says +0.0034%. Offspring actually moves *down*: 0.00074% and 0.00122% in the two-species run, and up 0.00024% in the one-species run. The 3.4e-5 looks like the `ode_times[100]` shift (3.7e-5), not offspring. Worth fixing so nobody chases the sign later.

**`Species::consumption_rate` assumes a monotone grid.** It now integrates over `r_heights_rev()` with `new_node` prepended, which is the same assumption #574 had to abandon for the competition profile, where inverted heights are handled by sorting in `compute_competition_unordered`. The uptake integral has no equivalent fallback, so an inverted grid gives a partially-cancelling trapezium. Not a regression — develop integrated over the same unsorted grid — and CI won't catch it while the dry-start TF24 test is skipped. But this PR extends that function, so it's the natural moment to make the two paths consistent. Happy for it to be a follow-up issue if you'd rather keep this PR closed.

---
_Drafted with [Claude Code](https://claude.ai/code)_

dfalster:
Agreed on the direction — `ode_rates()` computing is the right shape, and yes please to the paired odelia PR. Three things before you cut it.

**It won't hold by construction as written.** `set_ode_state(it, time)` builds the field with `compute_environment(true)`; `introduce_new_nodes` uses `(false)`. Moving `compute_rates()` into `ode_rates()` guarantees the rates match the *state* just set, but not the *environment* they were computed against — so a future path that reaches a state without refreshing the field reintroduces the same bug one level up. If you want the strong invariant, the environment refresh has to move with it: one sync point on `Patch` doing both, called by every mutating path.

**Please demonstrate the bit-identity rather than assert it.** Under the refactor `introduce_new_nodes` drops its `compute_rates()` and `set_state_from_system` → `ode_rates()` computes at the same point against the same `rescale = false` field, so the result should be bitwise equal to this branch's head. That's the difference between the FF16 re-blessing surviving the refactor and needing a second one.

**There is a cheaper option in odelia alone, and the reason to reject it is the argument you should be leading with.** Have `set_state_from_system` stop copying rates and leave `dydt_in_is_clean = false`; `setup_dydt_in` then calls `derivs`, which re-sets the state and recomputes from it, so consistency holds however the caller arrived. Same class of bug closed, no plant change at all — and no signature change either, because `set_state_from_system` would no longer call `ode_rates` and could stay `const&`. Half the blocked-in-odelia table goes away. But it routes through `set_ode_state(it, time)` and so rebuilds the field with `rescale = true`, which moves the numbers again, whereas yours keeps the field `introduce_new_nodes` deliberately built. That's a stronger argument for your version than the reverse-mode one you're currently leading with.

Cost isn't an objection, for the record: there are five `ode_rates` call sites in odelia, three of them inside `derivs` where `set_ode_state` immediately precedes, so nothing in the RK stages recomputes twice. The extra work is one evaluation per `set_state_from_system` — the same order as the 141 introductions this branch already pays for.

Two small things while you're in there. `Patch::r_compute_rates()` already exists as an explicit compute-from-R entry point, so after this there'd be two ways to do it — collapse them. And since we're already pinned to odelia master for the mutant cache hooks (#564), fold the signature change into that pin rather than making it a separate bump.

---
_Drafted with [Claude Code](https://claude.ai/code)_

src/tf24_strategy.cpp
dfalster: Correcting myself here — it can't go. Its one remaining caller is the root-mass distribution (`tf24_strategy.cpp:427`), which passes `pars.root_depth_shape_eta` (0.2), not `pars.eta` (12), and the `canopy_shape` member is bound to `pars.eta` at `initialise()`.

It *could* go if TF24 carried a second `CanopyShape` initialised at the root exponent, and that would be bit-identical — `select_pow_eta` falls back to `std::pow` for 0.2, and both forms guard `z > height` the same way. But it would hang a shading model and an `eta_c` off something that is a soil-depth profile with neither, so I'd leave the split alone. The header comment still describes `Q` as the canopy's cumulative form though, so that's worth a fix.

---
_Drafted with [Claude Code](https://claude.ai/code)_

inst/include/plant/models/tf24_environment.h
dfalster: Thanks @itowers1 — and confirmed on "not in use": the only references anywhere are the RcppR6 binding and the new round-trip test, with no C++ caller, so the fix corrects an exported utility without moving any simulated number. The forward direction on the line above already divides by 1e6 with an explicit `Pa -> MPa` comment, so this is just its exact inverse.

Agreed the real fix is the consistency pass — that's #587.

---
_Drafted with [Claude Code](https://claude.ai/code)_

dfalster:
Follow-up opened: #589, the structural version of change 3, plus traitecoevo/odelia#46 for the two signatures it needs. Both drafted against the questions raised above.

**The bit-identity is demonstrated, not asserted.** #589 is bitwise equal to this branch on FF16 (one and two species), K93, TF24 at full patch lifetime and FF16 schedule refinement — same offspring production to all 17 digits, same accepted step counts. So the re-blessing you land here stays valid through the refactor, which was the thing worth checking. Runtime is unchanged too: 44.26 s vs 44.36 s on TF24 full lifetime over six alternating reps, against a 1.5 s spread within each build.

**One thing worth knowing before this merges.** Moving the compute into `ode_rates()` made `patch$ode_rates` segfault on a fresh R-constructed patch, and the cause is already on `develop`: `Patch` holds `environment_type* environment_ptr` at its own `environment` member with default copy semantics, so every copy's pointer refers to the *source's* members. `scm.h` does `patch = sys` and RcppR6 hands copies to R. It survives today only because the sole dereference is `compute_rates()`, which every live path reaches on an object whose source is still alive — the first new caller to dereference it crashes. #589's first commit replaces it with an index. Nothing for this PR to do, but it is a live trap for anything else that touches that pointer.

Also filed #588: `Node::compute_initial_conditions` computes rates and then seeds mortality and log-density, so the rates it stores describe the state before the seeding — the same shape one level down. Harmless through the solver once #589 lands (the patch recomputes first); it is what makes `test-patch.R:86` merely equal rather than identical, at ~1e-22.

---
_Drafted with [Claude Code](https://claude.ai/code)_

aornugent:
Thanks for the reviews. I have gotten a little tangled up cause I didn't see Daniel's PRs (#588 #589). 

This branch addresses the reviews and implements #589. Commits: `13fba26`, `15e430f`, `2c49239`, `f436e9b`. Because #589 is stacked, I think we close it instead. #588 is independent and should be merged after this one. 

I've just merged https://github.com/traitecoevo/odelia/pull/46 which means the CI should hopefully go green.

## Re-blessing

Committed in FF16's "offspring arrival": `4.215899`, `c(11.995204, 16.474988)`, `293`. Line 215 now records `16.8895016`. That assertion keeps its `1e-4` tolerance and passed throughout, so the value is written down for the next change. No file under `FF16_reference/` was regenerated.

One other assertion changed, in `test-patch.R`. `patch$ode_rates` is now compared with `expect_equal`, not `expect_identical`. Reading the rates computes them at the state just introduced. That does not give the same last bits as the rates `compute_initial_conditions` stored. `$derivs` has the same difference two lines below, where a comment said "should be identical, but are merely equal". That comment is now unnecessary and is removed.

Full suite, serial, `-O2`: all 42 files pass except `Report generation`, which needs `ggridges`. That failure is unrelated.

## FF16 figure

You are correct. The +0.0034% was the `ode_times[100]` change. The offspring values, against the previous assertions, are −0.00072% and −0.00122% for two species and +0.00025% for one species. `ode_times[100]` is −0.0037%. K93 is +0.000038% at the same step count. Use these in place of the figure in the description.

## `Species::consumption_rate`

Fixed here, not deferred. It now sorts the height and rate pairs when the heights are not in order, as `compute_competition_unordered` does. An ordered grid runs the same arithmetic as before, so the result is bit-identical for every run in CI and in the scenario gateway.

## `introduce_new_nodes`

The `compute_rates()` call is removed. `compute_rates()` now runs in `ode_rates()` instead of in the two `set_ode_state` overloads, as proposed in aornugent/plant#65. The odelia change is aornugent/odelia#48.

The two forms give identical numbers. I built the tree twice, changing only where `Patch` computes its rates, and printed 17 digits:

| | in `introduce_new_nodes` | in `ode_rates` |
|---|---|---|
| TF24 offspring | `129.21868679939018` | `129.21868679939018` |
| FF16 one species | `16.889501626141417` | `16.889501626141417` |
| FF16 two species | `11.995204444405863`, `16.474988184812009` | `11.995204444405863`, `16.474988184812009` |
| K93 offspring | `0.12811440867867754` | `0.12811440867867754` |

Every digit agrees, and so do the step counts: TF24 3722, FF16 271 and 293, K93 348. The re-blessing above therefore does not need to be done a second time.

**On the invariant.** You are right that it does not hold by construction, and I do not think it can. The guarantee is that the rates match the state, not that they match the light field. The choice of field belongs to the state change, not to the rate evaluation. `set_ode_state(it, time)` builds with `rescale = true`, which is only correct when the node list did not change. `introduce_new_nodes` must use `false` over the wider node list. The mutant overload must not build a field at all. `ode_rates()` cannot make that choice unless something tells it, and that something is the mode flag this change removes. So the field build stays in the function that changes the state, and only the rate evaluation moves. A future state-changing path must still call `compute_environment`. That is one line, in a function that already decides how it changed the state.

**On the odelia-only option.** Your argument is the stronger one, and aornugent/odelia#48 now leads with it. Leaving `dydt_in_is_clean = false` sends the work through `set_ode_state(it, time)`, which builds the field with `rescale = true` and discards the field `introduce_new_nodes` built. That changes the numbers. The reverse-mode reason is now a secondary point.

**`Patch::r_compute_rates()`** is removed. Reading `$ode_rates` from R computes the rates, so the second entry point had no remaining use. NEWS has the migration line. The odelia pin is unchanged, as you suggested: `Remotes: traitecoevo/odelia` picks up the new signatures with the mutant cache hooks.

## A defect this exposed

`Patch` held a raw pointer to its own `environment` member. The default copy constructor copied that pointer, so every copy pointed into the patch it was copied from. RcppR6 gives R a copy, and `SCM`'s `sync_patch` makes one too. Nothing read the pointer on those objects while `ode_rates()` only returned stored values. Once `ode_rates()` computes, the first `patch$ode_rates` from R reads freed memory and the process stops.

The pointer is gone. `rate_environment()` now returns a reference computed on each call from data the patch owns: its own environment, or on a mutant step the recorded environment at the index that step set. The numbers do not change, and four sites no longer have to reset the pointer.

This also settles the first of the two instrumentation notes in the description. A mutant step computes against the recorded environment it experiences, and a copied patch cannot hold a bad pointer.

## `TF24_Strategy::Q`

Renamed and re-documented for its remaining caller, the root mass distribution at `pars.root_depth_shape_eta`. The argument is now `rooting_depth`, not `height`. The canopy uses `CanopyShape::Q` at `pars.eta`. The split stays, as you said.

## #577 

Closed in https://github.com/traitecoevo/plant/pull/585/commits/38214b2b9833de3a80cd7e917ad4b46937b91595

## #578

Removing this from the PR changes no code. `set_shutdown_state` on develop already sets `E_up_` and `soil_consumption_` to zero, from #570, so this branch carried a duplicate of that fix and the diff never touched the function. The version on the refactored leaf package is not blocked. Change 1 in the description claims more than the diff does, for that reason.

One related change is still in the PR: `dprofit_droot_collar_psi` now calls `refresh_soil_potentials()`. That is a different member — `psi_soil_inverted_` and `root_vuln_integral_soil_`, which only `prepare_collar_solve` used to compute, so the profit derivative read whatever the last solve wrote. TF24f calls this function on its forward path. @dfalster - I think this is independent of the leaf package refactoring?

Issue #588, opened 2026-08-03
Node::compute_initial_conditions leaves rates describing the state before its own seeding writes


dfalster:
`Node::compute_initial_conditions` computes the newborn's rates and *then* writes
two states:

```cpp
individual.set_initial_states(environment);
compute_rates(environment, pr_patch_survival);          // <- rates computed here

const double pr_estab = individual.establishment_probability_of_newborn(environment);
individual.set_state("mortality", -log(pr_estab));      // <- state written after
const double g = individual.rate(HEIGHT_INDEX);
...                                                      // log-density seeded too
```

So the rates it leaves stored describe the state *before* the mortality and
log-density seeding, and `mortality_dt` reads cumulative mortality. It is the
same shape as the defect fixed in #585 and traitecoevo/odelia#45, one level down.

**Currently harmless through the solver.** Once `Patch::ode_rates` computes
(#589), the patch recomputes before the solver ever reads
these, so the stale vector is never integrated. It is observable only by a caller
that reads a `Node`'s stored rates directly — which is what
`tests/testthat/test-patch.R:86` does, and why that assertion had to relax from
`expect_identical` to `expect_equal`: the discrepancy is the last bits of a
heartwood rate at ~1e-22.

**Worth fixing anyway**, because "the stored rates describe the stored state" is
either an invariant or it is not, and a future caller reading a newborn's rates
has no way to know. Either recompute after the seeding writes, or seed the states
before computing.

PR #589, opened 2026-08-03
Compute rates where they are read, not where state is set

dfalster:
The structural version of #585's introduction fix, as discussed there. **Stacked on #585** — it deletes a line that PR adds, so it is branched off it and the diff shows only its own three commits once #585 lands. **Draft because it is blocked on traitecoevo/odelia#46**, which it needs to compile.

## The problem

`ode_rates()` returned whatever the last `set_ode_state()` had computed, so any caller that arrived at a state by another route read rates belonging to a different one. `Patch::introduce_new_nodes` widened the state and rebuilt the light field without recomputing; odelia's `set_state_from_system` then seeded `dydt_in` from those stale rates and marked them clean, and under first-same-as-last that vector became `k1` of the next step. #585 fixes that one path. This removes the possibility rather than the instance: `compute_rates()` moves out of both `set_ode_state` overloads and into `ode_rates()`, so the rates are always those of the state the patch currently holds, and the `compute_rates()` #585 added becomes unnecessary rather than merely correct.

It also separates loading a state from evaluating at it, which is what a reverse pass wants — re-establish a stage's state without paying for a rate evaluation the recordings will redo.

## What the review asked for

> Please demonstrate the bit-identity rather than assert it.

**Bit-identical to #585 on every case measured** — same offspring production to all 17 digits, same accepted step counts:

| | #585 head | this branch |
|---|---|---|
| FF16, one species | `16.88950160654473` (271 steps) | identical |
| FF16, two species | `11.995204438595735`, `16.474987922830458` (293) | identical |
| K93 | `0.12811441134553064` (345) | identical |
| TF24, full patch lifetime | `42.815680329152386` (5 913) | identical |
| FF16, schedule refinement | `17.200884309431157` | identical |

As predicted from the call graph: the forward path calls `set_ode_state` then `ode_rates`, so the same work happens in the same sequence, and the *count* is unchanged — `set_ode_state` loses a compute and `ode_rates` gains one; `introduce_new_nodes` loses one and `set_state_from_system` gains one.

**Runtime unchanged.** TF24 at full patch lifetime, six alternating reps per build to cancel drift: mean **44.26 s** (#585) vs **44.36 s** (this branch), against a 1.5 s spread *within* each build (43.65–45.13 vs 44.01–45.17). The 0.2% gap is noise, which is what the unchanged call count predicts.

> It won't hold by construction unless the environment refresh moves with it.

Correct, and this PR does not claim otherwise. What holds by construction is that **rates always describe the state just set**. The environment build stays with the mutating path, and the two paths still differ — `set_ode_state` uses `compute_environment(true)`, `introduce_new_nodes` uses `(false)`. Unifying that is a separate change, and deliberately not bundled here: it would move numbers, and this one must not. Filed as a follow-up if wanted.

The claim is also bounded one level down: `Node::compute_initial_conditions` still computes rates before its own seeding writes. Harmless through the solver now (the patch recomputes before anything reads them), observable only to a direct reader of a `Node`'s stored rates. That is #588.

## A latent bug this exposed, fixed in the first commit

Moving the compute into `ode_rates()` made `patch$ode_rates` segfault on a fresh R-constructed patch. `Patch` held `environment_type* environment_ptr` pointing at its own `environment` member, with default copy semantics — so every copy's pointer went on referring to the *source's* members. `scm.h` does `patch = sys`, and RcppR6 hands copies to R. It has been survivable only because the sole dereference was `compute_rates()`, which every live path reached on an object whose source was still alive; the first new caller to deref it crashed. Replaced with an index, which survives a copy intact. Same environment selected, no arithmetic changed — the bit-identity above is measured with this in.

## Testing

Full suite: **2 569 pass, 0 fail, 6 skip** (macOS, R 4.6.1), built against odelia#46.

Two re-recordings, in their own commit:

- The three FF16 "offspring arrival" assertions — the numbers #585 asks to re-bless, unchanged from that branch. Committed so this branch is green on its own; identical to whatever #585 lands, so they will not conflict.
- `test-patch.R:86` relaxes to `expect_equal` for the ~1e-22 heartwood slot (that is #588), and line 96 tightens the other way: "should be identical, but are merely equal" becomes identical, because reading the rates and asking for the derivative of the same state now go down one path.

---
_Drafted with [Claude Code](https://claude.ai/code)_


aornugent (approved):

aornugent:
I picked up these changes on #585 - sorry for the double up.

Issue #575, opened 2026-07-28
[TF24 hydraulics] Dryland patches fill with cohorts born at exactly zero density that then grow to canopy height


dfalster:
Split out of #571, where the crash was fixed (#574) and this was left. This is the
*ecological* half, and it is what stops a dryland site giving an interpretable answer.

## What happens

In the #571 dry start (5 layers at θ = 0.005, below the 0.01 residual, 1 m/yr
rainfall, TF24, 10 yr), **59 of 93 cohorts are born at exactly zero density and then
grow to ~15 m.**

Not "die" — *born* dead. `Node::compute_initial_conditions` sets the boundary
condition

```cpp
const double g = individual.rate(HEIGHT_INDEX);
// NOTE: log(0.0) -> -Inf, which should behave fine.
set_log_density(g > 0 ? log(birth_rate * pr_estab / g) : log(0.0));
```

so any cohort introduced while height growth is ≤ 0 gets `log_density = -Inf`,
i.e. density *exactly* 0 (no underflow involved). The next lines then pin
`log_density_dt = 0`, so it can never recover. Measured across snapshots at
lifetimes 1/2/4/6/9: **zero cohorts ever transition from finite density to -Inf.**
Every ghost is inert from birth.

They then grow anyway, because height growth does not depend on density. 53 cohorts
are introduced in the first 0.04 yr into dry soil, are all born inert, and by year 9
are stacked at ~15.7 m.

## Why it matters

1. **The answer is uninterpretable.** A genuine ecological failure to establish and
   a patch full of zombie cohorts look identical from the outside. #571 noted this;
   the measurement now names the mechanism.
2. **It was the upstream cause of the crash.** Cohorts converging into a band
   ~2.6e-5 m wide (against the refiner's 1.5e-5 m floor) is what let them invert in
   height, which broke the quadrature grid — see #574. That fix makes the integral
   correct regardless; it does not make the distribution meaningful.
3. **Most of the compute is spent on inert cohorts.** 59 of 93 nodes carry zero
   weight but are still integrated, still refined against, and still occupy
   quadrature points.

State after #574 (the crash fix), TF24 dry start, 10 yr:

| | FF16, 40 yr | TF24 default, 10 yr | TF24 dry start |
|---|---|---|---|
| live cohorts | 108/108 | 84/93 | **34/93** |
| tallest 50 span | 0.775 m | 3.263 m | **0.049 m** |
[...]

Issue #566, opened 2026-07-22
[TF24 hydraulics] Leaf thermal damage, repair & acclimation (ATLS)


dfalster:
## Summary

Add a mechanistic leaf **thermal damage / repair / acclimation** layer (the ATLS
model) on top of the Penman-Monteith leaf temperature from #523. PM supplies a
physically-consistent `Tleaf`; this epic consumes `Tleaf` to track protein-pool
damage and downscale photosynthetic capacity, capturing heat-driven loss of
function and acclimation that reversible Arrhenius scaling cannot represent.

**Depends on #523 (Penman-Monteith).** PM is a hard prerequisite — the damage
kinetics are driven by `Tleaf`, which only becomes physically meaningful once the
leaf energy balance is in place.

## Model

From the ATLS leaf demo (`ATLS_leaf_demo`) and the reconciliation note that ships
with the PM work.

**Core (Lumry–Eyring):** the damage kinetics are a Lumry–Eyring protein scheme —
a three-state, closed pool

- functional/native `N`, reversibly damaged `D`, irreversibly damaged `I`
  (`N + D + I = 1`),

with a *reversible* unfolding equilibrium `N ⇌ D` feeding a slow *irreversible*
step `D → I`. The defining feature of Lumry–Eyring (Lumry & Eyring 1954) is that
irreversible damage `I` forms **only via the reversible intermediate `D`**, never
directly from `N`. This intermediate-gated `N ⇌ D → I` topology *is* the model;
it's the minimal standalone unit to port and test first.

**Extensions layered on the core (separable):**

- an acclimation state `A` that raises the critical temperature `T_crit(A)` — a
  genuine extension beyond classic Lumry–Eyring, addable as its own layer;
- temperature- and acclimation-dependent Arrhenius rates with smooth damage/repair
  switches — standard kinetics parameterising the Lumry–Eyring rate constants.

**Coupling to carbon:** photosystem-II damage downscales electron transport via
`jmax → jmax·N` at TF24's `electron_transport()` / `jmax_` — a downstream consumer
of the pool state, not part of the damage kinetics itself. PM feeds the
*reversible* Arrhenius response of `jmax_`; ATLS adds a second,
hysteretic/*irreversible* multiplier `N` on the same term.

Note: ATLS's "detailed" energy balance (Sperry et al. 2017) is algebraically the
same steady-state balance as PM's `Tleaf`, just written in conductance rather than
resistance form — so ATLS should adopt the #523 energy balance rather than its
[...]

PR #590, opened 2026-08-03
Prototype of age integration

aornugent:
I went down a rabbit hole trying to understand how our size structured solver interacted with NSC storage. In part, we now have a model where growth rates are not *only* determined by size. There's a long write up at: [reports/13-carried-state-invalidates-the-compression-term.md](https://github.com/aornugent/plant-dev/blob/claude/plant-storage-scm-analysis-efiolz/docs/reports/13-carried-state-invalidates-the-compression-term.md)

This PR introduces: `Control$node_density_in_birth_date` as a toggle to switch between size or age based integration


dfalster (commented):
## Overall: yes, and I think it's more important than the framing suggests

Strong agreement with the direction. The write-up sells this as a fix for the NSC storage interaction, but it's a bigger structural win than that, and worth saying so explicitly because it changes how much of the roadmap it unblocks.

The point I'd add: **`plant` has always been a characteristics method** — the SCM *is* the method of characteristics — and `∂g/∂h` was the last vestige of the Eulerian formulation the model was derived from. Carrying the density along characteristics (as in the usual de Roos presentation, which is what we implemented) versus carrying cohort *number* (the Escalator Boxcar Train variant) were always the two options. We took one; this PR takes the other, and it's the one that fits a Lagrangian solver. So this isn't bolting a special case onto the solver, it's removing a seam.

Things that fall out for free, beyond the storage fix:

- **The `g > 0` boundary special case disappears.** The division by `g` in `eq-bc1`, the "0 otherwise" branch, and the requirement that `S_G/g → 0` are all artefacts of the height coordinate. Birth density is `birth_rate·pr_estab`, finite always. The zero-density cohorts we'd concluded were expected SCM behaviour (#575) are ordinary cohorts occupying a compressed band of height space.
- **The density runaway cannot happen.** `check_finite_node_densities`/the ceiling of 50 exist because `N ~ 1/|∂h/∂τ|` can blow up. Your measured max of +2.4168 in height against −0.0015 in birth date is the bound, not a coincidence. Worth adding: in a *stationary* patch the height-coordinate density genuinely diverges as `N ~ (h_max−h)^(d/c−1)` when `d < c` (mortality below the growth-shutoff gradient), and `log_density_dt → c−d > 0`, so `log_density` grows without bound. Your runs cap at 105 years so you don't see it, but any long run with a low-mortality strategy is guaranteed to trip that ceiling — and it isn't an error, it's the correct solution in a coordinate where it's singular.
- **Age is monotone by construction; size is monotone only by assumption.** This is the sharpest version of the crossing-characteristics finding. `dα/da = 1 > 0` always, so cohorts can never overtake in age. Storage-gated growth can violate the size assumption; it cannot violate this one. The fragility was inherited from the coordinate, not the biology.
- **It's dimension-agnostic.** `∂log ν/∂a = −d` is unchanged for 1 state or 20, because the label space is the dimension of the *birth manifold*, not the state space. Given where TF24 is heading with acclimation and further NSC states, that matters more than the current fix.

One scoping note for whenever this reaches the theory docs or a paper: the equivalence needs deterministic characteristics from a finite-dimensional birth manifold. `plant` satisfies that; structured population models generally don't (anything with stochastic individual growth has no characteristics, and age no longer determines state). Worth stating so the claim doesn't get over-generalised.

## Code review

What's right and easy to miss:

- The `-height` abscissa is **bit-identical**, not merely close: `(−h0) − (−h1)` equals `h1 − h0` exactly in IEEE754. Good trick.
- The `SpeciesBase::control()` diagnosis is correct — `get_control` appears nowhere in the tree and `Control control` is public on `Strategy`. Genuinely dead code that had never instantiated.
- Routing schedule refinement through the same abscissa (`r_compute_competition_effect_by_nodes_error`) is the easy thing to forget, and the `std::abs` argument for why the height branch is unchanged holds.

Issues, most to least important:

1. **The branch was 6 commits behind `develop`, and #574 conflicts directly.** `develop` has since added `compute_competition_unordered` for inverted heights — which integrates in **height**. A naive merge sends the birth-date coordinate down it, silently swapping coordinates whenever cohorts cross. That's exactly the regime your seasonal-stress run measures (66 crossings / 10,011 pairs).
2. **The height-based early exit survives unchanged.** `if (h0 < height) break;` assumes decreasing height. As you note in Appendix F, this doesn't error — it silently discards shaded cohorts and overestimates growth. `develop` already has `heights_are_decreasing()`/`scan_heights()` for precisely this.
3. **Nodes can share a birth date.** `Patch::set_initial_state()` and `r_set_state()` build nodes as copies of the boundary node, and birth dates aren't part of the ODE state, so a patch seeded or imported *without* per-node times gives every node the same one. On the birth-date axis that's zero-width intervals throughout — the competition profile goes to ~0 silently. A faithful `export_patch_state()` round-trip carries the times and is fine.
4. **The boundary node's abscissa was one derivs call stale.** `compute_rates()` stamps it, but the stepper calls that *after* the `set_ode_state()` that rebuilds the environment. Small in practice (see below) but it made the spatial quadrature a function of the ODE step size.
5. **The PR fails two existing tests.** `test-model-version.R`'s drift guard, for TF24 and TF24f — a new `Control` field changes each strategy's recorded scientific surface.
6. Minor: the control lookup sits inside `compute_competition`'s loop, which runs once per spline knot per RK stage; and `f_h1`/`f_h0` now mean "f at abscissa x".

## Our additional commit

Pushed `bf1ec46a` to this branch: merges `develop` and addresses the above. Full suite green (0 failures, 5 pre-existing skips); model-version snapshot accepted with **no** scientific version bump, since no default moved. New `tests/testthat/test-density-coordinate.R`, 21 assertions, built around convergence rather than fixed tolerances — that's what actually catches a stale or duplicated abscissa.

Three corrections to my own review, in fairness:

- **Your numbers reconcile, and your NEWS direction is right.** The `1.2e-3` (FF16) and `1.5e-4` (K93) figures are at *two schedule interleavings*, not the default — I measured `1.0e-2, 4.2e-3, 1.2e-3, 2.9e-4` (FF16) and `2.6e-3, 5.9e-4, 1.4e-4, 3.5e-5` (K93) over successive halvings, so your third row matches exactly. Only the refinement level was unstated. I'd add the finding that the gap is almost entirely the *height* coordinate's error: FF16's birth-date answer moves 18.6956 → 18.6996 across an 8× refinement while its height answer climbs 18.5071 → 18.6941 toward it. At the default schedule birth date is ~50× more accurate for the same node count. That's a stronger selling point than agreement.
- **I overstated issue 4.** A/B'd it by disabling the fix: FF16 offspring production moves from 6.879e-07 to 6.880e-07. The mechanism is real, the magnitude is negligible — the boundary node carries almost no leaf area. Kept the fix (it removes a step-size dependency, and the whole integral *is* that segment while a species has one node), but issue 3 is the one that can actually zero a profile.
- **One finding of mine was simply wrong.** I thought species could disagree about the coordinate, since it lives on `Control` per strategy. `Patch::add_strategies()` overwrites every strategy's control with the patch's, so it's unreachable. I wrote a guard, found it dead, removed it, and pinned the normalisation with a test instead.

Two things back to you:

- **Could you share the configuration behind TF24's 42.1 / 400.9?** I couldn't reproduce it. At `lma = 0.0825`, `hmat = 5`, `birth_rate = 20`, `max_patch_lifetime = 30` I get height 474 → 587 → 696 against birth date 3495 → 3729 → 3840 over the same three schedules — same direction, ratio 7.4 → 6.3 → 5.5, but neither series converged. I've put mine in NEWS with the config stated, because the durable claim is the *diagnostic*: refinement doesn't close the TF24 gap the way it closes FF16's, and that distinction is what separates a wrong compression term from a coarse quadrature. Happy to swap yours back in with its config attached.
- **The remaining Appendix F items are yours** — the two stochastic-path fixes (`StochasticPatch::ode_size`/`set_ode_state`/`ode_rates` environment forwarding, and `StochasticSpecies::introduce_new_node`'s missing `set_initial_states()` for the birth reserve fill). I've left those alone.

And a consequence someone needs to own before this flips default-on: the ATLS thermal-damage work (#566) was built and evaluated on the uncorrected solver, including the conclusion that the large TF24t-vs-TF24 R0 gap is a fixed-birth-rate density-dependence artefact rather than a bug. If recruitment-phase offspring production was off by the factor you're measuring, that conclusion needs re-deriving on the corrected solver.

Also worth a multi-species run before default-on — everything here is single-species, and crossing characteristics plus the summed light profile is where two strategies would interact.

aornugent:
### All review items are closed. Two things moved to their own PRs.

Your six issues were closed by your own commits. Thank you — items 1 to 6 and the multi-species run.

**The TF24 configuration**. You could not reproduce it because the pair was wrong. 42.14 is the value at 141 introductions; 400.92 is the value at 561. No single run gives both. At the default schedule the pair is 42.14 / 395.44. Configuration: lma = 0.1978791, birth_rate_y = 1, max_patch_lifetime = 105.32, everything else default, at 141dc8df. On this branch tip the same configuration gives 42.5772 and 395.4377 — the height arm moves +1.04%, the birth-date arm −0.001%.

**Both series converge**. Extended to 1121 introductions and extrapolated: height reaches 60.294, birth date 401.722. The limits differ by 6.66 times, so the height coordinate is 85.0% low. FF16's two limits agree to 5.3e-5, K93's to 5.6e-7. This is stronger than "refinement does not close the gap", because it removes the reply that the solver needs more nodes.

**Your correction is accepted**. The 1.2e-3 and 1.5e-4 figures were at two interleavings, not the default. Your point that the gap is almost entirely the height coordinate's error is the better selling point, and I have used it.

**Cost**. The corrected coordinate's accuracy saturates near 200 nodes. schedule_eps = 0.02 buys 357 nodes and no more accuracy than 0.2 buys with 203. At 0.2 it is 1.87 times faster at equal accuracy. Worth changing the default for TF24 in this coordinate.

**test-density-coordinate.R was 529 s, half the suite**. Now 26 s, and it guards more: 43 expectations to 251. Run cost is roughly cubic in node count, about 9× per level, so every ladder was dominated by its last rung — and one single-species block was 173 s on its own, an FF16 run at eight times the default schedule. Trend blocks now assert the per-halving ratio over two levels instead of averaging over three, which is the same claim or stronger. Five new O(1) guards replace inference from a trend with direct assertion. Two mutations confirm they fail when the defect is present.

**The two stochastic fixes are done, in their own PR**. They change TF24's stochastic numerics and need its seed-pinned baselines regenerated, so they do not belong here. Soil water now follows the deterministic trajectory to 3e-4 and leaf area comes within 1.4% at ages 1 to 3, from 11.6–16.0% low. FF16 and K93 stay bit-identical. A third defect and a fourth surfaced while fixing them, both included.

dfalster:
yay, nice work!

Issue #584, opened 2026-07-30
[TF24 hydraulics] The collar bracket's clamp to root_psi_crit is dead code (sign slip in std::max)


dfalster:
Found while diagnosing #576, not fixed there because it changes results for runs that currently complete.

## The code

`Leaf::prepare_collar_solve` ([src/leaf_model.cpp](https://github.com/traitecoevo/plant/blob/develop/src/leaf_model.cpp)):

```cpp
bound_a = -root_zero_E;
bound_b = std::max(-root_crit, -root_psi_crit);
```

The block comment above the master solver says the collar potential is bracketed between `root_zero_E` and `root_crit`, **"clamped to `root_psi_crit`"**.

It is not clamped. `root_crit` is a **signed (≤ 0)** potential — `find_root_psi` searches `[-psi_crit, wettest_soil_layer]`, both ≤ 0 — so `-root_crit` is a **positive magnitude**, while `-root_psi_crit` is **negative** (`root_psi_crit` is stored as a positive magnitude, 5.870283 at the TF24f defaults). So `std::max` compares a positive against a negative and always returns `-root_crit`. The clamp never binds.

Reading the surrounding sign conventions, the intent looks like:

```cpp
bound_b = std::min(-root_crit, root_psi_crit);
```

## It is reachable

`-root_crit ≤ psi_crit`, and for TF24f `psi_crit = 5.91988 > root_psi_crit = 5.870283`, so there is a ~0.05 MPa window in which the dry end of the bracket exceeds the root-side critical potential and the intended clamp *would* bind. Measured on a bare `Leaf` at the TF24f (`lma = 0.0825`) parameters, varying `psi_soil`:

| psi_soil (MPa) | `-root_crit` | exceeds `root_psi_crit`? | `bound_b` now | intended |
|---|---|---|---|---|
| 5.0 | 5.203966 | no | 5.203966 | 5.203966 |
| 5.5 | 5.586930 | no | 5.586930 | 5.586930 |
| 5.8 | 5.826764 | no | 5.826764 | 5.826764 |
| **5.9** | **5.907738** | **YES** | **5.907738** | **5.870283** |

So in that window the golden-section search is handed a dry endpoint past the potential at which root conductivity is down to 5%, which is exactly what `root_psi_crit` exists to exclude.

## Why it is worth closing deliberately

[...]

Issue #47, opened 2013-12-12
Move scheule building code into tree


richfitz:
Currently in `scripts/build_schedule-fun.R`.

This is just a reminder for me before I start relying on that code elsewhere...

PR #591, opened 2026-08-03
Move the leaf model out to phylloptim

dfalster:
plant's leaf gas-exchange and hydraulics model moves out of this repo.
`src/leaf_model.cpp` is deleted and `inst/include/plant/leaf_model.h` becomes a
shim aliasing `plant::Leaf` to `phylloptim::Leaf`, consumed via `LinkingTo`. The
generated RcppR6 glue names `plant::Leaf` throughout and is unchanged.

TF24 and TF24f only; FF16 and K93 are untouched. Offspring production moves
−0.20% to +0.23%, inside the tolerances those tests already carry.
`scientific_version` goes 4 → 8, because two changes are user-visible:
`opt_root_psi` is now the positive magnitude, so anything downstream that
negated it must stop; and the root-collar bracket is now genuinely clamped to
`root_psi_crit`, which changes dry-corner behaviour in a 1.2 MPa window (#584).
`atm_kpa` defaults 100.5 → 101.3 kPa, matching the constant the leaf's
ppm→Pa conversion always used.

Requires phylloptim >= 0.1.0 and odelia >= 0.2.1, pinned both by a `LinkingTo`
floor and by a `Remotes` sha. Breaking changes have `old -> new` lines in
NEWS.md, and agents.md gains §3.4 recording where the leaf model now lives.

Supersedes #567.


dfalster:
CI here is red for a reason that has nothing to do with the swap, and it will
stay red until someone makes a call that is not mine to make.

```
* leaf=traitecoevo/leaf_cpp@0709eb1: ! pkgdepends resolution error
Caused by error:
! Can't find GitHub repo traitecoevo/leaf_cpp.
```

**`traitecoevo/leaf_cpp` is private. `plant` and `odelia` are public.** So
`setup-r-dependencies` on plant's runners, using the default `GITHUB_TOKEN`
scoped to this repository, cannot see it — on every platform, before a single
line is compiled. `windows-latest`, `macOS-latest` and `lint` all fail at the
same step, which is why the failures look broad and are actually one thing.

Worth being clear that this is not the `@sha` pin or the `name=owner/repo`
rename syntax misbehaving. Both are fine; the repo is simply not visible.

Two ways out, and they are quite different commitments:

1. **Make `leaf_cpp` public.** It is where the other two already are, it is what
   the package is for, and leaf_cpp#1 (deciding the name) and PLAN item 14
   (publication framing) are the surrounding questions. Effectively
   irreversible in the sense that matters — the history becomes public — so it
   is a decision, not a fix.
2. **Give this workflow a PAT with read access to `leaf_cpp`**, as a repository
   secret, and pass it as `GITHUB_PAT`/`GITHUB_TOKEN` to
   `setup-r-dependencies`. Keeps leaf_cpp private, costs a secret to rotate, and
   means anyone forking plant cannot build it.

Until one of those happens, this PR cannot go green, and leaf_cpp#33 and #34 —
the two stages of the R-interface work that are coupled to this branch — stay
blocked behind it.

Everything else about the branch is in good shape: it compiles and passes
locally against leaf `master`, and leaf_cpp#30 has since landed the R interface
on top without touching what plant consumes.

dfalster:

Note the open question below on the `atm_kpa` default — **still needs your call**, it has not been answered.

---

plant's leaf gas-exchange and hydraulics model moves out of this repo.
`src/leaf_model.cpp` is deleted and `inst/include/plant/leaf_model.h` becomes a shim
aliasing `plant::Leaf` to `phylloptim::Leaf` from
[traitecoevo/phylloptim](https://github.com/traitecoevo/phylloptim), consumed via `LinkingTo`.
Nothing else in plant changes shape: the ~17k lines of generated RcppR6/Rcpp glue name
`plant::Leaf` throughout and keep compiling unchanged.

Also supersedes #567 (Penman-Monteith), now closed — the PM model lives in the leaf package
and its tests were ported there.

## What this changes for a plant user

TF24 and TF24f only; FF16 and K93 are untouched.

**Results.** Offspring production moves −0.20% (1 species) and +0.23% (2 species), inside
the 2e-2 tolerance those tests already carry, so both keep `develop`'s pinned values. A
seeded stochastic run produces the **same 103 individuals born and 28 alive** as `develop` —
identical integers, not a tolerance match. All 8 dry-scenario gateway classifications are
unchanged.

**`scientific_version` 4 → 8**, because two of the changes are user-visible:

- **The `opt_root_psi` aux changes sign.** It is now the positive magnitude, matching
  TF24f's `opt_root_psi_state`, which always held it. Previously the aux reported the signed
  potential while `tf24f_strategy.cpp` negated it back for the state — one quantity under two
  conventions. Anything downstream that negated the aux must stop.
- **Dry-corner behaviour changes.** The root-collar bracket is now actually clamped to
  `root_psi_crit`, the potential at which root conductivity is 5%. The clamp compared a
  magnitude against a signed potential and could never bind (#584), so the solver optimised
  over a collar the root system cannot supply. The window is 1.2 MPa wide at TF24's defaults.

**`atm_kpa` defaults 100.5 → 101.3 kPa, and I would like your call on this.** The leaf's
ppm→Pa conversion was the hard-coded constant `0.1013` — which *is* 101.3 kPa
(`1e-6 × 101300 Pa`). The driver said 100.5. So TF24's conductance side responded to 100.5
while Γ\*, Kc, Ko, Km and the ci root-find bounds assumed 101.3. The leaf package now derives
the conversion from `atm_kpa`, making the model consistent at any pressure, which is what
makes the default worth fixing too. `100.5` entered in `34d46ac2` (#446), an interface
refactor that does not mention atmospheric pressure, with no rationale recorded anywhere,
while every leaf-level test used 101.3 — so it reads as an artefact rather than a site
elevation. **Say if it was deliberate.** `atm_kpa` is still a driver; set it per site for
altitude.

## Dependencies

Requires **leaf >= 0.1.0** and **odelia >= 0.2.1**, pinned two ways:

```
LinkingTo: Rcpp, BH, odelia (>= 0.2.1), leaf (>= 0.1.0)
Remotes:   ..., traitecoevo/odelia@806924e, traitecoevo/phylloptim@0709eb1
```

The floor guards whatever is installed (R checks it at install, so a mismatch stops at
dependency resolution rather than inside a header); the `@sha` fixes what gets fetched, so a
merge in a sibling repo cannot silently change plant's numbers. plant previously declared
neither, which is how it came to require an odelia commit it could not name: #585 made
`Patch::ode_rates` non-const, and odelia 0.2.0's `r_ode_rates` takes the system by `const&`,
so `develop` does not build against released odelia either.

## Two things to know before reading the test diff

**`root_carbon_per_leaf_area` is `mass_root_prop` divided by `area_leaf`.** The leaf model is
purely intensive now. In plant's C++ that division cancels exactly — `mass_root()` is
`pars.a_r1 * area_leaf`, so the per-leaf-area value is `root_mass_carbon_scale * pars.a_r1`,
bit-identically. In tests that pass the two separately it does not cancel, and getting it
wrong compiles and runs: it leaves a root system 20× too weak at `area_leaf = 0.05`.

**One test leaf was incoherent and the clamp fix exposed it.** It paired roots that lose
conductivity by 1.95 MPa with soil at 2.0 MPa — drier than those roots tolerate — and had
been relying on the dead clamp to operate at all. Now solved at a soil potential its roots
can reach.

## Migration

Every breaking change has an `old -> new` line in NEWS.md under "Breaking changes", in the
format the `plant-update-interface` skill reads. `agents.md` gains §3.4, "The TF24 leaf model
lives in another repo", so the next person hunting a leaf bug finds the pointer rather than
the absence.



---

## Update: the leaf package is now `phylloptim`

Renamed in phylloptim#47 (decision on phylloptim#1). `leaf` is taken on CRAN by an unrelated
symbolic-regression package, and `leaf_cpp` was never a legal R `Package:` name — R allows
only letters, numbers and dot, so the underscore fails `R CMD check`. The new name is
*phyllo-* (leaf) + optimality.

For this PR that is a small, contained change, because the coupling was already funnelled
through one shim: `inst/include/plant/leaf_model.h` now does `#include <phylloptim.hpp>` and
`::phylloptim::` instead of `::leaf::`. `plant::Leaf` and the re-exported constants keep their
names, so nothing downstream of the shim moves and the generated glue is untouched.

`LinkingTo` and `Remotes` follow. **The `Remotes` pin currently points at phylloptim's rename
branch** (`traitecoevo/phylloptim@6b73490`) because #47 is not merged yet — re-pin to a master
sha before this merges.

Re-verified after the rename: plant builds clean against the renamed package, and the TF24 and
leaf tests are **438 passing, 0 failures**, with the same 3 pre-existing skips.

dfalster:
## The CI blocker above is retired, and the branch now merges

Three things changed on the branch; it went `CONFLICTING` → `MERGEABLE`.

### 1. The private-repo blocker no longer exists

My earlier comment said CI would stay red until someone decided whether to make
`leaf_cpp` public. **phylloptim#47 settled that by renaming**, and
`traitecoevo/phylloptim` is public — so the default `GITHUB_TOKEN` can see it and
no PAT is needed. Confirmed on this push: `setup-r-dependencies` is
`success` on both `macOS-latest` and `windows-latest`, which is the exact step
that used to fail on every platform before a line was compiled. Neither of the
two "quite different commitments" had to be made.

### 2. Merged `develop` — two conflicts, both in generated or blessed files

* **`R/RcppR6.R`** — a generated-glue hash. Regenerated from the merged
  `inst/RcppR6_classes.yml` with RcppR6 0.2.4 rather than picking a side, so it
  carries both develop's age-integration methods (#590) and this branch's `Leaf`
  surface. `src/RcppR6.cpp` and the `RcppR6_*.hpp` headers regenerated
  *unchanged*, which is the check that the automatic merge of those was right.
* **`tests/testthat/_snaps/model-version.md`** — the two sides moved different
  parts of the same TF24/TF24f rows. Resolved as the union: develop's
  `control.node_density_in_birth_date`, this branch's `pars.d` and
  `pars.use_energy_balance`, `TF24@v8` / `TF24f@v8.1`. FF16 and K93 merged
  cleanly and already agreed.

### 3. Re-pinned phylloptim to a master sha, as the description asked

`6b73490` was the rename *branch* head. #47 squash-merged as a **new** commit, which
orphaned it — it still resolves through the API, so nothing looks broken, but it is
on no branch and one gc from unresolvable. Now `@818ad43`, master's head.

That is two commits further on than what this branch was measured against, and the
difference is confined to gradient work: #45 adds entry points, #46 routes the stem
spline through accessors that are the identity at `stem_b == stem_b_spline_`. plant
never calls `perturb_stem_b`, so the forward model is untouched — and measured that
way rather than assumed.

### Re-verified after the merge and the re-pin

| run | result |
|---|---|
| `test_check("plant")`, `NOT_CRAN=true` | **2846 pass, 0 fail**, 2 intentional skips |
| scenario gateway, `PLANT_RUN_SCENARIOS=1` | **43 pass, 0 fail, 0 skip** — no classification moved |

⚠️ Worth knowing: a bare `test_check("plant")` reports a confident-looking
`FAIL 0 | PASS 2825` while `skip_on_cran()` silently skips **`test-model-version.R`** —
all four `scientific_version` snapshots, i.e. precisely the conflict resolved above —
along with the PM demo tests. `NOT_CRAN=true` is what makes them run.

### Also fixed: documentation this branch had already falsified

The shim header still warned that `atm_kpa` defaults to **100.5**, which `4a2091b4`
changed to 101.3 earlier on this same branch. Several places still called the
dependency `leaf`, one with link text `leaf` pointing at the phylloptim URL, and the
namespace was still written `leaf::Leaf`.

### Two things still yours to decide

1. **The `atm_kpa` question is still unanswered.** The branch implements 101.3 and the
   evidence reads as an artefact of #446 rather than a site elevation — but endorsing
   that is your call, not something to infer from silence.
2. **`overstorey_staging/`.** #563 removed `overstorey-staging/` (hyphen); this adds
   `overstorey_staging/` (underscore), needing a *new* `.Rbuildignore` entry because
   the name change dodges the old rule. Per #563 the convention was "stage while the
   branch is in flight, then move to Overstorey and delete" — so a staging copy added
   *as* the branch lands might belong in Overstorey directly. Dev-only and not
   shipped, so cheap either way. Left as-is deliberately.

dfalster:
Answered by Daniel: phylloptim was asked to set a default matching the prior values used in plant. Checked which prior value that is, because plant held two that disagreed — and it is **101.3**, so what this PR does is what was asked.

On `develop` today:

| location | value | what it governs |
|---|---|---|
| `inst/include/plant/leaf_model.h:58` | `umol_per_mol_to_Pa = 0.1013` (= 101.3 kPa) | Γ\*, Kc, Ko, Km, the ci root-find bounds |
| `inst/include/plant/leaf_model.h:259` | `atm_kpa = 101.3` | plant's own leaf-model default |
| `inst/include/plant/models/tf24_environment.h:88` | `atm_kpa = 100.5` | the TF24 driver |

So 101.3 is plant's prior value in **two** places, including the leaf model's own signature default, and 100.5 appears in exactly **one** — the TF24 environment driver, where it entered in `34d46ac2` (#446) with no recorded rationale. It disagreed with plant's own leaf-model default in the same repo, not merely with phylloptim.

That makes the driver constant the artefact rather than a site elevation, which is what the measurement already indicated: pinning 101.3 moved the leaf swap from +2.4% to −0.20%.

**This question is closed — no change needed to the PR.** `atm_kpa` remains a driver, so set it per site for altitude.

dfalster:
Filed the `overstorey_staging/` question as #596 — the demo goes to Overstorey, deferred so it does not hold this up.

One thing recorded there worth knowing before it is picked up: `tests/testthat/test-pm-leaf-demo.R` sources the helpers by relative path and `skip_if_not(file.exists(...))`, so deleting the directory will not fail the test — it silently becomes a permanent skip, still reporting green while guarding nothing. That is the exact rot the test was written to prevent, so the move needs a decision about where the smoke test lives, not just a `git mv`.

Issue #482, opened 2026-06-18
[other] Test coverage gaps flagged by TODOs


dfalster:
From the TODO audit in #473. Remaining test-coverage gaps marked in the suite:

- `tests/testthat/test-node.R:76` — "Not done yet" (node tests).
- `tests/testthat/test-stochastic-patch-runner.R:61` — more tests on `collect` output.
- `tests/testthat/test-ode-individual-runner.R:147` — a useful helper to construct splines for … (test utility gap).
- `tests/testthat/test-disturbance.R:63` — "check what this is for" (clarify + cover).
- `tests/testthat/test-strategy-ff16-reference-comparison.R:60,65,87` — "check with Daniel", size-concept issues, and growth-variable checks currently commented out.
- `tests/testthat/FF16_reference/extra.R:179` — hard-coded upper limit in the reference model that could be derived.

Tracking issue: #473

dfalster:
The `test-stochastic-patch-runner.R:61` item ("more tests on collect output") is addressed in PR #506. Investigating it surfaced #498: `run_stochastic_collect()` was silently returning empty output and the `expect_silent` test could not catch it.

The test is now: shape / non-empty / monotonic-time / `is_alive`-dims / finite-height-where-alive checks for all three strategies, plus a reproducibility + seeded-baseline test (fixed n introduced / n alive at the final step) as a trajectory-regression guard.

The other #482 items (test-node.R:76, ode-individual-runner spline helper, disturbance coverage, strategy-comparison commented checks, reference hard-coded param) are untouched.

PR #593, opened 2026-08-04
Integrate the environment in the stochastic solver

### LINKED ISSUE #592: The stochastic solver does not integrate the environment, so TF24 runs at frozen soil moisture
## The defect

`StochasticPatch`'s ODE system is the species alone. `ode_size()`, `set_ode_state()`, `ode_state()` and `ode_rates()` do not chain through the environment as `Patch`'s always have, and `compute_rates()` never accumulates resource consumption, so `Environment::compute_rates()` is never called.

Any environment carrying ODE state is therefore **held at its initial value for a whole run**. TF24's carries nine such states — five soil-moisture layers and four cumulative fluxes.

**FF16 and K93 carry none.** That is why this went unnoticed for so long: the stochastic solver is exact for every strategy it had previously been used on, and wrong only for the one that carries a store.

## What it costs

Stochastic TF24 soil water sits at 0.214 for the whole run while the deterministic solver recharges to 0.3106 and draws it down to 0.2992. A permanently drier stand grows more slowly, so stochastic leaf area at patch ages 1 to 3 ran **11.6% to 16.0% below** the deterministic solver's.

That matters beyond the stochastic path itself: the individual-based solver is the only external check available on the deterministic one, since it tracks individuals and has no density transport term. As shipped it was not a valid check on TF24.

## Three related defects, all in #71

1. **The environment is not part of the ODE system** — the above.
2. **Recruits are born with no initial states.** `StochasticSpecies::introduce_new_node()` computed rates without calling `Individual::set_initial_states()`, which the deterministic path calls from `Node::compute_initial_conditions()`. A TF24 seedling started with an empty carbohydrate store rather than `a_st3` of its capacity. Worth 0.45% of leaf area at age 1.
3. **An empty patch discarded the integrated state.** `compute_environment()` clears the competition profile when no individuals are present, which is right, but TF24's `clear_environment()` also restored the soil states — so once the environment joined the ODE system, that threw away what the solver had integrated, on every derivatives evaluation while the patch was empty.

**Defects 1 and 3 mask each other.** `StochasticPatchRunner::reset()` advances to the first arrival in a single `advance_fixed` step, which for `run_stochastic_collect()`'s default first arrival — uniform on (0, 50) — is one step of up to fifty years. While the state was being discarded that was harmless. Kept, it drives soil water out of range and nothing establishes. Fixing the state handling alone produces nine test failures and a runner that introduces no individuals. That leg now uses `advance_adaptive`.

A fourth, found while fixing the others: `stochastic_schedule()` passed `patch_area` positionally into `stochastic_arrival_times()`'s `delta_t` slot. Arrival rates never scaled with area, and the binning interval was set to the area — at 50 m² a 105-year run had three bins, harmless for a constant birth rate and destructive for a variable one.

## After the fixes

Soil water follows the deterministic trajectory to about 3e-4. Leaf area at ages 1 to 3 moves from 11.6–16.0% below to **within 1.4%**.

FF16 and K93 stochastic runs are bit-identical, checked with `identical()` on the raw collected arrays. TF24's seed-pinned survivor count moves and is regenerated deliberately. No `scientific_version` bump: the strategies' equations and defaults are untouched, the deterministic solver is bit-identical, and the drift guard runs and does not react.

TF24 stochastic runs cost about 40% more, from the stiffer system taking smaller adaptive steps.

aornugent:
Closes #592

`StochasticPatch`'s ODE system was the species alone: `ode_size()`, the three
accessors and `compute_rates()` never chained through the environment as `Patch`'s
always have, so an environment carrying ODE state was held at its initial value for
a whole run. TF24 carries nine such states, FF16 and K93 none, which is why the
stochastic solver looked exact for every strategy it had been used on.

Stochastic TF24 soil water now follows the SCM's trajectory on the same drivers
instead of sitting at its initial 0.214 all run, and leaf area at patch ages 1-3
moves from 11.6-16.0% below the SCM to within 1.4%. Three smaller defects had to go
with it: a recruit's strategy-specific initial states were never seeded, an empty
patch discarded the state the solver had integrated, and `stochastic_schedule()`
passed `patch_area` into `stochastic_arrival_times()`'s `delta_t` slot so arrivals
never scaled with area. `run_stochastic_collect()` now reports the environment as
`env` -- breaking: it was `light_env`, a name nothing produced, so the field was
always NULL. FF16 and K93 runs are unchanged bit-for-bit; TF24 costs about 40% more.
No `scientific_version` bump: the equations and defaults are untouched, and what
changed is one solver's realisation of the declared model.


dfalster:
## Review

I built this branch and ran it. The defect is real, the diagnosis is right, and the fix is in the right place: the stochastic patch's ODE system genuinely was a proper subsystem of the declared model, and appending the environment to it is the same shape `patch.h` has always had. All 225 assertions across the two stochastic test files pass, the full suite (52 files) passes with no failures, and the `test-model-version.R` drift guard passes with `NOT_CRAN=true`.

Two things need attention before merge: one factual claim that doesn't hold, and one behaviour change that isn't in NEWS. The rest is small.

---

### 1. The "bit-identical for FF16 and K93" claim is not true, and the cause is the `reset()` solver switch

Both the PR body and the NEWS bullet say FF16 and K93 stochastic runs are bit-identical. They are not. I isolated this by building the branch twice, differing only in `stochastic_patch_runner.h:143` (`advance_adaptive` vs the `advance_fixed` it replaces), and running `run_stochastic_collect()` with `set.seed(1)`, `patch_area = 1`:

| strategy | `identical(species)` | max abs diff | max rel diff | entries differing |
|---|---|---|---|---|
| FF16 | FALSE | 6.5e-4 | 1.0e-5 | 5453 / 39248 |
| K93  | FALSE | 9.4e-14 | 8.5e-14 | 12614 / 34515 |

The mechanism is the solver's step-size memory, not the state. Only `SolverInternal::step()` writes `step_size_last`; `step_to()`, which `advance_fixed` drives, never touches it, and nothing between `reset()` and the first `advance()` restores it — `set_state_from_system()` resizes and re-reads state and rates but leaves `step_size_last` alone. So the adaptive walk over the (empty, for FF16/K93) system at reset ratchets `step_size_last` up by ×5 per accepted step until it saturates at `ode_step_size_max = 5` yr, and the first real step after the first arrival begins from a rejected 5-year attempt instead of `ode_step_size_initial = 1e-6`. Since the system is empty there, that carry-over is the *only* channel available — which is what makes it certain rather than inferred.

This is not a correctness problem: 1.0e-5 relative sits inside `ode_tol_rel = 1e-4`, so both runs are valid realisations. But the claim is load-bearing for "no `scientific_version` bump", so it should say what's true — something like "FF16 and K93 trajectories agree to within solver tolerance (max 1e-5 relative for FF16); they are not bit-identical, because the adaptive leg leaves a larger `step_size_last` behind."

If bit-identity is worth keeping, the minimal fix is to only take the adaptive leg when there is something to integrate:

```cpp
if (t >= 0.0) {
  if (patch.ode_size() > 0) {
    solver.advance_adaptive({solver.time(), t});
  } else {
    solver.advance_fixed({solver.time(), t});
  }
  patch = solver.get_system();
}
```

Worth also noting that this leg now starts TF24's first real step from a 5-year `step_size_last` too, where before it started from 1e-6. Error control catches it, but the shrink loop can `util::stop("Cannot achieve the desired accuracy")` if it bottoms out, and starting at `step_size_max` makes a first-step rejection certain rather than incidental.

### 2. NEWS.md doesn't mention the `reset()` change at all

The five new bullets cover the environment integration, the recruit seeding, the empty-patch state, the area scaling, and the `env` rename. The `advance_fixed` → `advance_adaptive` switch — the only change in this PR that moves FF16 and K93 output — appears in the PR body under "One interaction worth knowing about" and nowhere in NEWS. Given how often a number in this repo's prose has outlived the run that produced it, this one has earned its own bullet, with the honest statement from (1) attached to it rather than to the environment-integration bullet.

Related, smaller: the `light_env` → `env` rename is marked **Breaking** inside a bullet under "Minor changes & bug fixes", while the file has a `### Breaking changes` section at the top whose stated job is to be read by `plant-update-interface`. It should be listed there too.

### 3. Nothing pins the quantitative content of the new coupling

The new tests establish that the environment's states move, that uptake is positive, and that uptake stays below infiltration. All three would still pass if the coupling were wrong by a constant factor. The two quantities the new code actually introduces are untested:

- the `/ area` normalisation in `StochasticPatch::compute_rates()`, and
- the per-individual (rather than density-weighted) summation in `StochasticSpecies::consumption_rate()`.

`expect_lt(flux[[4]], flux[[2]])` is satisfied by any positive uptake, so a future regression in either would land silently. A cheap test that would catch both: introduce a fixed set of individuals, then assert that the patch's per-area depletion equals `sum(individual$consumption_rate(i)) / patch_area` — or, more coarsely, that doubling `patch_area` with the same individuals halves per-area depletion.

Separately, the PR's headline evidence — stochastic soil water tracking the SCM's own trajectory on the same drivers to ~3e-4, and leaf area landing within 1.4% — is the strongest regression guard available for this fix and isn't in the suite. If it's expensive, it's a good candidate for the `PLANT_RUN_SCENARIOS`-style opt-in gate rather than being left out.

### 4. Two comments now describe code that moved

- `tf24_environment.h:515-518` still introduces `clear_environment()` as "Clear the light profile **and** return the soil moisture and the cumulative flux accumulators to the state the run started from … keeps a second run from silently continuing out of the first run's depleted soil." That is now `clear_state()`'s job, and the comment sits above the function that no longer does it — the most misleading possible placement, since the whole point of the split is that `clear_environment()` must *not* restore state.
- `tf24_environment.h:195` says `initial_states` is "Restored by `clear_environment`."

### 5. Design note on the split (not a blocker)

The PR describes the empty-patch fix as mirroring `patch.h`, but `Patch::compute_environment()` doesn't call `clear_environment()` at all — when the patch is empty it simply skips recomputation and leaves the profile alone. So there was a second option that needs no base-class change: make `StochasticPatch::compute_environment()` skip the clear the way `Patch` does. Adding a virtual to the base `Environment` obliges every strategy's environment to understand a distinction only TF24 currently has. The chosen split is defensible and I'd keep it — the two operations genuinely are different — but a sentence on why is worth having, because the next reader will check `patch.h` for the mirror and not find it.

### 6. Small things

- **`resource_depletion` allocates per derivative evaluation.** `Patch` keeps it as a member, `reserve()`s it in `reset()` and `clear()`s it at the end of `compute_rates()` specifically so the RHS doesn't allocate; the new `StochasticPatch::compute_rates()` declares it function-local. Given TF24 stochastic already costs ~40% more here, this is free to mirror.
- **`std::accumulate` with no `#include <numeric>`.** Compiles by transitive include, and `patch.h` already relies on the same thing, so this is consistency rather than a bug — but the include is one line.
- **The PR body is now stale against the diff.** It doesn't mention two of the five changes (the arrival-area scaling and the `env` reporting rename), and its numbers are superseded: it says TF24's survivor count "moves 28 to 31", while the diff has TF24 at `n_total = 81L, n_alive_final = 3L` and `patch_area` reduced from 50 to 1. NEWS is current; the body is what a reviewer reads first.
- **`patch_area = 50` → `1` in the `#482` baseline** is the right call, but it does change the regime the test covers, not just its size: the same ~105-individual stand is now 50× denser, and the survivor counts it pins (5, 3, 2) are a heavily thinned tail. Those are weak golden values — a real regression could plausibly leave 2 survivors. `n_total` is the informative half of that baseline now; consider asserting something with more signal alongside it (e.g. total leaf area at a fixed age).

---

### Checked and correct

Recording these because they're the places this change could have gone silently wrong:

- **ODE leg consistency across deaths.** `ode_size`, `set_ode_state`, `ode_state` and `ode_rates` all iterate `node_begin()/node_end()` in `SpeciesBase`, which for `StochasticSpecies` is the alive-filtered iterator. The species block's width is therefore the same in all four, so the environment's offset after it is stable when `deaths()` marks a node dead mid-run. A mismatch here was invisible before this PR and would have corrupted soil state after it.
- **`consumption_rates` sizing.** `StochasticPatch::reset()` does not call `resize_consumption_rates()` as `Patch::reset()` does, but `Individual::compute_rates()` self-heals when `vars.resource_size != environment.n_resources()`, and every alive node has had `compute_rates` run before `consumption_rate(i)` is read (both at introduction and at the top of `StochasticPatch::compute_rates`). Safe.
- **The recruit seeding matches the deterministic path exactly** — `set_initial_states()` then `compute_rates()`, the same order as `Node::compute_initial_conditions`.
- **The `clear()` split can't reach the SCM.** `Environment::clear()` now composes to exactly TF24's old `clear_environment()` body, and `Patch` never calls `clear_environment()` alone, so the deterministic path cannot observe the split. FF16 and K93 carry no state either way.
- **The `patch_area` fix is right.** `stochastic_arrival_times(max_time, species, delta_t = 0.1, patch_area = 1)` — the old positional call set `delta_t = patch_area`, which at area 50 gave three bins over a 105 yr lifetime. Both halves of the described defect follow.
- **The flux indices in the new test are correct.** `TF24_Environment::compute_rates` writes the accumulators in order rainfall, infiltration, deep drainage, total uptake, so `flux[[4]] < flux[[2]]` is uptake below infiltration as intended.
- **`n_environment_ode_states` covers exactly the three types** `get_list_of_strategy_types()` returns, so no `[[` lookup can fall off the end.
- **The `env` rename is safe.** No consumer of `light_env` exists anywhere in this repo or the sibling checkouts; the collector was reading a name nothing produced.
- **Merge conflicts are mechanical** — `NEWS.md` and generated `R/RcppR6.R` only, both against #590.

### Verdict

Approve on the substance once (1) and (2) are addressed — the claim corrected or the code changed to make it true, and the `reset()` change written into NEWS. (3) is the one I'd push for beyond that: this fix's value is that the two solvers now agree, and nothing in the suite would notice if they stopped.

---
_Reviewed with [Claude Code](https://claude.com/claude-code)_

dfalster:
Merged develop in and applied the review fixes. Two commits: [`eb3e55d9`](https://github.com/traitecoevo/plant/pull/593/commits/eb3e55d9) the merge, [`f33490d6`](https://github.com/traitecoevo/plant/pull/593/commits/f33490d6) the fixes. PR description rewritten — it was describing the first commit, not the branch. Now `MERGEABLE`.

### The merge

Three conflicts. `R/RcppR6.R` is generated, so I regenerated it from the merged `inst/RcppR6_classes.yml` rather than hand-merging; it carries both #590's birth-date accessors and this branch's `node_ode_size`.

The one worth flagging: **develop's Known-issues list had an entry for the `patch_area`-into-`delta_t` bug this branch fixes**, and it merged cleanly because this branch never touched that region. Removed — it described code that no longer exists.

The test-file conflict was the seeded-baseline rationale. Kept this branch's (the baseline moved to `patch_area = 1`) and folded in what develop's had recorded, because re-deriving the baseline costs something real: at area 50 the TF24 pair was 103/28 either side of the phylloptim move, and that exact match of two seeded integers was a sharper statement than any tolerance check that the swap preserved TF24's behaviour. That equivalence still holds — it just isn't what the test pins any more. Said so in the comment and in the `atm_kpa` NEWS entry that made the original claim, rather than letting it quietly become false.

### The bit-identity claim is now true rather than corrected

I gated the adaptive leg on there being something to integrate:

```cpp
if (patch.ode_size() > 0) {
  solver.advance_adaptive({solver.time(), t});
} else {
  solver.advance_fixed({solver.time(), t});
}
```

Verified the claim properly this time, by building develop plus this branch's `R/stochastic.R` alone — the only changes that touch FF16/K93 RNG draws or field names — and comparing:

| | `identical(time)` | `identical(species)` |
|---|---|---|
| FF16 | TRUE | TRUE |
| K93 | TRUE | TRUE |

So every C++ change here is provably inert for those two, which is a stronger statement than the branch was making before and now actually holds. The reasoning for the guard is in the code, at length, because the mechanism is not guessable from the call site: `step_size_last` is solver state that survives the leg, and only `step()` writes it.

### The uptake arithmetic is pinned

This was the gap I most wanted closed. `expect_lt(flux[[4]], flux[[2]])` is satisfied by any positive uptake, so neither the `/ area` normalisation nor the per-individual summation was tested.

My first attempt was wrong, and the way it failed is worth recording: I assumed a lone recruit is in full sun regardless of patch area, since competition counts leaf area *strictly above* a height. It isn't — TF24 averages light over the crown, so an individual sees its own leaves and area scaling moves its physiology. Two individuals at the same height shade each other slightly for the same reason (measured ratio 1.999971, not 2).

The formulation that works exploits that competition is per unit ground area: N recruits on N m² are in *exactly* the same environment as one on 1 m², so per-area uptake must be invariant — the sum scales with N, the division divides it back out. It holds bit-for-bit, and the test asserts the premise (the individuals' own rates are `identical`) rather than assuming it, so a failure tells you which half broke. A second case shows the sum reaches every individual and not just the first.

While there: `introduce_new_node_and_update()` rebuilds the competition profile without recomputing rates, so `patch$ode_rates` in the neighbouring test was still what `reset()` left on an *empty* patch — `all(is.finite(...))` passed without ever seeing a computed environment rate. It now refreshes through `set_ode_state()` first.

### Also

- `TF24_Environment::clear_environment()`'s comment still described the state restoration that moved to `clear_state()` — sitting above the function that deliberately no longer does it. Split to match; `initial_states` now points at `clear_state()`.
- The base class records why there are two virtuals, **including that `Patch` doesn't clear at all**, so the next reader doesn't go looking for a mirror in `patch.h` that isn't there.
- `resource_depletion` is a reserved member as in `Patch`, not a vector allocated on every derivatives evaluation.
- `#include <numeric>` for `std::accumulate`, which was arriving transitively.
- NEWS: new entry for the `reset()` change (it wasn't documented at all), and the `light_env` → `env` rename now also appears under **Breaking changes**, where `plant-update-interface` reads from.

### Verification

Full suite green under `NOT_CRAN=true` — 54 files, including `test-model-version.R`, which runs and does not react. `test-stochastic-patch.R` 116 assertions, `test-stochastic-patch-runner.R` 120. The seeded baselines needed no re-derivation after either the merge or the guard, which is itself a datum: they are coarse enough that a 1e-5 perturbation doesn't move them, so they were never going to catch the thing I found.

I did **not** run the scenario gateway (`PLANT_RUN_SCENARIOS=1`). It isn't needed for the deterministic-path argument — `Environment::clear()` composes to exactly TF24's old `clear_environment()` body, and no other caller of `clear_environment()` exists — but it hasn't been run, so I'm not claiming it.

### Left undone

Nothing asserts the headline result: that the stochastic and SCM soil trajectories agree on the same drivers to ~3e-4. That's the property this PR exists to establish, and it's the strongest guard available for it. I've noted it as a known gap in the description rather than inventing a cheap proxy. Worth its own opt-in check.

---
_[Claude Code](https://claude.com/claude-code)_

dfalster:
CI is red on this branch, and it isn't this branch. All three checks fail in `setup-r-dependencies`, before anything compiles:

```
! Could not solve package dependencies:
* deps::.: dependency conflict
```

Same failure on `develop` itself: `R-CMD-check` passed on #590 and has failed on every run since #591. plant pins `traitecoevo/odelia@806924e` while `phylloptim`'s own `DESCRIPTION` asks for `traitecoevo/odelia` unpinned, so pak is handed two different refs for one package and refuses. Filed as #598 with the detail — it needs a decision about the family pinning convention, not a patch to one `DESCRIPTION`, so I've left it out of this PR.

Which means CI cannot vouch for this branch until #598 is fixed. What I have instead is local: full suite green under `NOT_CRAN=true`, 54 files, including `test-model-version.R`. Plus the FF16/K93 bit-identity check against develop-plus-`R/stochastic.R`, which is the claim CI would never have tested anyway.

dfalster:
Rewrote the description to fit #595's rule — it becomes the squash commit verbatim, and what I had was ~90 lines of measurements and rejected alternatives, which is exactly what that rule moves down here. Subject is 58 chars including the ` (#593)`. The working detail it displaced:

**What the fix does.** Mirrors `patch.h` — the ODE size, the three accessors, and resource accumulation in `compute_rates()`, with the consumption vector sized by `environment.n_resources()` rather than the environment's ODE width, so TF24's four diagnostic flux slots aren't mistaken for resources. Adds the per-node and per-species `consumption_rate()` forwarders that needs, and exposes `node_ode_size()` as `Patch` already does, because the runner's schedule setters have to guard on the individuals' share and not the total or an empty TF24 patch refuses a schedule.

`clear_environment()` is now the competition profile alone; restoring state moved to `clear_state()`, which only `Environment::clear()` calls. Worth knowing that `Patch::compute_environment()` doesn't clear at all — it leaves the profile standing on an empty patch — so this distinction is the stochastic path's alone, and the base class records why it's two virtuals rather than one.

**Numbers.** Soil water tracks the SCM to ~3e-4: recharge from 0.214 to the drainage equilibrium 0.3106, then drawdown as leaf area grows. FF16/K93 bit-identity verified against develop plus this branch's `R/stochastic.R` alone — the only changes touching their RNG draws or field names — `identical()` TRUE on both `time` and `species`.

**Baselines.** `test-stochastic-patch-runner.R` moved to `patch_area = 1` (the default) and its values were re-derived; at 50 m² the corrected schedule implies ~5300 individuals rather than the ~105 that test has always actually run. The cost: at area 50 the TF24 pair was 103/28 either side of the phylloptim move, and that exact match of two seeded integers was a sharper statement than any tolerance check. The equivalence still holds, it just isn't what this test pins now — noted in the test comment and in the `atm_kpa` NEWS entry that made the original claim.

**Verification.** Full suite green under `NOT_CRAN=true`, 54 files, `test-model-version.R` included and not reacting. `test-stochastic-patch.R` 116 assertions, `test-stochastic-patch-runner.R` 120. The scenario gateway stays behind `PLANT_RUN_SCENARIOS=1` and has **not** been run.

**Known gap.** Nothing asserts the headline result — that the stochastic and SCM soil trajectories agree on the same drivers. That's the property this PR exists to establish and the strongest guard available for it. Deserves its own opt-in check rather than a cheap proxy.

Earlier comments on this PR cover the review findings, the `reset()` step-size mechanism, and the uptake-arithmetic test in more detail.

dfalster:
Diagnosis of the Windows failure. **It is not a Windows quirk — it is a real bug this PR exposes, and macOS is passing by luck.**

All three failures are one error:

```
find_root_psi(find_root_crit=1) failed: ... Parameters a and b do not bracket
the root: a=2.0240868498177655e-07; min=0.000000; max=7.085493
```

`a` is `wettest_soil_layer`, i.e. ψ_soil = 2.02e-07 MPa. Inverting TF24's retention curve at the defaults (`a_psi` 1.78e3, `n_psi` 6.57, `θ_sat` 0.428):

```
psi = a_psi * (theta/theta_sat)^-n_psi / 1e6  =>  theta = 1.705
```

**θ = 1.705 m³/m³ — 3.98× saturation, and above 1.0, so more water than there is soil volume.** The collar bracket is `[ψ_soil, ψ_crit]`, and at a ψ_soil that far into nonsense it stops bracketing the root. The solver is reporting the soil state correctly; the soil state is wrong.

## Why this PR surfaces it

`clear_environment()` used to restore `vars.states = initial_states`, and it runs on every derivatives evaluation while the patch is empty. That reset the soil continuously, so an over-saturated layer could never persist. Splitting `clear_state()` out is right — the soil states *are* part of the ODE system and resetting them mid-integration was destroying integrated state — but it removes the accidental scrubbing that was hiding this.

## Why nothing stops θ going up

The dry end has three guards; the wet end has none on the state:

| | dry | wet |
|---|---|---|
| `psi_from_soil_moist` | `max(theta, theta_r)` floor, plus `min(psi, soil_psi_max_)` cap | **nothing** |
| rate in `compute_rates` | positivity guard, NaN-safe (`!(rate > 0)`) | **nothing** |
| `soil_K_from_soil_theta` | `max(theta, 0)` | `min(theta, theta_sat)` ✅ |

`soil_K_from_soil_theta` already clamps the wet end, and **its comment explicitly anticipates this exact state** — "once the soil feedback is perturbed, a theta far above saturation". So the hazard was recognised for the conductivity curve (#485/#549) and never applied to the retention curve.

That clamp also removes the drainage feedback that would otherwise correct the excess: above θ_sat, outflow is pinned at `K_sat`, so a layer whose upstream neighbour is also over-saturated has inflow == outflow and sits there at rate 0. With no individuals, `resource_depletion` is 0 too, so nothing removes the water.

## Suggested fix

Mirror the existing dry-end floor in `psi_from_soil_moist`, exactly as `soil_K_from_soil_theta` already does:

```cpp
const double t = std::min(std::max(soil_moist_, soil_moist_residual),
                          soil_moist_sat_layer);
```

θ > θ_sat is outside the retention curve's domain in precisely the way θ < 0 is outside the conductivity curve's. This floors ψ_soil at `a_psi/1e6` = 1.78e-3 MPa, which keeps the collar bracket well-formed.

That treats the symptom, and is worth doing on its own. The cause — the soil state reaching 4× saturation at all — deserves its own look, since a wet-end guard on the rate would stop it arising rather than making it harmless. Worth checking whether the over-saturation is a transient RK-stage overshoot that now persists, or genuine accumulation.

⚠️ **Do not read macOS green as "works on macOS".** The state is 4× saturation on both platforms; the platforms differ only in whether the trajectory happens to land somewhere the bracket still holds. This is latent everywhere.

dfalster:
Follow-up — I worked out **why** θ reaches 4× saturation, and it changes the fix I suggested above. **Take the ψ clamp as defence-in-depth, not as the fix.**

The soil ODE is violently stiff. `soil_K_from_soil_theta` has exponent `2*n_psi + 3` = **16.14**, so K spans 13 orders of magnitude across the θ range, and `K_sat/dz` is large:

| θ | K | K/dz (per year) |
|---|---|---|
| 0.05 | 1.45e-13 | 4.8e-13 |
| 0.20 | 7.58e-04 | 2.5e-03 |
| 0.30 | 0.527 | 1.76 |
| 0.40 | 54.7 | 182 |
| 0.428 (θ_sat) | 163.0 | **543** |

A **dry** layer sitting under a **saturated** one has inflow `K_sat` and outflow ≈ 0, so

```
d(theta)/dt = (163.0411 - 1.45e-13) / 0.3 = 543.5 / year
```

A single step of 0.001 yr (~9 hours) moves θ by **0.543** — larger than θ_sat itself. The observed 1.705 is reached in ~0.003 yr. So this is a **one-or-two-step overshoot of a stiff transient**, not water slowly accumulating: an empty patch has `resource_depletion = 0` and, left alone, actually drains.

That matters because clamping only ψ would leave the soil state at θ = 1.7 while making the symptom invisible — the water balance would be silently wrong, and `the collected trajectory reports a moving soil state` would be asserting on nonsense.

## What I think the fix is

The exact mirror of the dry-end guard already sitting five lines above, in `compute_rates`:

```cpp
// existing, dry end
if (theta <= soil_moist_residual && !(rate > 0.0)) { rate = 0.0; }

// proposed, wet end
if (theta >= soil_moist_sat_layer && !(rate < 0.0)) { rate = 0.0; }
```

Physically that is saturation-excess runoff, which the model already believes in — it is what `a_infil`/`b_infil` do for layer 0. This just extends it to layers that are filled from above rather than from rainfall. `!(rate < 0.0)` keeps the NaN-safety of the #549 version.

That stops θ leaving `[theta_r, theta_sat]` at all, which makes the retention curve's domain problem moot and keeps the water balance meaningful.

## Why it is worth doing here rather than deferring

`clear_environment()` was resetting the soil on every derivatives evaluation of an empty patch, so this transient was being continuously scrubbed. Your split is correct and I would not undo it — but it is what makes the stiffness visible, so the guard belongs with it.

Happy to push the guard and the ψ clamp to this branch and confirm CI, if you want it — it is your PR, and extending saturation-excess to depth is a modelling call rather than a mechanical fix, so I have not touched it.

dfalster:
**Retraction: the fix I proposed above does not work. Do not apply it.** I built it and measured it, and it makes things worse.

Same machine (macOS), same tests, `filter="stochastic"`, `NOT_CRAN=true`:

| build | result |
|---|---|
| this PR head, unmodified | **408 pass, 0 fail** |
| + saturation rate guard + ψ domain clamp | **363 pass, 4 fail** |

It converts a Windows-only failure into a failure on macOS as well — the three tests that were red on CI, plus the regression test I added. Nothing has been pushed to this branch.

## What is still true

The soil state is genuinely broken. ψ_soil = 2.02e-07 MPa inverts to **θ = 1.705**, which is 3.98× saturation and more water than there is soil volume. That arithmetic is straightforward and I stand behind it. The stiffness figures are real too — exponent `2*n_psi+3` ≈ 16, `K_sat/dz` ≈ 543/yr, so a dry layer under a saturated one clears θ_sat in one 0.001-yr step.

## What I got wrong

Two things.

**A rate guard cannot enforce a state bound under an explicit stepper.** `if (theta >= theta_sat && !(rate < 0.0)) rate = 0` is evaluated at the *start* of a step, where θ is still legal and the rate is legitimately positive. The stepper then integrates straight past θ_sat anyway. It only bites *after* the bound is already breached, so it changes the trajectory without preventing the overshoot — which is the worst of both.

**Clamping ψ to the saturated value is not conservative.** After the clamp the failure reappears at `a=0.00178` — exactly `a_psi/1e6`, ψ at θ_sat. So a bracket that worked with the absurd 2e-7 fails with the physically sensible 0.00178. I do not have an account of why, and I am not going to guess at one: it means the collar solve has a sensitivity at the wet end that I have not characterised, and "clamp to the physical domain" is not obviously safe there.

Sweeping ψ_soil directly through `find_root_psi(crit=1)` at TF24's curve parameters, wet soil on its own is fine — `root_crit` ≈ 7.05 all the way down to ψ_soil = 1e-6. So the trigger is a *combination* of soil state and individual (plant size / conductance), not wetness alone. My "wet soil breaks the bracket" framing above was too simple.

## What I would do next

Instrument the throw to dump the full `psi_soil` vector, `leaf_specific_conductance_max` and `root_carbon_per_leaf_area` at the failing call, and reproduce that exact operating point in isolation. Until the bracket failure is characterised at a known state, any guard on the soil side is speculative — as this one turned out to be.

Apologies for the noise; the earlier recommendation was posted before I had measured it.

dfalster:
Merged the odelia pin (#597). That fixed the dependency-resolution break in #598 — `lint` and **macOS now pass**. Windows fails, and it is worth being precise about what it is, because it is not what I first assumed and my first attempt at a fix was wrong.

### What Windows is failing on

Three TF24 stochastic tests, all with:

```
find_root_psi(find_root_crit=1) failed: ... Parameters a and b do not bracket the root:
a=2.0240868498177655e-07; min=0.000000; max=7.085493
```

`a` is the wettest soil layer's ψ. My first guess was that this branch lets soil recharge toward saturation where ψ → 0 and the collar root-find's bracket collapses, so I added the saturation clamp to `psi_from_soil_moist` that `soil_K_from_soil_theta` already has. **It changed nothing** — same seeds, same values — so I reverted it. The reason is arithmetic I should have done first: `psi_from_soil_moist` cannot return 2e-07, because its floor at saturation is `a_psi/1e6 = 1.78e-03`, five orders of magnitude larger. The ψ being reported is not a soil state this model can produce.

### It is pre-existing, and this branch improves it

Seeds 1..40, TF24 `run_stochastic_collect` at `patch_area = 1` (~105 individuals per m²), same R and same schedules, only the C++ differing:

| model code | seeds that throw |
|---|---|
| `develop` | **17 / 40 (43%)** |
| `develop` + this branch | **5 / 40 (12.5%)** |

So integrating the environment makes the dense regime markedly *more* robust, and it is broken either way. Filed as #599 with the full diagnosis, including where I would look next (`roots_::begin_solve()` loops `i < max_soil_layer` over `psi_soil_[i]`; a `max_soil_layer` past the vector's end would give exactly this signature — a small arbitrary positive double, varying by platform and run).

Why it has never been seen: this file has always run TF24 at `patch_area = 50`, which under the `delta_t` defect seeded 50 m² like 1 m² — density ~2 m⁻², where it is robust. This PR fixes that defect, so `patch_area` now means what it says, and moving the baseline to 1 m² put the suite in the 50x denser regime for the first time. Near the boundary it is a platform lottery: seed 1 passes on macOS and throws on Windows.

### This needs your call, so I have stopped

Three options, and the choice is not mine to make:

1. **Re-sparsen the test config** to the density develop actually supports. Cheapest, gets CI green, and honest only if the limitation is written down — but with the schedule fixed, density is `max_patch_lifetime × birth_rate` regardless of area, so a sparse stand means changing the birth rate or the lifetime, which changes what the baseline means and needs the values re-derived again. I tried this and got it wrong once already: a config I believed was 89–123 individuals was actually producing 1, and `stochastic_schedule()` returns a 1-arrival schedule for a birth rate it accepts without complaint. I am not confident enough in that path to ship it unsupervised.
2. **Fix #599 first** and let this PR follow. Correct, but it is model work in the hydraulics and probably in phylloptim, plausibly with a `scientific_version` question attached.
3. **Land this with the dense config and Windows red**, on the grounds that 12.5% beats 43%. I would not.

My preference is 2, then 1 as a documented interim. The branch is at `7b0e5f5a`, exactly the state I verified: full suite green on macOS under `NOT_CRAN=true` (54 files), FF16/K93 bit-identical.

dfalster:
Windows should be green now. `e07b9f96`.

### What changed

Nothing in the library. No default moved, no model is touched, and the step size stays a caller's decision — `test-stochastic-patch-runner.R` sets `ode_step_size_max = 0.05` for its own TF24 runs, and only for TF24. FF16 and K93 keep plain `Control()`, so their seeded baselines are untouched.

The limitation itself is recorded under **Known issues** in NEWS with the numbers, because a caller who runs TF24 stochastically at the default still meets it. The suite being deterministic is not the same as the bug being fixed, and the entry says so. #599 carries the real fix.

### Why the test needed it at all

I went looking for the mechanism rather than guessing, and it is not what the failure message suggests. The soil balance is stiff — the conductivity exponent is `2*n_psi+3 ≈ 16`, so `K_sat/dz` is ~543/yr at the defaults — and at a 5 yr cap the explicit stepper *diverges* rather than losing accuracy. A temporary diagnostic on the ψ_soil handoff caught a failing run at:

```
theta = [-51.4, 52.0, 0.146, nan, nan]
```

Once the rates are non-finite the step is not even rejected, because `adjust_step_size` derives its error ratio from a NaN and `NaN > 1.1` is false. The collar root-find is then handed a soil potential no retention curve can produce, which is the error you see.

| | seeds that throw |
|---|---|
| develop | 17/40 |
| develop + this branch | 5/40 |
| ... at `ode_step_size_max = 0.05` | **0/60** |

Not monotone — 0.2 fails *more* often than 5 does — so 0.05 is measured, not derived from a stability limit.

### Two things worth knowing

**Cost: none.** Full suite 42s, identical to before. I had assumed smaller steps would hurt and measured instead: the controller was already taking small steps through the body of a run, so the cap only removes excursions. That also means there was no case for skipping TF24 stochastic in regular testing — it would have dropped the coverage for exactly what this PR fixes, to save nothing.

**The recovered saturation guards are not in here.** They are on [`fix/soil-saturation-guard`](https://github.com/traitecoevo/plant/tree/fix/soil-saturation-guard) with their evaluation: 4/40 rather than 5/40, and seed 1 moves from passing to failing. A rate guard cannot bound a diverging state — it fires on `theta >= theta_sat`, and an unstable step leaves the guarded range entirely between evaluations.

Issue #60, opened 2014-03-24
Problem compiling on latest OSX


dfalster:
Possibly due to update with clang (to v 5.2, compared to 5.0)?

Full error message below

```
➜  tree git:(master) make
In file included from strategy.cpp:1:
In file included from ./strategy.h:8:
In file included from ./control.h:9:
./ode_control.h:39:20: error: forward declaration of class cannot have a nested name specifier
RCPP_EXPOSED_CLASS(ode::OdeControl)
                   ^~~~~
/Library/Frameworks/R.framework/Versions/3.0/Resources/library/Rcpp/include/Rcpp/macros/module.h:70:9: note: expanded from macro 'RCPP_EXPOSED_CLASS'
  class CLASS;                    \
        ^
In file included from strategy.cpp:1:
In file included from ./strategy.h:8:
./control.h:99:20: error: forward declaration of class cannot have a nested name specifier
RCPP_EXPOSED_CLASS(model::Control)
                   ^~~~~~~
/Library/Frameworks/R.framework/Versions/3.0/Resources/library/Rcpp/include/Rcpp/macros/module.h:70:9: note: expanded from macro 'RCPP_EXPOSED_CLASS'
  class CLASS;                    \
        ^
In file included from strategy.cpp:1:
In file included from ./strategy.h:12:
In file included from ./integration.h:8:
./functor.h:101:20: error: forward declaration of class cannot have a nested name specifier
RCPP_EXPOSED_CLASS(util::RFunctionWrapper)
                   ^~~~~~
/Library/Frameworks/R.framework/Versions/3.0/Resources/library/Rcpp/include/Rcpp/macros/module.h:70:9: note: expanded from macro 'RCPP_EXPOSED_CLASS'
  class CLASS;                    \
        ^
In file included from strategy.cpp:1:
In file included from ./strategy.h:12:
./integration.h:68:20: error: forward declaration of class cannot have a nested name specifier
RCPP_EXPOSED_CLASS(integration::QAG)
                   ^~~~~~~~~~~~~
/Library/Frameworks/R.framework/Versions/3.0/Resources/library/Rcpp/include/Rcpp/macros/module.h:70:9: note: expanded from macro 'RCPP_EXPOSED_CLASS'
  class CLASS;                    \
        ^
In file included from strategy.cpp:1:
In file included from ./strategy.h:13:
./interpolator.h:72:20: error: forward declaration of class cannot have a nested name specifier
[...]

PR #597, opened 2026-08-05
Pin odelia to v0.2.1, matching phylloptim

dfalster:
plant's develop cannot resolve its dependencies: phylloptim declared a
bare `Remotes: traitecoevo/odelia` while plant pinned a sha, so pak saw
two sources for one package and refused to solve. They agreed only until
odelia's master next moved, which it did.

odelia now carries a v0.2.1 tag and phylloptim names it too, so the two
specs are identical and cannot drift apart. The phylloptim pin moves to
e60a26af to pick that up; `inst/include/` is byte-identical across the
bump, so no results change.

Co-authored-by: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -53,8 +53,8 @@ Suggests:
     kableExtra
 Remotes:
     richfitz/RcppR6,
-    traitecoevo/odelia@806924e,
-    traitecoevo/phylloptim@818ad43
+    traitecoevo/odelia@v0.2.1,
+    traitecoevo/phylloptim@e60a26af
 URL: https://github.com/traitecoevo/plant
 BugReports: https://github.com/traitecoevo/plant/issues
 Config/roxygen2/version: 8.0.0


dfalster:
Working detail, moved out of the PR description. Under squash merge the description becomes the commit message verbatim, and none of this belongs in `git log` — see the [family guideline](https://github.com/traitecoevo/plant-meta/blob/main/governance/commit-messages.md). Posting it here first so nothing exists nowhere.

CI on this branch: `lint`, `macOS-latest` and `windows-latest` all green, with `setup-r-dependencies` succeeding on both platforms — the step that fails on `develop`.

---

plant's `develop` has been red since #591 merged. All three runners fail at `setup-r-dependencies`, before anything compiles:

```
! Could not solve package dependencies:
* deps::.: dependency conflict
* any::sessioninfo: dependency conflict
* any::rcmdcheck: dependency conflict
```

Neither `sessioninfo` nor `rcmdcheck` is the culprit — pak names innocent packages and never the actual conflict.

## What broke

phylloptim's DESCRIPTION carried a **bare** `Remotes: traitecoevo/odelia` — odelia's master tip, whatever that is today. plant pinned `@806924e`. Two different sources for one package, so pak refuses to solve the graph at all. #591 inherited this the moment it added phylloptim as a dependency.

**It passed on the PR and failed on merge with an identical tree.** `bfbac45e` and `b81786c6` have the same tree hash. The specs agreed only because `806924e` *happened* to be odelia's head; traitecoevo/odelia#50 landed at 01:16, between the green run and the red one, and they stopped agreeing. Nothing in plant changed — there is no point looking at the merge diff.

## The fix, on both sides

Matching a moving tip from here alone would only buy time until odelia's next push. So:

1. **odelia now has a `v0.2.1` tag** — its first release, at `806924e`, the commit declaring the version this `>= 0.2.1` floor was cut for and the one plant is validated against.
2. **phylloptim#60** points its remote at that tag (merged).
3. **This PR** names the same tag.

Two textually identical specs cannot drift apart, and a tag stays put where a branch tip does not.

## Why the phylloptim bump is inert

`818ad43 -> e60a26af` only picks up #60. `inst/include/` is **byte-identical** between the two, so the headers plant compiles are exactly the ones #591's 2846-test and 43-scenario runs were measured against. The rest is vignettes and tests.

## Verification

Reproduced and checked offline with pak 0.9.5 against the real remotes, rather than through CI:

| DESCRIPTION | `pak::lockfile_create("deps::.")` |
|---|---|
| current `develop` | **fails to solve** |
| this PR | **solves**, one `odelia 0.2.1`, `phylloptim 0.1.0` |

Refs #591, traitecoevo/phylloptim#60, traitecoevo/odelia#50

PR #600, opened 2026-08-05
Trim the slowest tests and lighten CI

dfalster:
The test suite was ~65% of every R CMD check job, and CI started only two
test processes -- testthat parallelises by file, so the slowest file floored
the phase.

test-tf24-arid-corner.R ran one dry-start TF24 run four times; it is now
memoised. #571's nine-point moisture sweep keeps the three values bracketing
the residual-moisture floor. Two tests go: one that could only skip once #574
fixed its reproducer, one re-asserting a bound the test above it checks.
Serial suite 147 -> 116 s, 2758 -> 2751 passing.

Benchmarks move to weekly, both workflows cancel superseded runs,
ubuntu-latest joins the matrix, TESTTHAT_CPUS is 4.

Co-authored-by: Claude Fable 5 <noreply@anthropic.com>


dfalster:
## CI results — all three green, including plant's first Linux run

Compared against develop at the same base (`d16d1357`, run `30968688529`) so this is apples-to-apples.

| | macOS | Windows | Linux (new) |
|---|---|---|---|
| **test phase** | 239s → **159s** (−33%) | 376s → **213s** (−43%) | **129s** |
| test processes | 2 → **4** | 2 → **4** | **4** |
| install/compile | 184s → 160s | 206s → 204s | 190s |
| deps | 78s | 84s | **205s** ⚠️ |
| **job total** | 10.5m → **8.4m** (−20%) | 14.2m → **10.8m** (−24%) | **9.9m** |

`FAIL 0` on all three. `PASS 2822` (macOS/Linux), `2817` (Windows).

### `TESTTHAT_CPUS: 4` worked

All three now report `Starting 4 test processes` instead of 2. Combined with the file trims, that's what produced the 33–43% cut in the test phase — a bigger effect than the trims alone, and it applies to every file rather than just the ones I touched. Keeping it.

### Correction to my own reasoning about Linux

I predicted Linux would be the fastest runner at ~5–7 min. **It isn't — 9.9m, slower than macOS's 8.4m.** But not for the reason that matters:

- Linux has the **fastest test phase of the three** (129s).
- It loses on **dependency install: 205s against 78s/84s**, because `odelia`, `phylloptim` and `RcppR6` are GitHub remotes with no Linux binaries, so they build from source.
- **This was a cold cache** — the first Linux run this repo has ever had. `setup-r-dependencies` caches, so subsequent runs should drop that 205s substantially. Worth re-reading after a few runs before drawing conclusions.

So the "make PRs Linux-only for speed" follow-up is **not yet justified on latency grounds** for a public repo where minutes are free. It remains the right move for any *private* repo in the family (1× vs 10× billing). I'd rather re-measure once the Linux dep cache is warm than gate the matrix on a cold-cache number.

### The bottleneck has moved

With tests down to 129–213s, **compile/install is now the largest single step everywhere** (160–204s). That makes **ccache the highest-value remaining lever**, not further test trimming — 25 template-heavy TUs recompiled from scratch every run. I deliberately left it out of this PR so it wouldn't confound the Linux experiment; it's the obvious next piece.

Also still open, unchanged: `test-strategy-tf24.R` is now the slowest file (23.4s local, ~19%) but **#583 modifies it**, so it should be trimmed there or after that merges.

dfalster:
## Re-measured with #583 and #593 merged in

Both are now in this branch (no conflicts — neither touches the two files I edited). Re-ran everything on the merged tree with `NOT_CRAN=true`.

### My contribution, isolated on the current base

My changes are R-only, so the compiled library is identical between this branch and develop — which means I could run develop's versions of the two files against the *same* build for a clean A/B:

| File | develop | this PR |
|---|---|---|
| `test-tf24-arid-corner.R` | 36.0s (PASS 21, SKIP 1) | **12.9s** (PASS 15, SKIP 0) — **−64%** |
| `test-tf24-water-budget.R` | 11.7s (PASS 26) | **8.9s** (PASS 25) — **−24%** |

Full suite on this branch: **115.7s, 322 tests, PASS 2868, FAIL 0, SKIP 1.** Since only those two files differ, develop's equivalent is 141.6s, so the saving is **−18%** — a bit better than the −16% I reported before, because the two files' shares grew slightly.

### Two things #583/#593 changed that are worth knowing

**#593 made the stochastic tests *cheaper*, not dearer.** I flagged in the PR description that its ~344 new test lines were a growth risk and suggested reviewing them for shared fixtures. That was wrong in the event: `test-stochastic-patch-runner.R` went **12.4s → 3.2s** despite gaining 165 lines. Retracting the concern — no action needed there, and my "cheapest moment is at review" note was solving a problem that didn't materialise.

**#583 made `test-strategy-tf24.R` the dominant hot spot.** It's now **25.2s, 21.8% of the suite** — a bigger share than `test-tf24-arid-corner.R` ever had after this PR. As predicted it added a `run_scm` to that file, and the seasonal-drought test at `max_patch_lifetime = 30` is still ~15s of it. **That is now the clear next target**, and it's unblocked since #583 has landed.

### One correction pushed to the branch

The matrix comment claimed ubuntu-latest was "the fastest and cheapest runner". CI disproved the first half — 9.9m against macOS 8.4m — so the comment now records what was actually measured, and that the Linux-only-PRs follow-up is on hold for a public repo where minutes are free. It's still the right move for any private repo in the family. Details in the previous comment.

dfalster:
## Correction: Linux *is* the fastest — once its cache is warm

I said in the last comment that Linux wasn't the fastest runner, at 9.9 min. That was measured on the first Linux run this repo has ever had, so the dependency cache was cold. The second run tells a different story:

| | cold (first run) | warm (second run) |
|---|---|---|
| Linux — deps step | **205s** | **37s** |
| Linux — whole job | **9.9 min** | **5.5 min** |
| macOS — deps step | 78s | 53s |
| macOS — whole job | 8.4 min | **5.7 min** |

So warm Linux is now the **fastest leg we have**, edging macOS. The 205s was `odelia`/`phylloptim`/`RcppR6` building from source — they're GitHub remotes with no Linux binaries — and `setup-r-dependencies` caches that away on subsequent runs.

Both figures are now in the workflow comment, because the cold one is what anyone re-measuring on a fresh branch or after a dependency bump will hit first, and it points the opposite way.

**This re-opens the follow-up I'd put on hold**: making Linux the only per-PR job is now justified on latency as well as cost. Still a separate PR — Windows should be gated on `src/**` and `inst/include/**` rather than dropped.

### Also visible in these two runs: #593 made the check step much cheaper

macOS `check-r-package` went **367s → 242s** between the pre-merge and post-merge runs of this same branch. That's #593's stochastic rewrite showing up at CI scale, consistent with the local `test-stochastic-patch-runner.R` drop of 12.4s → 3.2s. Worth noting since I'd flagged its new tests as a growth risk — they were the opposite.

Windows is still running; I'll confirm when it lands.

dfalster:
## Original PR description, relocated

Moved here verbatim so nothing is lost: the body above became the commit message
on squash, and the convention
([plant-meta/governance/commit-messages.md](https://github.com/traitecoevo/plant-meta/blob/main/governance/commit-messages.md))
caps it at ~10 lines. This is the material that belongs in a comment.

---

## Why

The suite ran **147 s** serially and was **~65% of each `R CMD check` job**, with one file at 24% of it. testthat parallelises *by file*, so the slowest file floors the test phase — and CI starts only **2** test processes, which makes that floor close to binding.

Measured baseline (run `30968688529`, green `develop` at `d16d1357`):

| Phase | macOS | Windows |
|---|---|---|
| setup-r + deps | 81 s | 161 s |
| compile/install | 124 s | 172 s |
| examples + static checks | 22 s | 58 s |
| **tests** | **292 s (67%)** | **401 s (64%)** |

Examples are *not* a problem (9 topics, ~0.01 s total). Queue wait is negligible (3–5 s).

## Result

**147.4 s → 124.1 s serially (−16%)**, `FAIL 0`, and no file is now over 19%.

| File | before | after |
|---|---|---|
| `test-tf24-arid-corner.R` | 35.5 s (24.1%) | **12.7 s** (10.3%) |
| `test-tf24-water-budget.R` | 11.7 s | **8.7 s** |

## What changed in the tests

`test-tf24-arid-corner.R` computed the **same dry-start TF24 configuration four times** at ~2.9 s a run — the height-ordering test, the completion test, the `theta = 0.005` sweep iteration (which duplicates the completion test's own run, `0.005` being `tf24_dry_start()`'s default), and the diagnostic-message test. The completed run is now memoised and shared, reusing the `short_run()` pattern already in `test-density-coordinate.R`.

**#571's nine-point soil-moisture sweep is cut to three** — the values bracketing the residual-moisture floor. The wide sweep recorded a non-monotone failure set that ruled out any single threshold as the cause; with the cause fixed in #574 it was spending six extra runs re-asserting what the boundary already covers. The full set is recorded in a comment and in #571 if it ever needs widening.

Two tests are **removed** rather than trimmed:

- **`"a failed light spline reports the patch state that caused it"`** could only ever skip. #574 made the reproducer complete, so its `skip_if(is.na(msg), ...)` fired on every run: it paid for a full SCM run and then asserted nothing. The file records how to bring it back as an *unconditional* test — drive the message builder from a synthetic patch state rather than provoking a real spline failure.
- **`"TF24 water budget closes independently of layer count"`** re-ran the same three layer counts as the test above it, to re-assert a bound that test already checks inside its own loop.

### Coverage accounting

Every lost expectation is one of the above, and no other file moved:

```
tests 313 -> 311   (the two removed test_that blocks)
PASS  2758 -> 2751 (6 dropped sweep iterations + 1 duplicate water-budget assertion)
SKIP  8 -> 7       (the always-skipping test)
FAIL  0
```

## What changed in CI

- **Benchmarks off the push/PR path.** `run_plant_benchmarks()` prints a `bench::mark` table and returns — nothing compares it against a baseline, uploads it, or fails on a regression, so it bought no signal for 3–5 min per trigger on a macOS runner. Now `schedule` (weekly) + `workflow_dispatch`, on `ubuntu-latest`. If it should become a real gate, it needs an artifact upload and a comparison against the previous develop run first.
- **`concurrency` + `cancel-in-progress` on both workflows.** Neither had one, so every intermediate push to a PR branch ran a full matrix to completion alongside its successor.
- **`ubuntu-latest` added to the `R-CMD-check` matrix.** It had been commented out since CI was first set up (#295), so plant has *never* had a green Linux run — despite no `SystemRequirements`/`OS_type` in `DESCRIPTION`, all deps being CRAN plus two public remotes, and `src/Makevars.win`'s own comment treating Linux as supported. **This PR is the experiment**: it runs alongside Windows and macOS so a Linux failure can't hide, and `fail-fast: false` means it can't block them.
- **`TESTTHAT_CPUS: 4`** — both runners reported "Starting 2 test processes" though standard runners have 4 cores. ⚠️ Check the log line to confirm this actually takes effect; if the runners expose only 2 usable cores it's a no-op and should come back out.

## Follow-ups (deliberately not here)

- **Gate the expensive runners**: Linux on every PR, Windows only when `src/**` or `inst/include/**` changes (plant's most platform-fragile surface — `src/Makevars.win` links `odelia.dll` at load time because mingw `ld` forbids undefined symbols in a DLL, traitecoevo/odelia#29), macOS only on pushes to develop/master. Waiting on evidence from this PR that Linux is reliable.
- **`test-strategy-tf24.R`** is now the slowest file (23.3 s, 18.9%), almost all in the seasonal-drought test at `max_patch_lifetime = 30`. Left alone because **#583 modifies this file** — worth doing in that PR or after it merges, not in a conflicting branch.
- **ccache** for the C++ compile (124 s macOS / 172 s Windows every run, 25 template-heavy TUs). Left out so an untested three-platform toolchain change doesn't confound the Linux experiment above.
- **#593** adds ~344 lines to the two stochastic test files. Cheapest moment to stop the suite regrowing is at review — worth checking whether its several new seeded runs can share one fixture, which is the same anti-pattern fixed here.

## Note on billing

Worth recording since it prompted this work: **plant is a public repo, so its Actions minutes are free and unmetered.** August's allowance was consumed by `phylloptim` (then `leaf_cpp`, then private) at 2041 billable minutes against a 2000-minute org allowance — macOS bills at 10×. Making it public resolved that. So this PR is a **latency** win, not a billing one.

dfalster:
## Retraction: my CI numbers above were over-claimed

@dfalster asked whether we'd really trimmed by that much. Checking properly: **less than I said, and macOS proves nothing at all.**

I'd been quoting single-run comparisons. With three paired runs per platform — develop vs this branch at the *same* base — the picture is:

**check-step seconds, paired by base**

| base | develop | this branch | Δ macOS | Δ Windows |
|---|---|---|---|---|
| pre-#593 | macOS 465 / win 674 | macOS 367 / win 509 | −98 | −165 |
| post-#593 | macOS 314 / win 668 | macOS 322 / win 517 | **+8** | −151 |
| post-#583 | macOS 393 / win 674 | macOS 305 / win 557 | −88 | −117 |

**Noise floor, from repeated runs within each arm:**

- **Windows on develop: 668, 674, 674** — a 6 s spread on ~670 s. Tight enough to trust.
- **macOS on develop: 314, 465, 393** — a **151 s spread on unchanged code**, wider than any effect I was claiming.

### So what actually holds

- **Windows check step: −144 s, ≈ −21%**, consistent in sign and magnitude across all three pairs, against a baseline whose spread is 6 s. This one is real.
- **macOS: unproven.** The deltas are −98, **+8**, −88. One paired run had this branch *slower*. macOS variance swamps the signal.
- **Linux: cannot be ranked against macOS** from what we have — warm totals are ~7.4 vs ~7.5 min, inside the noise.

### Specifically retracted

- ~~"macOS test phase 239 s → 159 s (−33%)"~~ and ~~"job total 10.5 → 8.4 min (−20%)"~~ — single-run pairs on the noisiest platform.
- ~~"warm Linux 5.5 min, the fastest leg we have"~~ — that 330 s job was the low end of 330/443/452. Typical warm is ~7.4 min, indistinguishable from macOS.
- The workflow comment has been rewritten accordingly; it no longer ranks Linux against macOS, and records that Windows is the only leg tight enough to A/B on.

### What was always solid

The **local** measurements, because they were controlled — same machine, same build, back-to-back:

- Two files A/B'd against the **identical** compiled library (my changes are R-only, so I ran develop's versions of just those files): `test-tf24-arid-corner.R` **36.0 → 12.9 s**, `test-tf24-water-budget.R` **11.7 → 8.9 s**.
- Serial suite **141.6 → 115.7 s (−18%)**, `FAIL 0`, with every lost expectation accounted for.
- `Starting 2 → 4 test processes` — a discrete fact from the log, not a timing.

**Bottom line: the suite is ~18% faster locally and Windows' check step ~21% faster in CI. The larger per-platform figures I posted earlier don't survive the noise.**

Issue #530, opened 2026-06-25
[speed] [acclimation] TF24f: share prepare_collar_solve across the centred-difference leaf evals


dfalster:
### Background

The centred-difference fix in #526 (`926b4f1b`) removed the O(h) tracking bias in `TF24f_Strategy::solve_leaf()`, but at a measured ~29% speed cost: `solve_leaf` now does **4** `Leaf::evaluate_root_collar_psi` calls per rate evaluation instead of 3 (centre + two perturbations + restore), and that root-collar solve dominates the per-step cost.

Same-machine timing (default horizon, `collect=FALSE`, k=2, h=0.05):

| build | run time | vs TF24 |
|---|---|---|
| TF24 (golden-section) | 208 s | — |
| TF24f forward (pre-#526) | 124 s | ~40% faster |
| TF24f centred (current) | 159 s | ~24% faster |

The ~29% slowdown tracks the 4/3 evaluation ratio almost exactly, confirming the leaf solve is the hot path.

### The opportunity

Each of the four `evaluate_root_collar_psi` calls independently re-runs `Leaf::prepare_collar_solve`, which rebuilds the same soil-side caches and feasibility setup — but within a single `solve_leaf` the plant/environment state is fixed, so that work is identical across all four calls.

Refactor `solve_leaf` (and/or `evaluate_root_collar_psi`) to:

- call `prepare_collar_solve` **once** to get the feasible interval `[bound_a, bound_b]` and the cached soil-side state, then
- evaluate profit at the three collar potentials (`used`, `used ± h`) directly, reusing those caches and clamping in-line.

This should bring the centred-difference cost back close to the old forward-difference cost (~124 s) while keeping the O(h²) accuracy.

### Notes / scope

- Behaviour must be unchanged: the boundary-clamp fallback to a one-sided difference near `[bound_a, bound_b]` edges must be preserved.
- `evaluate_root_collar_psi` is also called from the birth-init path (`initializing_`) and possibly elsewhere — keep the existing single-eval entry point working, or add a batch/"evaluate at these collar potentials with shared setup" helper rather than changing its semantics.
- Verify against #526's checks: steady-state gap to TF24 stays sub-mV and the tf24f test suite passes.

Follow-up to #525 / #526.

PR #605, opened 2026-08-05
Mirror the FD branch in TF24f's AD gradient

dfalster:
TF24f's AD branch went through evaluate_root_collar_psi, which hides
prepare_collar_solve's return value: it re-derived the soil-side caches on every
leaf evaluation, and asked for a gradient even where feasibility handling had
already forced the operating point. The FD branch has done neither since #530.

Bit-identical offspring_production at all 26 points of a 13-rainfall x two-branch
sweep, and 20% faster (2.61 -> 2.10 s at 1 m/yr); FD is the control and does not
move. #576's throw no longer reproduces on develop, so the test here is a guard.

Refs #576.


dfalster:
## Why the throw was correct

The failure was `Extrapolation disabled and evaluation point outside of interpolated domain`, from the **lower** end of phylloptim's `psi_from_transpiration` — not past the vulnerability curve, so "widen the domain" was aimed at the wrong end. The spline was right to refuse: it was being asked for the stem potential carrying a *negative* sap flux from a collar at `root_psi_crit`, i.e. wetter than saturation.

Worth recording because it is not visible from the phylloptim side: the throw is the **first statement** of `dprofit_at_collar_psi` (`find_psi_stem_from_psi_root`), so it happens before both of that function's own guards. Neither the `!isfinite(psi_stem)` check nor the `feasible` out-parameter added in phylloptim can see it — the header comment right beneath already documents that trap for the neighbouring failure ("the isfinite check ... cannot see a thrown exception"). Not asking is the only fix available at that layer.

## Measured: no movement where runs already completed

Double build of this branch against `develop` at 74b10ed7, phylloptim e265c6b, odelia 4b9668e. Same 13 rainfall values × both gradient branches, θ = 0.005, 5 layers, lma 0.0825, 10 yr, comparing `offspring_production` at 15 significant digits:

- **5 FIXED** — TF24f AD at rainfall 0.01, 0.03, 0.04, 0.05, 0.06
- **21 IDENTICAL** — every other AD point and all 13 FD points, bit-identical
- **0 MOVED**

So "behaviour changes only in states that previously threw" is measured, not asserted, and the TF24f scientific version does not move.

| rain | AD before | AD after |
|---|---|---|
| 0 | 0 | 0 |
| 0.01 | throw | 1.48140206500536e-22 |
| 0.02 | 2.34900129825802e-21 | identical |
| 0.03 | throw | 1.60813307143896e-20 |
| 0.04 | throw | 3.83039030613622e-20 |
| 0.05 | throw | 5.88132047989565e-20 |
| 0.06 | throw | 2.44057792209317e-19 |
| 0.07 → 2.0 | ran | identical |

The band was **not** monotone in dryness — rainfall 0 and 0.02 always ran, and 0.01 failed between them — which is what ruled out any single threshold as the cause. The R0 values in the failing band are 1e-22 to 1e-19, i.e. numerically extinct; the point of the fix is that the model *completes and reports* that rather than aborting the patch.

## Suite

`test_check("plant")` with `TESTTHAT_CPUS=4`: **2916 pass, 0 fail, 0 warn, 7 skip** (all `On CRAN` guards).

## Test cost, given #600

#600 had just trimmed this file (147 s → 116 s serial) by memoising the shared dry-start SCM and cutting #571's nine-point moisture sweep to three. So the new test adds **two** SCM runs, not a sweep: one per gradient branch at a single rainfall value. The assertion that matters is that the two branches *agree*, since disagreement was the bug — sweeping rainfall buys breadth in the wrong dimension at a full run apiece. The measured band is in #576 if it ever needs widening.

`tf24_dry_start()` gains a `rain` argument rather than a second near-identical helper being added beside it.

## Relationship to #582

This supersedes [#582](https://github.com/traitecoevo/plant/pull/582), which is closed unmerged. #582 predated [#591](https://github.com/traitecoevo/plant/pull/591) (leaf model out to phylloptim) and [#600](https://github.com/traitecoevo/plant/pull/600), and merging `develop` into it conflicts in five of its six files — including `src/leaf_model.cpp` as a modify/delete that would resurrect the file #591 deleted. The diagnosis carried over; the diff did not.

#582 also bundled a second, separable change: naming the spline, the point and the domain in the out-of-domain error, so that localising this kind of failure is a read rather than a bisect. That half does **not** belong here — the message now comes from odelia's `interpolator.hpp`, which has `u`, `min()` and `max()` in hand and reports none of them. It is being done bottom-up instead: odelia reports the point and domain (fixing every spline in the family, not just the two leaf ones), then phylloptim adds the spline and caller identity odelia cannot know. `tests/testthat/test-leaf.r:206` and `:214` still assert the bare odelia string, so that work lands behind a version-floor bump; it is deliberately not in this PR, which is the bug fix alone.

## Not fixed here

[#584](https://github.com/traitecoevo/plant/issues/584) — `prepare_collar_solve`'s dead `bound_b` clamp — is still open and still latent. Fixing it moves results for runs that currently complete, so it does not belong in a change whose claim is that nothing moved.

aornugent:
This also fixes a second failing state, which the description does not mention: a shaded plant on wet soil. Worth recording, because it fails in a different place.

```r
test_that("TF24f takes a zero gradient in deep shade on wet soil (#576)", {
  env <- Environment("TF24f")
  env$set_soil_number_of_depths(5)
  env$set_soil_water_state(rep(0.214, 5))
  env$set_fixed_environment(0.02, 150)

  for (ad in c(TRUE, FALSE)) {
    s <- TF24f_Strategy(); s$use_ad_gradient <- ad
    i <- TF24f_Individual(s)
    i$set_state("height", 5)
    i$set_initial_states(env)
    expect_no_error(i$compute_rates(env))
  }
})
```

On 74b10ed7 this gives
```
#> psi_stem_to_ci failed: ... Parameters a and b do not bracket the root:
#>   psi_stem=0.170872; psi_upstream=0.170867
```

On this branch: no error, the rate is 0, transpiration is 0. Same at openness 0.029, 0.02 and 0.001.

It is a different route from the dry one where the transport inverse fails. Here water is not the constraint: the soil is wet, the transport path works, and `psi_stem_to_ci` fails instead because `gc` is 0 and the residual has the same sign at both ends of [Gamma*, ca]. Ecologically the plant is suppressed — at 2 percent openness gross assimilation cannot cover dark respiration, so no stomatal opening gains carbon. It shuts its stomata at the wettest end of its range (collar and stem both 0.1709 MPa, net assimilation −R_d), compared to the dry plant which shuts down at the other end (both at psi_crit, 7.0855 MPa). That makes this the shade-mortality state, and the ordinary condition of a suppressed understorey. Your guard catches both because it sits at `prepare_collar_solve` and bails on every exit; narrowing it to the shutdown exits would reopen this one.

dfalster:
## Rebased onto develop, and the claim has changed

Rebased from `74b10ed7` onto `b3472d43`. The diff needed one real update: phylloptim 0.6.0 made the collar API take the cost curve as a template parameter, so the three AD-branch calls now read `prepare_collar_solve<Leaf::CostCurve::TF24>` and `profit_at_collar_psi<Leaf::CostCurve::TF24>`, matching what #606/#633 already did to the FD branch. Without that the branch does not compile.

More importantly, **the bug this PR was written to fix is gone**, so the reason to merge it is now different from the reason it was opened. Everything below is measured on this branch against its own parent, both built against odelia v0.3.1 and phylloptim `83fbb25` (the pinned SHA, exported with `git archive` — the sibling checkout is ahead of it) into private libraries, so neither build could pick up the other's.

### #576 no longer reproduces, on either side

The 13-rainfall × two-gradient-branch sweep from the [original comment](https://github.com/traitecoevo/plant/pull/605#issuecomment-5187887893), rerun on plain `develop`:

| | before (74b10ed7, Aug 5) | now (develop) |
|---|---|---|
| AD completing | 8 / 13 | **13 / 13** |
| FD completing | 13 / 13 | 13 / 13 |

All five previously-throwing AD points (rainfall 0.01, 0.03–0.06) complete on `develop` without this PR. So does @aornugent's second state — the [shaded plant on wet soil](https://github.com/traitecoevo/plant/pull/605#issuecomment-5188733478) — at all three openness values, on both branches. phylloptim 0.6.0 closed both upstream.

The R0 values also moved a long way in the meantime: AD at rainfall 0.01 was 1.48e-22 in August and is 5.64e-19 now. That is #633's territory (the `kg_to_mol_h2o` shift and the resistance-based interface), not this PR's.

### What this PR does now: nothing to the numbers, 21% off the clock

Same sweep, this branch vs its parent — **bit-identical `offspring_production` at 15 significant digits on all 26 points**, 0 moved. Timing, minimum of five runs each on a 10-year dry start:

| | parent | this branch |
|---|---|---|
| AD (default), rainfall 1 | 1.54 s | **1.22 s** |
| AD (default), rainfall 0.04 | 1.08 s | **0.87 s** |
| FD, rainfall 1 | 1.20 s | 1.19 s |

FD is the control: this diff does not touch it, and it does not move. AD lands at 1.22 s against FD's 1.19 s, which is the whole story — the exact gradient was never the expensive part, the two redundant `prepare_collar_solve` calls were, and #530 had already taken them out of FD.

### The guard is not dead code — it is 3–11% of all leaf solves

Worth recording, because "the bug is fixed upstream" invites deleting the guard along with it. A probe build counting both paths:

| run | AD solves | forced-exit | share |
|---|---|---|---|
| rainfall 1 m/yr | 642,550 | 21,806 | **3.4%** |
| rainfall 0.04 m/yr | 474,802 | 50,448 | **10.6%** |

A forced operating point is an ordinary state for a water-limited plant, not an arid-corner curiosity. Two consequences:

- It is where most of the 21% comes from, so the guard is the optimisation, not a side effect of it.
- **It should not be turned into a rejected ODE step.** With odelia 0.3.1's domain checks now in (#628), that is the tempting reading of "shutdown is unphysical" — but a stepper asked to shrink away from 3–11% of its own solves would shrink until it gave up. The infeasible-probe translation in #628 is for states that are genuinely unreachable; this one is reachable by design.

Why keep the guard at all, if `develop` already gives the same numbers without it? Because it gives them by accident: phylloptim returns a **0.0 sentinel** from `dprofit_at_collar_psi` on its own shutdown exits, and that is what the AD branch has been reading. The failure mode #576 actually hit was a *throw* from the transport inverse, which lands on the first statement of that function — before both of its own guards, so neither the `isfinite` check nor the `feasible` out-parameter can see it. Not asking is still the only protection available at this layer.

### Still on the table, deliberately not here

`dprofit_droot_collar_psi` seats the soil-side caches itself, so this branch goes from three prepares per solve to two, not to one. phylloptim's `dprofit_at_collar_psi<K>` is the post-prepare body and would close that. Left out because reseating in a different order can move the last bits, and bit-identity is the claim this PR rests on — it deserves its own measurement.

[#584](https://github.com/traitecoevo/plant/issues/584) — `prepare_collar_solve`'s dead `bound_b` clamp — is still open and still latent, for the same reason as before.

### Suite

`tests/testthat.R` with `TESTTHAT_CPUS=6`: **2915 pass, 0 fail, 0 warn, 7 skip** (all `On CRAN` guards). The regression test in `test-tf24-arid-corner.R` passes on the parent too, and its comment now says so in the first line — a guard, not a reproducer. Its cost is two SCM runs at ~0.9 s.

One caveat on the measurements: the two intervening develop commits, #607 and #608, landed while this was building. The sweep baseline is `b374ce3b`; #607 is NEWS plus test assertion strings and #608 is notes only, so neither can move a number. The suite result above is on the final rebased tree at `b3472d43`.

dfalster:
## Re-measured on `745dd600` (#619)

Rebased again, now onto #619. That one moves TF24 numerics (TF24@v9), so every number in [the comment above](https://github.com/traitecoevo/plant/pull/605#issuecomment-5423854886) was re-measured from scratch rather than carried over — fresh base and branch builds, both against odelia v0.3.1 and phylloptim `83fbb25` in separate private libraries.

The rebase itself was clean: #619 touches `tf24_strategy.cpp` and `patch.h`, this touches `tf24f_strategy.cpp`.

### The two conclusions hold

**Bit-identical `offspring_production` at all 26 sweep points**, 0 moved — as before. And #576 still does not reproduce on either side: 13/13 AD points complete on plain `develop`.

The absolute R0 values moved a long way under #619, which is why they had to be redone rather than reused — e.g. AD at rainfall 1 m/yr, 0.0154 → 0.000471.

### Speed: still ~20%, but AD no longer lands on FD's number

Minimum of five runs each, 10-year dry start:

| | parent | this branch | |
|---|---|---|---|
| AD (default), rainfall 1 | 2.61 s | **2.10 s** | −20% |
| AD (default), rainfall 0.04 | 1.74 s | **1.41 s** | −19% |
| FD, rainfall 1 | 2.20 s | 2.23 s | control |
| FD, rainfall 0.04 | 1.19 s | 1.21 s | control |

Correcting one thing I said last time: I described the fixed AD branch as reaching parity with FD, which was true on the old base and is not on this one. At rainfall 1 the fixed AD branch is now slightly *faster* than FD (2.10 vs 2.23 s); at 0.04 it is still 17% slower (1.41 vs 1.21 s). The ~20% saving is the robust part; parity was a coincidence of the old base.

Note the whole model got slower under #619 despite its own step count falling — AD at rainfall 1 went 1.54 s → 2.61 s on the parent. That is #619's to explain, not this PR's; flagging it only because the two rounds of numbers are otherwise not comparable.

### The forced exit is now half of all leaf solves at 0.04

The probe build, recounted:

| run | AD solves | forced-exit | share | was (pre-#619) |
|---|---|---|---|---|
| rainfall 1 m/yr | 738,481 | 28,686 | **3.9%** | 3.4% |
| rainfall 0.04 m/yr | 794,676 | 397,700 | **50.0%** | 10.6% |

At 0.04 m/yr, half of every leaf solve in the run now ends at a forced operating point. #619's own description reports 52.7% of a mature stand sitting on the old storage clip, so a reserve change putting plants into hydraulic shutdown far more often is at least consistent, though I have not tried to establish the mechanism.

Either way this sharpens the point from last time rather than changing it: **the forced exit must not be turned into a rejected ODE step**. With odelia 0.3.1's domain checks now in (#628) that is the tempting reading of "shutdown is unphysical", but a stepper asked to shrink away from half its own solves will shrink until it reports "Cannot achieve the desired accuracy". #628's `infeasible_error` translation is for genuinely unreachable states; this one is reachable by design, and now routinely.

One curiosity I have not chased: the forced-exit counts are *identical* between the AD and FD runs (397,700 and 28,686 in both), even though the two runs end at different R0. Recorded in case it means something to someone.

### Suite

`tests/testthat.R` with `TESTTHAT_CPUS=6`: **2940 pass, 0 fail, 0 warn, 7 skip** (all `On CRAN` guards). CI was green on the previous rebase (ubuntu + macOS passed, Windows was still running when this round started).

Issue #63, opened 2014-04-03
Give James access to tree git repo


dfalster:
username: jscamac

Issue #34, opened 2013-07-10
Disturbance regime within Parameters should be simplified


richfitz:
Perhaps don't keep the whole `Disturbance` object in the `Parameters` vector, but just the mean disturbance interval.  Then each `Patch` will create one, given the parameters.

Issue #33, opened 2013-07-10
Seed rain should be part of the `Parameters`


richfitz:
Set up should be similar to disturbance regime. This will simplify the setting of seed rain, and help give a more unified interface with the stochastic model.

The motivation here is simplifying the comparison with the reference model.  The counter motivation is just that we'll be solving for zero seed rain in running the EBT, which makes it seem better suited to being an argument.  But I don't think that's a good reason for having it separate.

PR #606, opened 2026-08-05
Follow phylloptim's resistance-based leaf interface

dfalster:
phylloptim#33 changed `Leaf::set_physiology` to take the per-layer root hydraulic
resistances rather than the root carbon they are derived from, because the supply
solve reads exactly `r_R_H_min` and `r_R_V_sum` and nothing else.
`TF24_Strategy` now owns the root-architecture model that produces them:
`beta_R_H` and `beta_R_V` stay here as TF24 parameters at the same values, and
`net_mass_production_dt` fills a `RootNetwork` member through
`phylloptim::root_network_from_carbon` before each `set_physiology`. Same split
`leaf_specific_conductance_max` already had — plant computes `kmax` from height
and passes a scalar.

Breaking for R callers: `Leaf()` drops `beta_R_H`/`beta_R_V` (17 arguments, not
19), and `set_physiology` takes `root_network`. Migration in `NEWS.md`.

**No TF24 results change** — bit-identical over 18 operating points × 26 states,
rates and auxiliaries against `develop`. `scientific_version` unchanged.

⚠️ **Do not merge alone.** Coupled with traitecoevo/phylloptim#63, and `Remotes:`
currently points at that PR's branch — re-pin to the squash-merge sha first.


dfalster:
## Verification

### Bit-identity, which the phylloptim golden file cannot establish

phylloptim's 288-point golden file constructs a fresh `Leaf` per point and never
goes through TF24. So the plant-side check is separate, and it is the one that
matters here: 6 heights × 3 soil-moisture profiles through
`TF24_Individual$compute_rates`, reading all 26 states, rates and auxiliaries,
compared with `identical()`.

```
dims identical: TRUE
colnames identical: TRUE
BIT-IDENTICAL: TRUE
non-finite pattern identical: TRUE
```

Reference arm: `develop` (74b10ed7) built against phylloptim master (0.1.0). Test
arm: this branch against phylloptim#63 (0.2.0). Both built and run in the same tree
and the same R library, sequentially.

The sweep is chosen for what it exercises rather than for coverage. Heights span
0.4–20 m, and rooting depth is `min(height, rooting_depth_max)` capped at 1.5 m, so
**which layers hold carbon changes across the sweep** — that is the case where
`max_soil_layer` truncates and the network is shorter than the soil profile, which
is exactly the new length check's territory. The moisture profiles are uniform wet,
uniform dry, and drying-with-depth, so `psi_soil` differs per layer in the third.

Probe script: `scratchpad/probe.R`, not committed.

### Test suites

| suite | result |
|---|---|
| full plant suite, `NOT_CRAN=true` | **0 failures, 2936 pass**, 1 skip (opt-in `PLANT_RUN_SCENARIOS`) |
| full plant suite, default | 0 failures, 2915 pass, 7 CRAN skips |
| `test-leaf.r` + `test-pm-leaf-demo.R`, `NOT_CRAN=true` | 0 failures, 232 pass, **0 skips** |

⚠️ Pass counts here are environment-dependent and should be compared against a
control run, not against these numbers.

### What changed in `test-leaf.r`

37 `set_physiology` call sites and 24 `Leaf()` constructor calls. Three
non-mechanical parts:

1. **`net()`, namespaced to phylloptim rather than reimplemented.** The vertical
   resistance scales with `dz²`, so a second copy of that arithmetic in plant would
   be one slip away from a silent squared factor that neither package could detect.
   phylloptim is a `LinkingTo` dependency already; it is now also in `Suggests` for
   this R-level use.
2. **`beta_R_H`/`beta_R_V` at 10× the package defaults** (3.4e3 / 9.4e4) were
   assigned separately in each of five `test_that()` blocks. They are now stated
   once, as `test_beta_R_H`/`test_beta_R_V`.
3. **The length-mismatch test needed splitting in two.** It passed a one-layer
   `psi_soil` with a two-layer `soil_depth` and expected `set_physiology` to reject
   it. Building the network for the two-layer profile now makes the network
   *consistent* with `soil_depth`, so the mismatch under test is psi_soil vs
   soil_depth as intended — and a second assertion covers the new case, a network
   with more rooted layers than the soil profile has layers. That one is not a wrong
   number but an out-of-bounds read in `uptake()`: `max_soil_layer` indexes
   `psi_soil_` and `grav_head_z_` directly, and the agreement used to come free from
   validating root carbon against `soil_depth`.

### `RootNetwork` is bound here but deliberately not exported

plant needs the *class* so RcppR6 generates the `Rcpp::as`/`wrap` that makes
`set_physiology` callable from R. It does not need to export the constructor:
phylloptim exports its own `RootNetwork()`, and two exported constructors of the
same name would mask each other under `library(phylloptim); library(plant)`. RcppR6's
`is<T>()` recognises the class *attribute*, not the package
(`x.inherits(class_name_r<T>())`), so a phylloptim-built network passes straight
through plant's binding — which is what `test-leaf.r` now relies on and demonstrates.

The whole block is temporary: phylloptim#34 deletes plant's `Leaf` bindings entirely
and collects the hazard-7 payoff. #33 had to come first precisely because it changes
bindings that #34 removes.

### Other call sites updated

- `overstorey_staging/pm_demo_helpers.R` — which is why `test-pm-leaf-demo.R` now
  runs rather than erroring under `NOT_CRAN=true`
- `scripts/validate_gamma_transform.R`, `notes/penman-monteith/step5_fick_vs_pm.R`
- `agents.md` §3.2, with the hazard below

`R/root_uptake.R` is left alone: it is a standalone R prototype of the uptake
equations with its own arguments, not a caller of the C++ model.

### One hazard is new, and nothing can catch it

The leaf is purely intensive, and the root resistances now arrive **already divided
by leaf area**. A network built from absolute carbon is five vectors of plausible
positive numbers — the leaf cannot tell, and `E_up` comes back wrong by the leaf
area with no error anywhere. Previously the division happened inside the leaf, so
the contract was enforced by construction rather than by convention.

In this repo the division is exact and free: `mass_root()` is strictly linear in
`area_leaf` (`pars.a_r1 * area_leaf`), so `root_mass_carbon_scale * mass_root_ /
area_leaf_` is exactly `root_mass_carbon_scale * pars.a_r1` and no division is
written. That is unchanged by this PR. But it now sits upstream of a boundary that
will not check it, so it is recorded in `agents.md` §3.2 alongside a second note:
**do not open-code the layer thickness** — `phylloptim::layer_thickness()` is the
shared definition, and `MultiLayerRoots::set_soil_state` uses the same one.

### Performance

+0.7% per leaf solve, measured on the phylloptim side (2.74 → 2.76 µs/solve,
min-of-2000, interleaved six times, non-overlapping). See
traitecoevo/phylloptim#63's first comment. Not separately measured through an SCM
run here: the leaf solve is a small fraction of a demographic step, and the change
is a fixed +0.02 µs per `set_physiology` rather than anything that scales.

### Before merging

`Remotes:` points at `traitecoevo/phylloptim@feature/take-resistances-not-carbon`
so CI can build against the coupled PR. **Re-pin it to phylloptim#63's squash-merge
sha before merging**, or the pin is a moving target — `LinkingTo: phylloptim
(>= 0.2.0)` is the version floor and the `Remotes` sha is the reproducible half;
both are needed.

dfalster:
## Re-pinned to the merged sha, and re-verified against it

`Remotes: traitecoevo/phylloptim@feature/take-resistances-not-carbon` →
`@037673b7`, which is phylloptim master's tip after both traitecoevo/phylloptim#63
and traitecoevo/phylloptim#66 merged. The branch that pin referred to no longer
exists.

**`LinkingTo: phylloptim (>= 0.2.0)` is unchanged and still correct.** phylloptim#66
changed `set_supply_single`'s signature and made the single-potential path's
resistance a per-call driver — plant binds neither and uses only the multi-layer
path, so its call is byte-identical across that change. The floor is the version
that introduced the API plant *does* need, which is 0.2.0.

### Re-verified, because the previous run was against the wrong headers

| check | result |
|---|---|
| TF24, 18 operating points × 26 states/rates/auxiliaries | **`identical()` TRUE** |
| full suite, `NOT_CRAN=true` | 0 failures, 2936 pass, 1 opt-in skip |
| translation units compiled | 3 of 3 (not a stale relink) |

Test arm: this branch + phylloptim@037673b7. Reference arm: `develop` +
phylloptim@e265c6b (the commit before #63). **Both arms built into their own
libraries**, with `R_LIBS` pointed at them explicitly, rather than into the shared
site library.

That isolation is not ceremony — it caught a real problem. A concurrent branch,
`phylloptim@diag/name-the-transport-spline` (version 0.2.1), had been installed into
the shared site library, so an earlier rebuild of this PR compiled against **that**
rather than against the sha this PR pins. The install reported three translation
units compiled and looked entirely healthy. Anything measured that way is measuring
an unrelated branch.

Two things worth carrying from it: a `LinkingTo` dependency in a shared library is
**mutable state** that a header-only consumer silently inherits, and the compile
count tells you the build was real but not what it was real *against*. When the arms
matter, give each one its own library.

PR #607, opened 2026-08-05
Assert what the hydraulic domain error diagnoses

dfalster:
`test-leaf.r` asserted only "evaluated outside its domain", which does not say which lookup failed -- and that is the fact worth pinning. The leaf reads the same transport spline from four places and holds a second that is its inverse, so what a caller needs from the message is which of those it was.

Both assertions now check the spline, which end was missed and the caller. The soil-side case names the **stem** argument, because `psi_stem` and `psi_soil` are equal there and the stem check comes first. NEWS records the message (#576) under Minor changes.

Test text only; no model behaviour moves. Rebased onto `develop`: the original DESCRIPTION hunk is obsolete, #633 having already taken plant to phylloptim 0.6.0 / odelia 0.3.1.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -639,6 +639,8 @@ were not previously recorded here:
 
 ### Minor changes & bug fixes
 
+* **An out-of-domain hydraulic lookup now names the spline, the point, the domain and the caller** (#576), via odelia >= 0.2.2 and phylloptim >= 0.2.1, both already required. The message used to be odelia's bare "Extrapolation disabled and evaluation point outside of interpolated domain.", which named none of the four — and since the leaf reads the same transport spline from four places and holds a second spline that is its inverse, that sentence did not distinguish the cases that matter. Localising #576 meant instrumenting the call sites by hand to find out which spline was read and at what value; the answer (the *lower* end of the inverse, where the demanded flux is negative) is what showed the spline had been right to refuse. Error text only; no model behaviour changes. `test-leaf.r` now asserts the spline, which end was missed and the caller, rather than the whole sentence — which is phylloptim's to word.
+
 * **The scenario gateway scores numerical viability and persistence as separate
   axes, and no longer leads with a match rate** (#572). `scenario_summary()`
   classified a run as a success on `finite && total > 0`; with `birth_rate = 1`
--- a/tests/testthat/test-leaf.r
+++ b/tests/testthat/test-leaf.r
@@ -235,16 +235,34 @@ expect_error(l$set_physiology(root_network = net(rep((root_carbon_) / area_leaf_
   
   upper_bound_int <- 3*((log(1/1e-5))^(1/2.04))
   #for situations where psi_stem exceeds tolerance of integrator
-  expect_error(l$transpiration(upper_bound_int, psi_stem[1]), "evaluated outside its domain")
-  
+  # The leaf reads the same transport spline from four places and holds a second
+  # spline that is its inverse, so "outside its domain" on its own does not say
+  # which lookup failed -- and that is the fact localising #576 turned on. Assert
+  # the three parts that make the message a diagnosis (which spline, which end,
+  # which caller) rather than the whole sentence, which is phylloptim's to word.
+  err <- expect_error(l$transpiration(upper_bound_int, psi_stem[1]),
+                      "evaluated outside its domain")
+  expect_match(conditionMessage(err), "transpiration_from_psi")
+  expect_match(conditionMessage(err), "beyond the upper end")
+  expect_match(conditionMessage(err), "Leaf::transpiration, at psi_stem")
+
   #for situations where psi_soil exceeds psi_crit + tolerance
   
   psi_soil = upper_bound_int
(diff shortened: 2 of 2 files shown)


dfalster:
## Blocked, and on which PR — measured, not guessed

**This is gated on [phylloptim#65](https://github.com/traitecoevo/phylloptim/pull/65), not on [odelia#53](https://github.com/traitecoevo/odelia/pull/53).** I expected the opposite and checked, which is worth recording: `l$transpiration()` reaches phylloptim's `stem_curve_integral`, whose new domain guard fires *before* odelia's `Interpolator::eval` is ever entered. So odelia's improved message never reaches these two assertions; phylloptim's does. odelia#53 still matters for plant's light spline and drivers, but nothing in `test-leaf.r` depends on it.

Finding that out needed a `git clean -xdf src` first. phylloptim is header-only and `LinkingTo`, so plant's `.o` files carry no dependency on its headers: `R CMD INSTALL` after updating phylloptim relinked stale objects and the suite passed, which looks exactly like "the change is compatible". It is not. Only the clean rebuild showed the failure:

```
── Error ('test-leaf.r:206:3'): Basic functions ──
Leaf hydraulics: transpiration_from_psi (the cumulative xylem conductivity integral G,
psi in +MPa) evaluated outside its domain: psi = 9.93826 lies 3.65946 beyond the upper
end of [0, 6.2788]; asked by Leaf::transpiration, at psi_stem.
```

## Why not just re-pin the whole sentence

The old assertion was `expect_error(..., "Extrapolation disabled and evaluation point outside of interpolated domain.")` — the entire message, which is also the *only* message any interpolator produced. It could not tell one spline from another, and it would have to be rewritten in full every time the wording changed. The new assertions check the three facts that make the message a diagnosis:

- which spline (`transpiration_from_psi` — there are two, and they are inverses of each other with different units),
- which end was missed (`beyond the upper end`),
- which caller was asking (`Leaf::transpiration` — the same spline is read from four places).

Line 214 gets the caller assertion rather than the end assertion, because it reaches the same failure from the soil side rather than the stem side, and the caller is what distinguishes them.

## Verification

Against phylloptim at `diag/name-the-transport-spline` and odelia 0.2.2, plant clean-rebuilt from source:

- `test_check("plant", filter="leaf")`: **216 pass, 0 fail, 2 skip** (was 214 + 2 skip; the two extra are the added assertions).
- `test_check("plant")` full, `TESTTHAT_CPUS=4`: **2916 pass, 0 fail, 0 warn, 7 skip** — all skips are `On CRAN` guards.

## Before merging

- [ ] phylloptim#65 merges and is tagged 0.1.1.
- [ ] Repin `Remotes: traitecoevo/phylloptim@diag/name-the-transport-spline` to the merge SHA (or `@v0.1.1`). The branch pin is here only so CI can run against the change; it must not merge in that state.
- [ ] `LinkingTo: phylloptim (>= 0.1.1)` is already set and needs no further change, provided phylloptim#65 lands with its version bump.

## Sequence, for the record

1. [plant#605](https://github.com/traitecoevo/plant/pull/605) — the #576 bug fix itself. Independent of all of this; does not need either message change.
2. [odelia#53](https://github.com/traitecoevo/odelia/pull/53) — point and domain in `Interpolator::eval`, tagged 0.2.2. Covers every spline in the family.
3. [phylloptim#65](https://github.com/traitecoevo/phylloptim/pull/65) — spline, units and caller for the transport pair, tagged 0.1.1.
4. **This PR** — plant's assertions follow phylloptim.

Only step 4 depends on an earlier step. 1, 2 and 3 can land in any order.

dfalster:
## Unblocked, and rebased

Both upstream halves landed on 2026-08-05: odelia#53 (0.2.2, names the point and the domain) and phylloptim#65 (0.2.1, names the spline, the units, the end and the caller). plant reached both via #633, which took `develop` to phylloptim 0.6.0 / odelia 0.3.1 — so the original DESCRIPTION hunk (floor `>= 0.1.1`, `Remotes` at phylloptim#65's branch) is obsolete and is dropped. The branch is now a single commit on top of `develop@b374ce3b`, touching `NEWS.md` and `tests/testthat/test-leaf.r` only.

## What the two assertions used to say, and what changed

#633 already replaced odelia's bare sentence with `"evaluated outside its domain"`, so this PR is the second half: naming the parts. The actual message at both call sites, printed from the installed build:

```
[phylloptim:infeasible:stem_curve_domain] Leaf hydraulics: transpiration_from_psi
(the cumulative xylem conductivity integral G, psi in +MPa) evaluated outside its
domain: psi = 9.93826 lies 3.59603 beyond the upper end of [0, 6.34223]; asked by
Leaf::transpiration, at psi_stem.
```

## The soil-side case names the stem argument

The original PR asserted only a bare `"Leaf::transpiration"` on the second case, on the reasoning that it is "reached from the soil side". Measured, it is not distinguishable from the first: `psi_stem = psi_soil = upper_bound_int` there, and `Leaf::transpiration` reads the spline at the stem argument first, so it throws before the upstream read. The label is `at psi_stem` in both cases. Asserted explicitly rather than left to a generic match, with the reason in a comment — otherwise the stem label on a soil-side test reads as a bug.

## Verification

- `testthat::test_local(filter = "^leaf$")` — **213 pass, 0 fail**, against a `make clean` + `R CMD INSTALL` of this branch (phylloptim 0.6.0, odelia 0.3.1 installed).
- Negative check, because a message assertion that reads nothing still passes: mutating the expected caller to `at psi_upstream` fails **both** new blocks with the real message quoted. The three `expect_match` calls bite.
- Scope: test text and NEWS only, one test file touched. No C++, no R source, no fixtures — nothing outside `test-leaf.r` can move.

Issue #599, opened 2026-08-05
Dense TF24 stochastic runs throw on an unattainable psi_soil (43% of seeds on develop)


dfalster:
Dense TF24 stochastic runs throw out of the leaf's collar root-find on a large fraction of seeds. This is in `develop`'s model code, not in any open PR, and it is currently invisible because no test exercises the configuration.

## Measurement

`run_stochastic_collect()`, TF24, `patch_area = 1` with default parameters — about 105 arrivals per m², so ~105 individuals on one square metre — over seeds 1..40, on macOS:

| model code | seeds that throw |
|---|---|
| `develop` | **17 / 40 (43%)** — seeds 1, 4, 5, 7, 9, 10, 11, 13, 16, 19, 21, 26, 28, 31, 37, 38, 40 |
| `develop` + #593 | **5 / 40 (12.5%)** — seeds 2, 4, 6, 13, 30 |

Both columns are the same R code and the same seeded schedules; only the C++ differs. So integrating the environment in the stochastic solver (#593) makes this markedly *better*, and neither version is sound.

Almost all failures are:

```
find_root_psi(find_root_crit=1) failed: Error in function boost::math::tools::toms748_solve<double>:
Parameters a and b do not bracket the root: a=4.2201073133559694e-08; min=0.000000; max=7.085493
```

One seed (6) instead reports `Integrand returned NaN at x=nan`.

## Why the reported ψ cannot be a soil state

`a` is `wettest_soil_layer`, the minimum over layers of the ψ_soil vector plant hands the leaf, which comes from `TF24_Environment::psi_from_soil_moist`. That function cannot return 4.2e-08 MPa. Its smallest possible value is at saturation:

    a_psi / 1e6 = 1780 / 1e6 = 1.78e-03 MPa

which is five orders of magnitude larger. I confirmed this is not a missing clamp: adding `std::min(theta, theta_sat)` to `psi_from_soil_moist` (mirroring the clamp `soil_K_from_soil_theta` already has, and which its comment says exists for exactly this reason — intermediate explicit-RK stages probing θ far above saturation) changes nothing. Same seeds, same values.

The maximum *accepted* θ over a whole run is 0.3113 against θ_sat = 0.428, i.e. ψ ≈ 0.0146 MPa. So the run never accepts a state anywhere near the reported ψ either.

[...]

PR #565, opened 2026-07-11
Design doc: multi-patch water redistribution

dfalster:
Adds `notes/427-multipatch-water.md`, a design doc with no implementation, for
the water-flow arm of the multi-patch epic.

It separates three water processes — event-time surface redistribution,
within-patch vertical buckets, donor-controlled subsurface flux — toggled by
climate and soil rather than by a hand switch; puts N patches and their
topographic and soil attributes in a `Landscape` container that leaves `Patch`
and `Environment` largely untouched; and uses donor control to make the
cross-patch Jacobian triangular, so a single downslope sweep replaces a stiff
coupled ODE. It also records the soil-vs-redistribution identifiability hazard
and a sequencing that covers the Mulga target in its first two steps.

Refs #427. Draft, for circulation rather than merge.

Issue #522, opened 2026-06-24
[TF24 hydraulics] [Env drivers] Enable event-driven hydraulics


dfalster:
Tracking a future possibility — **event-driven hydraulics**, and more generally a
discrete-**events** capability in `plant`.

## Motivation

The current TF24 hydraulic model takes rainfall as an *average* daily amount via a
smooth spline. That works, but it smears out the variability that matters
ecologically — in particular, large **rainfall pulses**. We'd like to be able to
represent rainfall (and other inputs) as discrete events.

## Suggested implementation

Model a rainfall pulse as a **discrete event**:

1. integrate the ODE system up to the event time;
2. apply the event — a sudden change in soil water; then
3. restart/continue the solver from the new state.

This is the standard "stop-the-integrator, apply a jump, resume" pattern for
discontinuities in an ODE system.

## Towards a general events capability

We effectively already have events in the model — **node (cohort) introductions**
are exactly this. The proposal is to generalise that into a first-class **events
mechanism**:

- a general `events` class/queue of `(time, action)` items applied during a run;
- rainfall pulses become one event type; node introduction another;
- leaves room for future event types, e.g. **management interventions**
  (thinning, harvest, planting).

This should probably be passed in **alongside or as part of the drivers**, so
events and continuous environmental forcing share one interface.

## Notes

- Cross-cuts `[TF24 hydraulics]` (the motivating use case — soil-water jumps) and
  `[Env drivers]` (events as a sibling to continuous driver forcing).
- ~~The fixed-step explicit solver used for TF24 soil water will need care here:
  event times must coincide with (or force) step boundaries so the jump isn't
  smeared or skipped, and any cached state/schedule is reset cleanly after a jump.~~
  **Corrected — there is no fixed-step solver.** The whole patch, environment included, is
  one flat system on odelia's *adaptive* RKCK stepper (`patch.h:827-829`, `scm.h:283`);
  `control.fixed_time_step` (forward Euler) exists but defaults to 0. The claim traces to a
  stale comment at `tf24_environment.h:363`, which a correct comment 30 lines later
  contradicts. The genuine concerns are different — see the scoping comment below.
- Reusing the existing node-introduction machinery as the prototype for the
  events queue is the natural starting point for scoping. **Confirmed: it is more than a
  prototype — the stop/apply/resume mechanism already exists in full.**
  `advance_adaptive(vector<double>)` lands on supplied times exactly, and
[...]

PR #608, opened 2026-08-05
Design pass: the soil cascade needs no change

dfalster:
theta_sat is already a barrier in the exact vector field, and error
control already rejects the overshooting steps. The excursions to
theta = 1.7 are explicit overshoot triggered by the step size a leg
inherits, not a water-balance defect.

Two faults instead: odelia accepts a NaN error estimate (odelia#52),
and a throw inside a stage evaluation preempts rejection entirely.
Capacity limiting is measured to cost stiffness for no benefit.

Design doc plus standalone scripts; no package code changes.
Refs #522, #599, #565.


dfalster:
## What this was convened to do, and why it concluded the opposite

Scoping #522 (rainfall pulses) produced a plan whose first stage was a redistribution fix in `compute_rates`: the cascade sets `water_input[i] = K(θ_{i-1})` and never checks whether layer `i` can accept it, so a saturated layer appeared able to overfill the one below — and pulses looked unsafe until that was fixed. The gate before implementing was to design the scheme properly. It does not survive measurement.

Everything below is from `notes/scripts/soil_cascade/` — a standalone R replica of `compute_rates`/`soil_K_from_soil_theta` plus odelia's exact `OdeControl` with plant's defaults. No plant build needed; `Rscript exp1.R`, `exp2.R`, `exp4.R`.

## 1. θ_sat is already a barrier

Inflow `K(θ_{i-1})` and outflow `K(θ_i)` saturate at the same `K_sat`, so the rate is *exactly zero* at θ_sat, and the `[0, θ_sat]` clamp inside K stops it going positive above:

| θ₁ (of θ_sat) | inflow | outflow | rate₁ (yr⁻¹) |
|---|---|---|---|
| 0.050 (0.117) | 163.041 | 0.000 | **+543.47** |
| 0.385 (0.900) | 163.041 | 29.769 | +444.24 |
| 0.424 (0.990) | 163.041 | 138.627 | +81.38 |
| 0.428 (1.000) | 163.041 | 163.041 | **0.00** |

So θ cannot exceed saturation in exact arithmetic. One Euler step from θ₁ = 0.05: h = 1e-3 → 0.593; **h = 0.08 → 43.5**, reproducing the recorded `θ = [-51.4, 52.0, …]`. Pure explicit overshoot.

## 2. Error control already rejects those steps

h = 1e-3, 1e-2 and 0.08 all **reject**. From a cold start the worst case integrates in **15 steps** with max θ = 0.428 exactly; `ode_step_size_max` barely matters (15 vs 21 at 0.05).

## 3. The trigger is the inherited step size

`step_size_last` persists across `set_state_from_system()` (`stochastic_patch_runner.h:146-159`), so a leg beginning after a discrete change starts at whatever the previous quiet leg grew to. Varying *only* that:

| h at leg start | leaf throws | leaf returns NaN |
|---|---|---|
| 1e-6 (fresh) | completes, 16 steps | completes, 16 steps |
| ≥ 1e-3 | **dies at step 0** | **dies, NaN accepted** |

Two independent faults:

- **odelia accepts a NaN error estimate** — `std::max(NaN, rmax)` is NaN and `NaN > 1.1` is false, so the accept branch runs. Filed as traitecoevo/odelia#52. Fixing it makes that path recoverable (14–15 steps, 9–12 rejections).
- **A throw inside a stage evaluation preempts rejection** — `set_ode_state` → `check_finite_ode_state()` runs at *every* RK stage, and phylloptim's collar root-find throws. The run dies having taken **zero** steps, and fixing the NaN does not help it at all.

## 4. Capacity limiting: measured, and rejected

| scheme | max &#124;upper-tri&#124; | lower-tri? | max &#124;λ&#124; |
|---|---|---|---|
| baseline (donor-only) | 0 | yes | 1.546e4 |
| receiver rejects excess | 0 | yes | 1.546e4 |
| donor throttles | 4.67e3 | **no** | 1.518e4 |

Donor-throttling — the literal reading of "a layer can't overfill the one below" — breaks lower-triangularity, which **vindicates the original donor-only choice**. Receiver-side rejection keeps it but costs **1.20× stiffness at 0.9 θ_sat and 1.45× at 0.99**, and at 0.99 turns a **+81 yr⁻¹ rate into −420 yr⁻¹** by discarding inflow the baseline retains — a real hydrological change needing a `scientific_version` bump and a scenario re-bless, to enforce a bound that already holds.

## 5. Relation to #565

Same criterion, independently: #565 uses donor control to triangularise the *cross-patch* Jacobian under the "design out stiffness rather than reach for implicit solvers" philosophy. This pass supplies measured backing for two things it asserts — the stiff diagonal (≈1.5e4 yr⁻¹, structure exactly triangular to machine precision) and the need for a spill term on large events. It also flags that its N = 15 bucket makes the §3 trigger **worse**: measured max |λ| at 0.99 θ_sat goes **1.77e4 (N=5, dz=0.3) → 5.30e4 (N=15, dz=0.1)**.

## 6. Consequence for #522

**Pulses are no longer gated on a physics change.** They need the event queue, the pulse action, and the R generator. The one physics-adjacent piece that *is* required is capping the jump at layer 0's free capacity, because a jump is applied outside the integrator where nothing can reject it — and a ~13 mm event already exceeds capacity from θ₀ = 0.40.

## Caveat

These experiments isolate the soil column with a stub for physiology, so they establish the *mechanism*, not the 43%-of-seeds figure. The 40-seed harness should be re-run against real fixes before #599 closes.

## Retractions from my own earlier notes on this

- I claimed on #522 that a general events queue would need care around "the fixed-step explicit solver". There is no fixed-step solver; that issue's body is now corrected.
- I claimed RODAS was unreachable without a multi-week scalar-templating project. True of `RodasStep` as gated, but odelia#42's `BlockFdJacobian` needs no `rebind` — see the addendum on #599.

Issue #615, opened 2026-08-14
[TF24 hydraulics] Height-dependent stem hydraulic resistance from anatomical profiles


dfalster:
## Summary

Replace TF24's linear height–resistance relation with a spatially resolved anatomical model of the stem, integrated to a closed form, in which the **height exponent is derived** from two independently measurable quantities — the conduit widening exponent (`D_c`) and the Huber-value profile exponent (`theta_c`) — rather than being assumed or fitted.

Full implementation plan: [`notes/plan-tf24-height-hydraulics.md`](https://github.com/traitecoevo/plant/blob/feature/tf24-height-hydraulics/notes/plan-tf24-height-hydraulics.md) (added on `feature/tf24-height-hydraulics`). Sub-epic of #424.

## Current behaviour

[`src/tf24_strategy.cpp:385`](https://github.com/traitecoevo/plant/blob/develop/src/tf24_strategy.cpp#L385):

```cpp
const double leaf_specific_conductance_max = pars.K_s * pars.theta / (height * eta_c);
```

Resistance is therefore strictly linear in height. Over a seedling-to-mulga range (0.3 → 8 m) that is a **27-fold** increase in resistance, across exactly the size range where most demographic action happens. An earlier rationalisation of linear scaling as the correct tall-tree asymptote is **retracted** in the plan (§1): it required a conduit-diameter cap whose mechanism is unknown. Absent a mechanistic plateau, linear is a convenient default, not an asymptote.

## Model

Coordinate `L` = distance from apex along the flow path; `L_tip` = terminal segment length.

- **Widening**, `D(L) = D_tip (L/L_tip)^D_c`, `D_c ≈ 0.2`, **no upper cap on D** (§2.1 — the cap is retained only as a post-hoc falsification diagnostic, §4.2).
- **Packing limit**, `n_A ∝ D^-2` under conserved lumen fraction — so sapwood-specific conductivity scales as `D^2`, not `D^4`: `k_s(L) = K_s (L/L_tip)^{2·D_c}`. This is the single most consequential correction relative to a naive single-tube treatment.
- **Huber profile**, `theta(L) = theta (L/L_tip)^{-theta_c}` with `theta_c = 2·D_c` as a *derived* default (zero new fitted parameters). Note the **minus sign**: θ falls basipetally while the Huber value 1/θ rises.

With `β ≡ 2·D_c + theta_c`, the integral closes to an effective path length `L_eff`, and `R_L = (theta/K_s)·L_eff`:

```
[...]

PR #617, opened 2026-08-14
Derive TF24 height-resistance from stem anatomy

dfalster:
Replaces stem resistance linear in height with a path integral over two
within-plant profiles, so the height exponent is derived from conduit
widening and the Huber profile rather than assumed.

Landed inert and bit-identical first, then switched on with K_s holding
resistance fixed at 1 m (scientific_version -> 10) and B_Hv1 re-anchored so
the move does not shift the vulnerability curve. theta_c is refused: theta
is read by the carbon budget too, so it profiles everywhere or not at all.

Epic #615, phases 0-2a. P5 is untested.


dfalster:
## What to look at

Four commits. The diff is best read in order, because the whole design of the change is that step 2 is provably inert and step 4 is where the science moves.

| commit | what |
|---|---|
| `184f7cc2` | the design note, as written |
| `25435e02` | seven corrections to it, found by reading the code |
| `62aad21c` | **Phase 1** — closed form, parameters inert, no version bump |
| `e02c562e` | **Phase 2a** — defaults on, `K_s` reparameterised, `scientific_version` → 9 |

The core is 30 lines: [`inst/include/plant/stem_hydraulics.h`](https://github.com/traitecoevo/plant/blob/feature/tf24-height-hydraulics/inst/include/plant/stem_hydraulics.h) and its one call site, which replaces

```cpp
const double leaf_specific_conductance_max = pars.K_s * pars.theta / (height * eta_c);
```

Nothing in phylloptim changes — `NEWS.md` already recorded that plant computes `kmax` from height and hands the leaf a scalar "because which conductance-versus-height model is in force is not the leaf's business".

## Verification

**Phase 1 inertness** is the load-bearing claim, since it is what makes the I7 regression gate real rather than asserted. Bit-identical at 17 significant figures on the one-species and two-species SCM scenarios in *both* density coordinates, and on all 8 gateway scenarios. The `beta == 0` branch returns its operand with no arithmetic performed at all, which is what makes it bitwise rather than approximately equal — see the comment there about `-ffp-contract`.

**Phase 2a movement**, all measured before any pin was touched:

| | before | after | Δ |
|---|---|---|---|
| one-species SCM | 81.8571 | 75.5522 | −7.70% |
| two-species, fast | 67.3220 | 62.9266 | −6.53% |
| two-species, slow | 2.7745e−4 | 2.0835e−4 | −24.91% |
| birth-date, fast | 287.1604 | 253.9007 | −11.58% |
| birth-date, slow | 59.5320 | 49.4161 | −16.99% |
| seeded stochastic | 81 / 3 | 67 / 1 | — |

Every SCM test runs at `hmat` = 5 m, entirely below the 16.6 m anchor, so they all sit in the more-resistant half of the rotation and all move down. The gateway runs at the default `hmat` and straddles the anchor, so it moves **both ways**: S01 +9.7%, S02 +93.5%, S07 +3.6%, S03–S06 and S08 −71% to −91%. 8/8 still run, 2/8 still persist, **no classification flips on either axis**; the large negatives are all on scenarios already at ~1e−13 offspring.

The model-version snapshot moved exactly four values per model — `model_id`, `D_c`, `K_s`, `L_tip` — and nothing else. Checked before accepting.

## Three things a reviewer should push on

**1. The `eta_c` convention.** The flow path runs to `H·eta_c` (0.886·H, the leaf-area-weighted mean leaf height), not to `H`. I made `eta_c` scale the *upper limit* of the integral rather than the resistance. The two readings differ by the constant `eta_c^(−β)` — 1.05× at β = 0.4 — which is H-independent and therefore wholly absorbed by the reparameterisation, so nothing downstream can distinguish them. I picked the reading whose derivation is honest, but it is a choice and it is worth a second opinion.

**2. `B_Hv1`.** `make_TF24_hyperpar` derives the whole vulnerability curve from `K_s`, so reparameterising `K_s` would have moved `p_50` from 2.889 to 4.438 MPa as a side effect. I re-anchored the intercept to hold `p_50` fixed. **But the cheaper question is whether that was the right call at all**: if the Ks–p50 relation was originally fitted on terminal branch segments — which is the usual measurement unit — then keying it on tip conductivity is *more* correct than before, and the re-anchoring should be undone rather than kept. I do not know what it was fitted on. Logged as SI-2.

**3. The advertised `H^0.2` is optimistic.** The height elasticity is `e/(1 − x^{−e})` with `e = 1−β`, not `1−β`; for β ≥ 1 it tends to **zero** (the saturating regime), not to `1−β`; and convergence is slow and slows as β grows — at β = 0.8 it is still 3.0% above its asymptote at H = 10⁶ m. The design note's §4.1 table quotes asymptotics and understates the closed form accordingly (β = 0.8 is 3.3×, not 1.9×, over 0.3→8 m). Over any real height range the realised exponent is meaningfully steeper than 0.2. This matters for how P5 and P7 get argued.

## Deliberate departures from the design note

- **§4.4's algebraic form is not implemented.** `L_tip^β(H^(1−β) − L_tip^(1−β))/(1−β)` retains ~4 significant digits at 1−β = 1e−12, and its `L_tip` = 0 admissibility rests on `pow(H, 1.0) == H` — a libm courtesy, not an IEEE-754 or C99 Annex F guarantee, and not something to hang a bit-identity gate on. Implemented with `expm1`; there is a test at 1e−12 that fails if anyone simplifies back.
- **No `hcost` rename.** There is no Sperry hydraulic-cost `theta` in either repo — it is `hydraulic_cost_` throughout. The *real* `theta` collision is phylloptim's `Leaf$theta_`/`theta_w_`/`theta_fc_`, which are soil water content and R-visible in the same session.
- **I11's schema gate does not exist.** plant has `scientific_version`/`model_id()`, a model version for logpile's cache, not a parameter-file format gate. Phase 2b has to build it.

## One behavioural change outside TF24

TF24f's convergence onto TF24 widened from under 1% to 1.19% at `k_acclim` = 10. I checked whether this was structural before touching the test: it is a pure tracking lag, closing monotonically with gain (1.186% / 0.396% / 0.107% / 0.020% at k = 10/30/100/300). The test now asserts convergence at k = 100 rather than having its tolerance widened, and additionally asserts the approach is monotone.

## Not in this PR

Phases 3–6: gravity as an explicit ρgH term, the `D_implied` falsification diagnostic, the diurnal closure, and the sweeps against I1–I11. **P5 is untested** — resistance at 60 m is now 0.60× what it was, and nothing here checks whether ESS height is still bounded. If it is not, the answer per I3 is to name the missing cost, not to re-steepen R.

Two staged inconsistencies carry a ledger with closure dates in the design note. SI-1 (structural vs hydraulic `theta`) is currently **latent** — `theta_c` = 0 means the two uses still agree numerically, and it only bites at 2b. A related trap: `dmass_sapwood_darea_leaf` hard-codes the closed-form derivative instead of differentiating `mass_sapwood`, so a profiled structural θ must be re-derived by hand and **nothing in the suite will fail if it is not**.

Full background: #615.

NEWS.md
```diff
@@ -294,6 +294,81 @@ were not previously recorded here:
...
+* **TF24 stem hydraulic resistance is now a path integral over the stem**, with
+  three new `TF24_Pars` fields — `D_c` (conduit widening exponent, default
+  `0.2`), `theta_c` (Huber-profile exponent, default `0`) and `L_tip` (terminal
+  segment length, default `0.02` m). Resistance now grows as `H^0.6` rather than
+  `H^1.0`. **`scientific_version` moves to 9 and TF24 output changes** (#615).
+
+  `K_s` keeps its name but narrows in meaning to the **terminal segment**, and
+  its default moves `1 -> 0.11681288783361685` (a factor 8.5607) so that
+  resistance is *unchanged* at `TF24_H_ANCHOR` = 16.5958691 m, TF24's default
+  `hmat`. Use the exported `TF24_K_s_at_tip()` to convert a `K_s` calibrated
```
dfalster: Update so that TF24_H_ANCHOR =1, i.e. keep resistance unchanged at height = 1m.

inst/include/plant/stem_hydraulics.h
```diff
@@ -0,0 +1,87 @@
...
+#ifndef PLANT_PLANT_STEM_HYDRAULICS_H_
+#define PLANT_PLANT_STEM_HYDRAULICS_H_
+
+#include <cmath>
+
+namespace plant {
+namespace stem_hydraulics {
+
+// Effective hydraulic path length of a stem whose anatomy varies as a power of
+// distance-from-apex L (#615; notes/plan-tf24-height-hydraulics.md sec. 4).
```
dfalster: Don't include issue numbers throughout. We can tell this from commit history

inst/include/plant/stem_hydraulics.h
```diff
@@ -0,0 +1,87 @@
...
+//
+//   D(L)     = D_tip * (L/L_tip)^D_c            conduit diameter (widening)
+//   n_A(L)  propto (D(L)/D_tip)^-2              conduit density (packing limit)
+//   k_s(L)   = K_s   * (L/L_tip)^(2*D_c)        sapwood-specific conductivity
+//   theta(L) = theta * (L/L_tip)^(-theta_c)     leaf area per sapwood area
+//
+// The exponent on conductivity is 2*D_c, NOT 4*D_c: the D^4 of Hagen-Poiseuille
+// survives only along a single continuous conduit, and at the tissue level a
+// conserved lumen fraction means widening is paid for by proportionally fewer
+// conduits, halving the effective exponent. That factor 2 is the single most
```
dfalster: That factor 2 is the single most....  -> delete

src/tf24_strategy.cpp
```diff
@@ -836,6 +850,34 @@ void TF24_Strategy::prepare_strategy() {
   height_0 = height_seed();
   area_leaf_0 = area_leaf(height_0);
 
+  // Stem hydraulic path (#615). The factor 2 on D_c is the packing limit: under
```
dfalster: what is path_beta_? "Stem hydraulic path" name is incomplete.

src/tf24_strategy.cpp
```diff
@@ -836,6 +850,34 @@ void TF24_Strategy::prepare_strategy() {
...
+  // a conserved lumen fraction, widening is paid for by proportionally fewer
+  // conduits, so sapwood-specific conductivity scales as D^2 and not the D^4 of
+  // Hagen-Poiseuille. See plant/stem_hydraulics.h.
+  path_beta_ = 2.0 * pars.D_c + pars.theta_c;
+  if (path_beta_ != 0.0) {
```
dfalster: Add comment what this implies -e.g. ie. resistance scales non linearly with height?

src/tf24_strategy.cpp
```diff
@@ -836,6 +850,34 @@ void TF24_Strategy::prepare_strategy() {
...
+    // L_tip is the anchor of both profiles, so it cannot be zero once either is
+    // active: k_s(L) = K_s*(L/L_tip)^(2*D_c) diverges everywhere as L_tip -> 0,
+    // giving zero resistance. That is a degenerate configuration to reject, not
+    // a numerical edge case to tolerate.
+    if (!(pars.L_tip > 0.0)) {
```
dfalster: !(pars.L_tip > 0.0) -> pars.L_tip < 0.0

tests/testthat/test-stochastic-patch-runner.R
```diff
@@ -211,7 +211,10 @@ test_that("collect output is reproducible and matches a seeded baseline (#482)",
   ## entry under Breaking changes in NEWS.md) and is not re-checked here.
   baseline <- list(
     FF16 = list(n_total = 83L, n_alive_final = 5L),
-    TF24 = list(n_total = 81L, n_alive_final = 3L),
+    # Moved 81/3 -> 67/1 at #615 Phase 2a: the stem path integral with
```
dfalster: don't need details in file. Put them in commit documentation instead

R/tf24.R
```diff
@@ -293,6 +309,68 @@ make_TF24_hyperpar <- function(lma_0=0.1978791,
...
+##'
+##' @param K_s Whole-stem sapwood-specific conductivity under the old model.
+##' @param D_c Conduit widening exponent.
+##' @param theta_c Huber-profile exponent.
+##' @param L_tip Terminal segment length [m].
+##' @param H_anchor Height at which old and new models agree [m].
+##' @param eta Canopy shape parameter, used for the crown-centre coordinate.
+##' @return The terminal-segment `K_s`.
+##' @export
+TF24_K_s_at_tip <- function(K_s, D_c = 0.2, theta_c = 0, L_tip = 0.02,
```
dfalster: Does this function have any enduring value beyond mapping to the old model? If we're defining values at the tip feels silly that we have a function which calculates it at the tip.

tests/testthat/test-strategy-tf24.R
```diff
@@ -287,7 +295,7 @@ test_that("offspring arrival", {
                        hyperpar = TF24_hyperpar, birth_rate = list(20))
 
   out <- run_scm(p1, env, ctrl)
-  expect_equal(out$offspring_production, 82.09077702, tolerance = 2e-2)
```
dfalster: Can't we run test so that we use values compatible with old model, to show that retains existing behaviour?

tests/testthat/test-strategy-tf24f.R
```diff
@@ -62,6 +65,11 @@ test_that("Defaults", {
     root_b = 3.898245,
     root_psi_crit = 3.898245 * log(1 / 0.05)^(1 / 2.680147),
     rooting_depth_max = 1.5,
+    # Stem hydraulic path (#615). Widening is on since Phase 2a; theta_c stays
+    # 0, so `theta` keeps its whole-plant meaning. See plant/stem_hydraulics.h.
+    D_c = 0.2,
```
dfalster, line 70: Wouldn't we set this to zero two to maintain consistency with previous tests?

NEWS.md
dfalster: Done. `TF24_H_ANCHOR` = 1 m, so `K_s` goes 1 → 0.33577377016801868 (ratio 2.9782, was 8.5607) and `B_Hv1` follows to 0.36591565341924093. Resistance is unchanged at 1 m and falls from there:

| H (m) | 0.394 | 1.00 | 5.00 | 8.00 | 16.60 | 30.0 | 60.0 |
|---|---|---|---|---|---|---|---|
| `R_new/R_old` | 1.38 | 1.00 | 0.55 | 0.46 | 0.35 | 0.28 | 0.21 |

**This turned up the most interesting result in the PR: the sign of the effect depends on stand density.**

Individually a plant is better off wherever it is taller than the anchor — single plant, wet soil, no competition, assimilation ratio 0.997 at 0.5 m, 1.001 at 1 m, 1.055 at 5 m, 1.143 at 10 m. Exactly the intended rotation. But lower resistance also means faster transpiration, so in a dense stand everyone draws the shared soil column down faster and the patch does *worse*. One-species SCM, hmat = 5, max_patch_lifetime = 5:

| birth_rate | 0.5 | 2 | 20 |
|---|---|---|---|
| ratio | 1.216 | 1.095 | **0.857** |

The pinned tests all run at birth_rate = 20 — the least favourable end — so they show a 4–17% *drop*. The gateway runs longer patches at the default hmat and moves the other way: up 2.2×–320× on every scenario, with S02 crossing R0 = 1 so persistence goes 2/8 → 3/8. Recorded in both NEWS and the v9 changelog, because either set of numbers read alone gives the wrong impression.

inst/include/plant/stem_hydraulics.h
dfalster: Done — stripped from all the new code comments, tests and R docs.

Two kept deliberately, both inside the `scientific_version` changelog block, because every existing entry there cites its issue (`v3 (#517/#554)`, `v4`, `v5`…) and that block is the one place the established convention is to name the change. Happy to pull those too if you would rather it were uniform.

inst/include/plant/stem_hydraulics.h
dfalster: Deleted.

src/tf24_strategy.cpp
dfalster: Renamed `path_beta_` → `stem_path_exponent_`, and the comment now says what it *is* rather than only how it is computed: the exponent of the stem path integral, `beta = 2*D_c + theta_c`, with leaf-specific resistance growing as `H^(1-beta)`.

src/tf24_strategy.cpp
dfalster: Added:

```cpp
// What it implies: leaf-specific resistance grows as H^(1-beta) rather than
// linearly with height. beta = 0 is the linear case; the default beta = 0.4
// gives H^0.6; beta >= 1 saturates, so resistance approaches a finite limit no
// matter how tall the plant grows. Larger beta therefore means a weaker height
// penalty on carbon gain.
```

src/tf24_strategy.cpp
dfalster: Changed, but not to `< 0.0` — that would switch the guard off for the case it exists to catch.

`L_tip == 0` is the dangerous value: `k_s(L) = K_s*(L/L_tip)^(2*D_c)` diverges everywhere as `L_tip → 0`, giving zero resistance. `!(x > 0.0)` is true for zero; `x < 0.0` is false for zero, so a zero `L_tip` would sail through and the symptom would be an unattributable NaN inside the leaf solver. Negatives are the case that would never occur in practice.

Your underlying point about the double negative being unreadable is right though, so it now reads:

```cpp
// Written to reject zero and NaN as well as negatives: `L_tip < 0.0` alone
// would let a zero through, which is the case this guard exists for.
if (pars.L_tip <= 0.0 || std::isnan(pars.L_tip)) {
```

The explicit `isnan` is there because `<= 0.0` is false for NaN, whereas the old `!(x > 0.0)` caught it for free. There is a test covering zero, over-long, and the beta == 0 case where `L_tip` is legitimately ignored.

tests/testthat/test-stochastic-patch-runner.R
dfalster: Done — comment removed, the numbers live in the commit message and the `scientific_version` changelog.

Worth flagging that the value moved again after the re-anchor: 81/3 → **104/3**, not the 67/1 you saw. Same reason as everything else on this PR — at a 1 m anchor plants above 1 m are cheaper, so more of them survive.

R/tf24.R
dfalster: Fair, and you are right that the name was the problem. Renamed `TF24_K_s_at_tip` → **`TF24_K_s_from_whole_stem`**, which says which direction it converts.

I have kept it rather than deleted it, on this argument: `K_s` is *defined* at the tip but it is almost always *reported* for a stem section at some height — that is the unit people publish. So the conversion is a recurring need for anyone bringing a literature value in, not just a one-off migration from the old model. Its docs now lead with that rather than with the migration story.

It also does real work here: the shipped `K_s` default and the test that asserts it both call it, so the constant in the header and the constant in the test cannot drift apart, and the `eta_c` convention cancels out of the ratio instead of being baked into a hand-computed number.

If you still think it should go, it is a one-line delete plus a literal in the header — say so and I will.

tests/testthat/test-strategy-tf24.R
dfalster: Yes — added, and it is the strongest test on the PR. New `test_that("the height-linear parameters reproduce the pre-path-integral results")` in the same file: set `D_c`, `theta_c`, `L_tip` to zero and `K_s` back to 1, then assert **this file's original pinned numbers, unmodified** — 82.09077702, 67.54060383, 287.16043704, 59.53195639. All four pass at the same 2e-2 tolerance.

That is end-to-end through full SCM runs in both density coordinates, not just the closed form, so it demonstrates the path integral adds a capability rather than changing existing behaviour.

Writing it caught a bug in my own test helper, which is worth recording: the first version reset only `p$strategies[[1]]`, so in the two-species cases species 2 was still on the path integral and I was comparing two different models. It now loops over all strategies, with a comment saying why.

tests/testthat/test-strategy-tf24f.R
dfalster: TF24f cannot differ here — it reuses `TF24_Pars` wholesale (`inst/RcppR6_classes.yml`, the `TF24f_Strategy` block nests the same `pars` type), so `D_c` is 0.2 for both by construction. Setting it to 0 in this list would just make the test fail.

I think the consistency you are after is now provided by the new test on the TF24 side (see my reply on test-strategy-tf24.R:290): the height-linear configuration is asserted explicitly, against the original pinned values, rather than being implied by the defaults. If you would like the same explicit old-model check mirrored for TF24f, say so — it is cheap, though TF24f's own convergence-to-TF24 assertion already covers much of it.

Two TF24f behaviours did move and are documented in that file:

- the tracking lag against TF24 at `k_acclim` = 10 is 0.376% (was under 1%), closing monotonically to 0.002% by k = 1000, so assertion (3) now demonstrates convergence at k = 100 rather than tolerating the lag;
- assertion (2)'s contrast fell from just over 10% to 8.7%, so its threshold is now 0.05. I checked whether widening the k bracket would restore the margin and it does **not** — offspring production is non-monotone in `k_acclim`, with a maximum near k = 0.1 (69.46 / 70.88 / 76.61 / 69.93 at k = 0.001 / 0.01 / 0.1 / 10), so a wider bracket gives a *smaller* contrast. That is recorded in the test so nobody rediscovers it.

dfalster:
## The structural/hydraulic θ split is retracted

You never wanted the two distinguished, and you were right — I had carried a staging decision from the design note that does not survive contact with what `theta` actually is.

**The problem.** `theta` is read by the carbon budget as well as the hydraulic term: `area_sapwood`, `area_bark`, their rate forms, `mass_sapwood` (hence construction cost, respiration, turnover and NSC capacity), and the hard-coded `dmass_sapwood_darea_leaf` derivative. Profiling it on the hydraulic side alone gives a plant that *conducts* as though θ varied along the stem while being *built and respired* as though it did not — two different plants sharing one trait name. That is not a staging step, it is an incoherent model, and it being convenient was the only thing recommending it.

**As shipped, the code permitted exactly that.** `theta_c` reached only the path integral; every structural use kept a flat `pars.theta`. So anyone setting `theta_c > 0` would have got the split silently.

**Fixed:** `prepare_strategy()` now throws on any non-zero `theta_c`. The field stays declared so the closed form can still be exercised and swept in isolation, but a strategy carrying it will not build. The guard is on `theta_c` itself, not on β — a beta-only check would miss `D_c = 0.2, theta_c = -0.4`, which sums to zero while still being the forbidden configuration. There is a test for that case.

`D_c` is unaffected and carries the height dependence meanwhile: it enters through `k_s(L)`, which has no structural counterpart to keep in step.

**Consequence for Phase 2b:** θ(L) is now one change, not two. When it lands it lands in the hydraulic path *and* the allometry in the same commit. Harder than the staged version, and the honest cost of the compensation mechanism. The design note's §4.4 staging is marked retracted, the phase table is rewritten, the SI-1 ledger entry is gone (there is no longer an inconsistency to close), and a new criterion **I13** says θ has exactly one meaning. The `dmass_sapwood_darea_leaf` trap is flagged there — it hard-codes the closed-form derivative rather than differentiating `mass_sapwood`, so it must be re-derived by hand and nothing in the suite will fail if it is not.

Only SI-2 (the `B_Hv1` re-anchoring) remains, and that one has a cheap resolution: find out what the Ks–p50 relation was fitted on.

---

### One thing worth recording separately

While verifying this I hit `Extrapolation disabled and evaluation point outside of interpolated domain` from TF24f at `k_acclim = 0.1`, which looked like a real crash introduced by the guard. It was not — it was the stale-`.o` trap documented in `agents.md`: I had edited `tf24_strategy.h` and then run `make compile`, which does not clean. After `make rebuild` every gain runs and reproduces the recorded numbers exactly (69.46 / 70.88 / 76.61 / 69.93 at k = 0.001 / 0.01 / 0.1 / 10). Full suite is green with zero failures.

Flagging it because the symptom was a plausible-looking numerical error in an unrelated strategy, which is precisely how that trap disguises itself.

dfalster:
## Corrected: the numbers in my first comment are superseded

That comment was written against `H_anchor` = 16.6 m. The anchor is now **1 m**, so every figure in its movement table and rotation table is stale. Current state, for anyone reading the thread in order:

| | value |
|---|---|
| `H_anchor` | 1 m |
| `K_s` | 1 → 0.33577377016801868 (ratio 2.9782) |
| `B_Hv1` | 0.4607063 → 0.36591565341924093 |
| p_50 | 2.8887 MPa, unchanged (would have been 3.593 unfixed) |

| H (m) | 0.394 | 1.00 | 5.00 | 8.00 | 16.60 | 30.0 | 60.0 |
|---|---|---|---|---|---|---|---|
| `R_new/R_old` | 1.38 | 1.00 | 0.55 | 0.46 | 0.35 | 0.28 | 0.21 |

**And the headline framing in that comment was wrong, not just its numbers.** It reported a single movement table as though the change had one sign. It does not — the sign depends on stand density:

| `birth_rate` | 0.5 | 2 | 20 |
|---|---|---|---|
| offspring ratio | 1.216 | 1.095 | **0.857** |

Individually a plant is better off wherever it is taller than the anchor (assimilation ratio 1.055 at 5 m, 1.143 at 10 m, single plant, wet soil, no competition), but lower resistance means faster transpiration, so in a dense stand everyone draws the shared soil column down faster and the patch does worse. The pinned SCM tests all run at `birth_rate` = 20 — the least favourable end, −4% to −17% — while the gateway runs longer patches at the default `hmat` and moves the other way by +120% to +32000%, with one scenario crossing R0 = 1 (persistence 2/8 → 3/8).

Quoting any single number for "the effect of this change" without its density is incomplete. Both NEWS and the `scientific_version` v9 entry now say so explicitly.

### Still-standing items from that comment

Unchanged and worth keeping in view:

- the **`eta_c` convention** question (upper limit vs prefactor) — still a judgement call, still unidentifiable from data;
- the **Ks–p50** question — if the relation was fitted on terminal branch segments, keying it on tip conductivity is *more* correct and the `B_Hv1` re-anchoring should be undone rather than kept. This is now the only staged inconsistency left;
- **the advertised `H^0.2` is optimistic** — the elasticity is `e/(1 − x^{−e})`, tends to zero rather than `1−β` for β ≥ 1, and at β = 0.8 is still 3.0% above its asymptote at H = 10⁶ m;
- **P5 is untested**, and more exposed than before: resistance at 60 m is now 0.21× what it was, down from 0.60× at the old anchor.

dfalster:
## Correction: my earlier stale-`.o` explanation was wrong

Further up this thread I attributed an `Extrapolation disabled and evaluation point outside of interpolated domain` error from TF24f to the stale-`.o` trap, on the grounds that `make rebuild` made it go away locally. **That was wrong.** The same error has now failed CI on ubuntu-latest against a clean build. Retracting that explanation.

### Status

| platform | result |
|---|---|
| macOS-latest | pass |
| windows-latest | pass |
| ubuntu-latest | **fail** |

`develop` is green, so this branch introduced it. It is one error, in one test: `test-strategy-tf24f.R:291`, `run_scm(set_k_acclim(pf, 0.1), ...)`.

### What is established

- **It is a time, not a height.** odelia's `Interpolator` defaults to `extrapolate = true` (`interpolator.hpp:147`); the *only* place plant turns it off is `extrinsic_drivers.h:17`. By elimination the out-of-domain evaluation is an **extrinsic driver asked for a value outside its time domain** — not the canopy spline, not a vulnerability curve, and nothing to do with plants growing taller.
- **It is a knife edge.** Locally on macOS I scanned `k_acclim` from 0.02 to 3 and found no cliff at all — a smooth hump peaking near k = 0.13 (76.15 / 76.61 / 76.76 / 76.59 / 75.70 at k = 0.08 / 0.1 / 0.13 / 0.15 / 0.2). Nothing discontinuous, and no error at any point. Two of three platforms pass.
- **The mechanism is plausible and not specific to this feature.** Changing the resistance changes the trajectory, which changes the adaptive solver's step sizes, which changes whether a step lands exactly on the end of the driver's domain or a few ULP past it. A driver tabulated over exactly the integration span has no margin for that.

The installed odelia is 0.2.1 — matching the `Remotes` pin — whose message predates the diagnostic detail (`u = …`, distance beyond, and the domain) that later versions emit. So the CI log cannot say which driver or what value, which is why this is narrowed rather than closed.

### What I have NOT established

Which driver, and what time. I cannot reproduce on macOS, and the 0.2.1 message carries no payload.

### Options

1. **Instrument and let CI answer** — push a temporary `tryCatch` around that run reporting the driver names, their domains, and the patch time reached. One CI cycle (~8 min), decisive.
2. **Pad the driver domain** — if it is the end-of-span knife edge, the fix is to tabulate drivers slightly beyond `max_patch_lifetime` rather than exactly to it. This would be a pre-existing-fragility fix that this branch merely exposed, and arguably belongs in its own issue.
3. **Bump the odelia pin** so the error carries its payload. Useful independently — the whole point of that message's detail is this situation — but it is a dependency change riding on a science PR.

My inclination is 1 then 2, and to file the underlying fragility separately rather than fold it in here. Not proceeding until that is agreed, since 2 and 3 both touch things outside this PR's scope.

dfalster:
## Diagnosed — and it was not what I said it was, twice over

The instrumented probe answered it locally, so no CI cycle was needed.

**My driver hypothesis was also wrong.** I probed every extrinsic driver in this run:

```
DIAG drivers: rainfall, leaf_temp, atm_vpd, PPFD, ca, atm_o2_kpa, wind_speed, atm_kpa
DIAG   rainfall:  constant (no domain limit on probed grid)
DIAG   leaf_temp: constant   ... all eight the same
```

All eight are **constant**, so there is no driver interpolator in this run at all. The elimination behind that hypothesis was incomplete: I had grepped only plant, found that odelia defaults to `extrapolate = true` and that plant disables it in exactly one place, and concluded that place must be the culprit. I never searched phylloptim.

**The actual culprit is in phylloptim:**

```
phylloptim/leaf_model.hpp:2016  transpiration_from_psi.set_extrapolate(false);
phylloptim/leaf_model.hpp:2019  psi_from_transpiration.set_extrapolate(false);
```

Those are the leaf's supply-curve splines, and they fit the symptom far better than a driver ever did: this PR changes stem resistance, which moves the leaf's operating point, and those splines are exactly what the leaf inverts at that operating point.

### There is a real bug in how their domain is built

`cumulative_vulnerability_integral` (`phylloptim/vulnerability.hpp:48`) accumulates its knot grid:

```cpp
double step = psi_max / resolution;
for (double psi = step; psi <= psi_max; psi += step) { ... }
```

`psi += step` accumulates rounding, so after `resolution` additions it can land either side of `psi_max`. The loop therefore yields **`resolution` or `resolution - 1` knots depending on the values of `b` and `c`** — meaning the upper bound of a spline that has extrapolation *disabled* is unstable:

| requested resolution | knots obtained |
|---|---|
| 50 | 49 **or** 50 |
| 100 | 99 **or** 100 |
| 1000 | 999 **or** 1000 |

At TF24's defaults (`b` = 3.4572679, `c` = 2.04, resolution = 100) the grid stops at 7.23583513 against a `psi_max` of 7.30892438 — **one full step, 1% of the intended domain, silently missing.**

`vulnerability_psi_max` even carries a comment saying the inverse "needs the same bound", and is named precisely so the two cannot disagree. The accumulation defeats that: the realised bound is not the function's value.

Filed as traitecoevo/phylloptim#92, with a fix and a suggested regression test.

### What is proven, and what is not

That an unstable domain edge exists in exactly the machinery that throws is established. That it is *the* fault behind the ubuntu-latest failure is **not**. The pinned odelia 0.2.1 predates the message payload that would name the spline and the offending value, so the last link is inference rather than measurement. I would rather flag that than present a tidy story — this is my third hypothesis on this failure and the first two were confidently wrong.

### Unblocking this PR

The fix lives in phylloptim, so it needs a release there and a pin bump here — a dependency change riding on a science PR. Options, in the order I would take them:

1. **Bump the odelia pin first**, so the error names the spline and the value. That converts the inference above into proof before anyone edits phylloptim, and the payload is useful independently — its whole purpose is this situation.
2. **Fix phylloptim#92, release, bump the pin here.** Correct, and the bug deserves fixing whatever happens to this PR.
3. **Leave this PR red and land it after phylloptim.** Honest, and blocks nothing else.

What I would *not* do is move the TF24f test off `k_acclim` = 0.1 to make CI green. The knife edge is real and the test found it.

The diagnostic instrumentation has been removed; the branch is unchanged from `cdd0a5ba`.

dfalster:
Updated from develop (11 commits, `39c0c657..8e49c58f`). Merge commit `a69e5c96`, re-measurement `5c269bb2`. Suite 3346 pass / 0 fail; gateway green under `PLANT_RUN_SCENARIOS=1`.

**`scientific_version` is 10, not 9.** develop took 9 for #619 (bounded storage) while this branch was open, so both sides had written "v9". TF24f auto-tracks, so it is `TF24f@v10.1`.

**Two conflicts were substantive rather than textual:**

- develop renamed the vulnerability-curve fields (`p_50`/`b`/`c` -> `stem_P50`/`stem_b`/`stem_c`, #606). The branch's `K_s` reparameterisation had to be re-applied on top of the new names, and the new `test-tf24-stem-hydraulics.R` was indexing the hyperpar output as `derived[, "p_50"]`, which no longer exists. Fixed; that file passes 209/209.
- the pinned SCM values on both sides were measured against different models. Everything is re-measured below.

**The reduction test now pins develop's own values**, which is what "set `D_c`/`theta_c`/`L_tip` to 0 and `K_s` to 1 and you recover the previous model" has to mean after a merge. It passes with develop's numbers adopted unmodified — the strongest available statement that the merge did not disturb the reduction.

**Re-measured deltas** (height-linear -> path integral, both on this tree):

| quantity | old | new | change |
|---|---|---|---|
| one-species SCM offspring | 30.2980 | 24.3214 | -19.73% |
| two-species, fast | 23.2557 | 18.5430 | -20.26% |
| two-species, slow | 4.0284e-6 | 1.8965e-6 | -52.92% |
| birth-date coordinate, fast | 233.3660 | 219.2668 | -6.04% |
| birth-date coordinate, slow | 43.7240 | 34.5877 | -20.90% |
| seeded stochastic counts | 79 / 3 | 77 / 3 | |

The density-dependent sign flip survives, slightly weaker: one-species SCM ratio 1.196 at `birth_rate` 0.5, 1.047 at 2, 0.803 at 20 (was 1.216 / 1.095 / 0.857). Single-plant assimilation ratios are essentially unchanged, as expected since they do not depend on develop's stand-level work: 0.9968 at 0.5 m, 1.0006 at 1 m, 1.0543 at 5 m, 1.1410 at 10 m.

**The gateway moved a lot more than before** and its baseline is re-blessed. Offspring production is up **3.1x to 2208x** across all 8 scenarios (was 2.2x-320x), and **two** scenarios now cross R0 = 1 rather than one, so persistence goes **1/8 -> 3/8** (the branch previously reported 2/8 -> 3/8 against the pre-#640 baseline). 8/8 still run, 0 crash.

The `R_new/R_old` resistance table is closed-form and unaffected; its assertions pass unmodified.

Issue #610, opened 2026-08-05
Design criteria for bounded states and AD-safe rates: extend model-robustness guidance, then audit


dfalster:
## Summary

`plant-meta/governance/model-robustness.md` already has the right principle — its
"Better than either: design the awkward limit out" section, and a decision table whose
**Never** cell for physically-unattainable states reads *"silently clamp, substitute a
fallback, or continue"*.

`TF24_Strategy::compute_rates` does exactly that:

```cpp
const double S = std::max(vars.state(state_idx_storage), 0.0);
```

So #609 is not a gap in the guidance, it is a **live violation of guidance we already
have**. What's actually missing is two things:

1. A **catalogue of constructive alternatives** — the guidance says don't clamp, but
   doesn't say what to do instead, so the clamp keeps being the obvious move.
2. **Differentiability as a stated criterion.** As we move to AD, `min`, `max` and
   `? :` in a rate are not merely inelegant: a kink in `dS/dt` propagates into every
   downstream sensitivity. The current guidance is silent on this, and it changes the
   ranking of the options.

This issue proposes closing both, and auditing the package for existing instances.

## Proposed criteria (draft — to be expanded in discussion with @dfalster)

The ordering that came out of #609. Stated as a preference ladder, most preferred first:

**1. Choose a functional form where the bound follows from the form.** Factor a rate
into provably non-negative fluxes, each with its own smooth limiter, rather than gating
a signed net flux behind a test. Then the bound is a property of the exact flow and
there is nothing to guard. Worked example in
[#609](https://github.com/traitecoevo/plant/issues/609#issuecomment-5189446308): the
charge/drain split, where `charge - drain == net_flux` exactly, so it lands as a
bit-identical refactor plus a separate commit switching the limiters on.

**2. Choose a basis for the state where the bad region is unreachable.** The ODE
analogue of picking a distribution with the right support: integrate `log S`, or
`S = w²`, and recover the constrained quantity through the link. Two caveats worth
writing down, because both cost us time to work out:

- A transform only enforces a bound the *exact* vector field already respects. It
  cannot fix a model whose flow leaves the domain — it converts a bound violation into
  a blow-up or a stall. So step 1 comes first, always.
- It is not neutral. There is no Jacobian to carry, but it changes what the error
  controller controls: an absolute tolerance on `log S` is a relative tolerance on `S`.

**3. Only then, reject the step** — traitecoevo/odelia#55.

[...]

PR #619, opened 2026-08-14
Bound the NSC storage pool by its own flow

### LINKED ISSUE #609: [TF24] Negative NSC storage state
## Summary

In `TF24_Strategy::compute_rates` the storage pool is read through a clamp:

```cpp
storage = std::max(vars.state(state_idx_storage), 0.0); 
```

Both consumers then use that clamped value: relative reserves

```cpp
r = storage_max > 0.0 ? min(storage / storage_max, 1.0) : 0.0);
```

and the outflow gate

```cpp
floor_gate = (storage + gate_ref) > 0.0 ? storage / (storage + gate_ref) : 0.0;
vars.set_rate(state_idx_storage, net_flux > 0.0 ? net_flux : floor_gate * net_flux);
```

The clamp hides a negative state rather than preventing one.

## Why this matters beyond bookkeeping

The comment at `:1005-1011` states the intent: "flooring storage at zero so relative reserves `r` stay in `[0,1]`", and acknowledges that "at the `S~0` starvation boundary the ungated part of the deficit is untracked". That is accurate about the ungated part, but the frozen-state behaviour is a different thing: the state variable retains a value that no longer participates in its own dynamics. Two patches whose storage overshot to `-1e-6` and `-1e-1` are indistinguishable to every consumer, and both resume from their own stale offset.

## The behaviour

The gate makes the *rate* vanish as the state approaches zero, which keeps the exact ODE from crossing zero. It does not keep a finite RK step from overshooting: from a small positive storage with a strongly negative `net_flux`, a stage can land the state below zero.

Once it is below zero:

- `storage` clamps to `0`, so `r = 0` and mortality sits at its bounded maximum. Correct in direction, but the size of the deficit is not observable.
- `floor_gate` is built from the same clamped value, so it is exactly `0`, so the outflow rate is `0`. The deficit is frozen: it cannot deepen, and nothing pulls it back toward zero.
- If `net_flux` later turns positive, the ungated branch applies and the state recovers from wherever it was frozen — so the plant must repay a debt whose magnitude was never modelled, starting from an arbitrary integrator artefact rather than from zero.

## Suggested resolution

Options, roughly in increasing invasiveness:

1. Clamp the *state* rather than each read — project storage back to zero once per step, so the stored value and the value the model sees agree.
2. Keep the state signed but build `floor_gate` from the raw state so a negative value produces a restoring (positive) rate rather than none.
3. Treat a negative storage state as an integrator failure and reject the step.

Option 1 is closest to the documented intent. Whichever is chosen, the two consumers should read the same quantity, and the comment should say what happens when the state is already negative, not only what happens as it approaches zero.

aornugent:
Both of the storage pool's bounds now follow from the shape of its own
flow, with no clamp on either: dS/dt is a charge and a drain, each
limited by the room the other has, so it cannot be negative at r = 0 nor
positive at r = 1. A step that overshoots anyway is refused and retried
rather than committed -- needs odelia >= 0.3.1. So:

* the state stays inside [0, capacity] on both coordinates, where it hit
  1.035 of capacity and went negative on 3 per cent of records, with
  52.7 per cent of a mature stand sat on the clip;
* dS/dt is smooth across the net flux's sign change, where its slope
  stepped by two at r = 1e-3 and without bound as the pool emptied;
* a run takes 488 accepted steps against 940, and max(S, 0), min(r, 1)
  and drain_ref are gone.

Reserves are mobilised in proportion to what is held, so offspring
production falls 2.7x on height and 19 per cent on birth date. TF24@v9.

Closes #609

Co-authored-by: Claude Opus 5 (1M context) <noreply@anthropic.com>


aornugent:
## The rate

```
charge = Ppos * (1 - G)          drain = Ppos - P
dS/dt  = charge * (1 - r) - drain * r  =  charge - (charge + drain) * r
```

`charge - drain = P - Ppos*G`, the old net flux, so naming the two directions moves nothing by itself. Each is limited by the room the other has: at `r = 0` the rate is `charge >= 0`, at `r = 1` it is `-drain <= 0`, and outside `[0, 1]` both limiters change sign and restore. The pool is a first-order filter on production, so `d(dS/dt)/dS` is the constant `-(charge + drain)/S_max`.

`Ppos >= 0` and `1 - G >= 0` hold without a test, which is what removes the `net_flux > 0` branch and its slope step.

## Why the drain is limited by `r` rather than by `r/(r + D)`

The rate's eigenvalue is available in closed form, because production does not read the pool:

```
d(dS/dt)/dS = (1/S_max) [ Ppos( -G'(r)(1-r) - (1-G) ) - drain * L'(r) ]
```

For `L(r) = r/(r + D)` the last term is `drain * D/(r + D)^2`, so the limiter's knee is an attracting fixed point at `r ~ D` relaxing in `S_max * D / drain`. At `D = 1e-3` that is **0.64 hours** at the stiffest record on the reference stand — a 0.41 m seedling with a storage capacity of 1e-6 kg — while the controller steps in days. The shipped rate has the same knee and never resolves it, because `max(S, 0)` makes the rate identically zero on the far side. Accepted steps over one 10-year run:

| drain limiter | accepted steps | wall |
|---|---|---|
| shipped: `max(S,0)` and `net>0 ?:` | 940 | 8 s |
| `r/(r + 1e-3)` | 12 865 | 114 s |
| `r` | **488** | **3 s** |

A change of state basis does not reach this: an eigenvalue at a fixed point is a property of the vector field, not of the chart, so `S = w^2` and `log S` leave it where it is. #610's rung 2 answers the shipped rate's shape — an exponential approach to a boundary with no interior fixed point — which a log coordinate does linearise into a drift.

## The family, priced on one trajectory

`min tau` is the shortest relaxation time over the run; `L` is the fraction of a deficit a plant at that reserve pays from reserves.

| `L(r)` | min `tau` | vs shipped | `L` at 0.01 / 0.1 / 0.5 |
|---|---|---|---|
| `r/(r+1e-3)` | 7.26e-05 | 0.03x | 0.909 / 0.990 / 0.998 |
| `r/(r+1e-2)` | 2.02e-04 | 0.09x | 0.500 / 0.909 / 0.980 |
| `r/(r+1e-1)` | 1.44e-03 | 0.66x | 0.091 / 0.500 / 0.833 |
| `1-(1-r)^8` | 1.70e-03 | 0.79x | 0.077 / 0.570 / 0.996 |
| `1-(1-r)^4` | 2.84e-03 | 1.31x | 0.039 / 0.344 / 0.938 |
| **`r`** | **5.50e-03** | **2.55x** | 0.010 / 0.100 / 0.500 |

`1-(1-r)^4` is the retreat if the shipped drawdown matters more than deleting the constant: it clears the step count and keeps most of the old shape, at the price of an exponent.

## Where the pool comes to rest

The fixed point solves `Ppos(1-G(r))(1-r) = drain*L(r)`, over the 155 records with `P < 0`:

| `L(r)` | rest `r`, median | max | mortality at rest | resting above 0.05 |
|---|---|---|---|---|
| `r/(r+1e-3)` | 0.0000 | 0.0022 | 5.51 | 0.000 |
| `r/(r+1e-1)` | 0.0000 | 0.0866 | 5.51 | 0.058 |
| `r` | 0.0000 | 0.2018 | 5.51 | 0.077 |

So a starving plant still rests on the boundary with mortality at its maximum. What moves is a tail of 12 records, 0.27 per cent of all records, and #620 shrinks that tail's worst case to 0.0349.

## The refusal is load-bearing

The proportional limiter puts the pool's relaxation on its own timescale, but the controller does not pace on the storage state: the largest accepted step is 12 to 25 times the shortest relaxation time across the fixtures. Those extremes need not co-occur, so it bounds the exposure rather than demonstrating an overshoot — but it is why the domain check stays rather than being dropped with the clamps.

## The bound, measured on the state

| lifetime | cohorts | min `r` | median `r` | max `r` | frac `r > 0.99` |
|---|---|---|---|---|---|
| 2 | 82 | +4.00e-01 | 0.400 | 0.800 | 0.000 |
| 10 | 97 | +4.93e-10 | 0.436 | 0.800 | 0.000 |
| 50 | 140 | +7.58e-02 | 0.401 | 0.800 | 0.000 |
| 105.32 | 177 | +2.28e-01 | 0.624 | 0.800 | 0.000 |

Against 1.035 of capacity and 52.7 per cent of a mature stand on the clip. Negative records go to none, on both coordinates.

## Validation

- **Gradient sweep**, on the AD branch this is ported from: 13 ladder files, 430 assertions, no failures — the block Jacobian forward against reverse, and every injected corruption still detected.
- **Suite**: TF24 75, TF24f 57, patch 145, scm 92, stochastic runner 120; 2740 pass overall.
- **Re-blessed**: four offspring-production values, and the stochastic runner's seeded TF24 count 81 → 79 with survivors unchanged at 3. FF16 and K93 are bit-unchanged, which is what confines the change to TF24's storage block.

Two test shapes changed. The kink check is now a convergence check, because production's smoothed positive part has a curvature of order `1/eps` there, so a one-sided stencil carries 8 per cent of its own error at any single width — the old rate reads 1.972 at every width, this one converges 1.0785, 1.0076, 1.00076, 1.000076. And a new test pins the relaxation time as a ratio against the pool's own rate, so it is a property of the form rather than a benchmark; it is the guard against restoring the narrow limiter.

## Dependencies

`stop_domain` and `ode_state_valid` need odelia >= 0.3.1. phylloptim pinned odelia at v0.2.1 in its own `Remotes`, which is two pinned refs for one package; traitecoevo/phylloptim#91 lifts it, and this branch takes phylloptim from master so the two resolve together.

dfalster:
## Rebased onto `develop`; the dependency block is gone

`develop` now carries #633, *Migrate to phylloptim 0.6.0 and odelia 0.3.1*, so `stop_domain` and `ode_state_valid` are available on the base branch and the third commit here — *Take phylloptim from master, so odelia can move to 0.3.1* — had nothing left to do. It is dropped: `DESCRIPTION` is now **identical to `develop`**, at `odelia@v0.3.1` and `phylloptim@83fbb25`, rather than the unpinned `traitecoevo/phylloptim` this branch was carrying. Two pinned refs became one pin, which is what traitecoevo/phylloptim#91 was for.

Three commits become two. The only conflict was `DESCRIPTION`, resolved to `develop`'s side on both hunks; `NEWS.md` and `src/tf24_strategy.cpp` merged without one, and #633's changes to the TF24 leaf calls are untouched — the storage block and the leaf interface do not overlap.

## Revalidated on the new base

Rebuilt from clean against installed odelia 0.3.1 and phylloptim 0.6.0, and the whole suite is green:

```
[ FAIL 0 | WARN 0 | SKIP 7 | PASS 2932 ]
```

2740 → 2932 is #633's own new tests arriving with the base, not new tests here. The 7 skips are the usual `on_cran` guards. **No re-blessing was needed on top of the rebase**: the four offspring-production values re-blessed in this branch, and the stochastic runner's seeded TF24 count of 79, all still hold against the migrated leaf. those fixtures assert at `tolerance = 2e-2`, and #633 moves TF24 output by +0.36% through `kg_to_mol_h2o` — inside that by a factor of five, where this branch moves the same numbers by 2.7x on height — so the two changes are separable in the fixtures as well as in the source.

Nothing in the analysis above changes: the rate, the bound, the step counts and the offspring-production movements were all measured on the storage block, which the migration does not touch.

Issue #126, opened 2015-02-17
Rename variables


dfalster:
With building of tree2, I suggest a mass renaming of variables to improve consistency and readability across the project. @richfitz already made one round of simplifications. Here's a proposed second round of changes, any objections?

s -> mass_seed

leaf_area -> area_leaf
leaf_mass -> mass_leaf
bark_area -> area_bark
sapwood_area -> area_sapwood
heartwood_area -> area_heartwood
stem_area -> area_stem
stem_diameter -> diameter_stem
mass_root -> root_mass
total_mass -> mass_total
live_mass -> mass_live
above_ground_mass -> mass_above_ground

leaf_area_deployment_mass -> darea_leaf_dlive_mass

reproduction_mass_fraction -> fraction_allocation_reproduction
growth_mass_fraction -> fraction_allocation_live_mass

dfecundity_dt -> reproduction_dt

net_mass_production -> net_mass_production_dt

dleaf_area_dt -> area_leaf_dt
dsapwood_area_dt -> area_sapwood_dt
dheartwood_area_dt -> area_heartwood_dt
dbark_area_dt -> area_bark_dt
dstem_area_dt -> area_stem_dt
dstem_diameter_dt -> diameter_stem_dt
droot_mass_dt -> mass_root_dt
dlive_mass_dt -> mass_live_dt
dtotal_mass_dt -> mass_total_dt
dabove_ground_mass_dt -> mass_above_ground_dt
dheartwood_mass_dt -> mass_heartwood_dt
# These will also be modified via above changes

dheight_dleaf_area
dsapwood_mass_dleaf_area
dbark_mass_dleaf_area
droot_mass_dleaf_area
dstem_diameter_dstem_area
# New variables

allocation_leaf
allocation_sapwood
allocation_bark
allocation_root

mass_production_dt
# Existing, won't be changed

assimilation_leaf
respiration_leaf, respiration_bark, ,...
turnover_leaf turnover_bark, ...

mortality_dt
mortality_growth_dependent_dt
mortality_growth_independent_dt

Issue #86, opened 2014-08-19
Port from modules to RcppR6


richfitz:
Motivation for this is to avoid the long load time, allow using devtools, and allow templated plant types.  This will also remove the virtual functions/inheritence in the low-level things, which will clean things up a little.

RcppR6 looks about ready: the main thing required will be default arguments as the current implementation does not support these, and will never support multiple constructors.  Converting things will take a little while though.  Here's an example (untested) switching the Interpolator class:

``` c++
  Rcpp::class_<interpolator::Interpolator>("Interpolator")
    .constructor()
    .constructor<bool,bool>()
    .property("type",     &interpolator::Interpolator::type)
    .method("init",       &interpolator::Interpolator::init)
    .method("eval",       &interpolator::Interpolator::r_eval)
    .method("deriv",      &interpolator::Interpolator::r_deriv)
    .property("x",        &interpolator::Interpolator::get_x)
    .property("y",        &interpolator::Interpolator::get_y)
    .property("xy",       &interpolator::Interpolator::r_get_xy)
    .property("size",     &interpolator::Interpolator::size)
    .property("min",      &interpolator::Interpolator::min)
    .property("max",      &interpolator::Interpolator::max)
    ;
```

``` yml
Interpolator:
  roxygen: @export
  name_cpp: interpolator::Interpolator
  constructor:
    # Need defaults to make this work nicely:
    args: [is_akima: bool=TRUE, is_linear: bool=TRUE]
  methods:
    init:
      args: [x: "const std::vector<double>&", 
        y: "const std::vector<double>&"]
      return_type: void
    eval: 
      name_cpp: r_eval
      args: [x: "std::vector<double>"]
      return_type: "std::vector<double>"
    deriv:
      name_cpp: r_deriv
      args: [x: "std::vector<double>"]
      return_type: "std::vector<double>"
  active:
    type: {type: std::string, access: method}
    x: {type: "std::vector<double>", access: member, name_cpp: get_x}
    y: {type: "std::vector<double>", access: member, name_cpp: get_y}
    xy: {type: "Rcpp::NumericMatrix", type: member, name_cpp: get_xy}
    size: {type: size_t, access: member}
    min: {type: double, access: member}
    max: {type: double, access: member}
```

Issue #84, opened 2014-08-12
Starting max fitness


richfitz:
Getting the community to compute maximum fitness before bounds will be useful, because if that is zero, then there is no point running an assembly.  However, given really wide lma values, we can get negative height plants because the root-found plant height is close to the root-find epsilon.

Better would be to start at the given point, and step until we "bracket" an optimum.  This is is going to be really hard to do in multiple dimensions though, but optim does not require bounds there.

PR #633, opened 2026-08-26
Migrate to phylloptim 0.6.0 and odelia 0.3.1

### LINKED ISSUE #622: Migrate to phylloptim 0.6.0's interface
phylloptim 0.6.0 (traitecoevo/phylloptim#126) changes `Leaf`'s constructor, its trait
vector and several of its method names. plant pins `traitecoevo/phylloptim@037673b7`,
which predates all of it — and predates `gradient.hpp` entirely — so this is a single
coordinated update rather than a drift.

**The good news first: the numbers do not have to move.** Verified below.

## The one thing that will silently break

`b` is not renamed to `stem_P50`. They are different quantities:

```
b = P50 / (ln 2)^(1/c)      so at plant's TF24 defaults, b = 2.5894 and P50 = 1.85
```

Passing plant's existing `b` into the `stem_P50` slot compiles, runs, and describes
a curve whose scale is **1.1465× too large**. Nothing warns. This is the item to get
right; the rest of this issue is mechanical.

## Why the conversion is exactly value-preserving

plant already parameterises its stem curve on P50. `tf24_strategy.h` has

```cpp
double p_50 = 1.85;
double c        = log(log(1-0.5)/log(1-0.88))/(log(p_50) - log(5.16));
double b        = p_50 / std::pow(-log(1 - 50.0/100.0), 1/c);   // == P50/(ln 2)^(1/c)
double psi_crit = b*std::pow(log(1/0.05), 1/c);                 // == P95
```

Those last two derivations are **exactly** the ones phylloptim now performs
internally, so handing over `(p_50, c)` and letting it derive `b` and `psi_crit`
reproduces plant's current numbers to the last bit. Checked:

| quantity | plant derives | phylloptim derives from `(P50, c)` | |
|---|---|---|---|
| `b` | 2.5894373183081751 | 2.5894373183081751 | identical |
| `psi_crit` | 7.0854928681315172 | 7.0854928681315172 | identical |

The root curve needs one conversion, because `root_b = 3.898245` is a literal rather
than derived: `root_P50 = root_b * (ln 2)^(1/root_c)` = 3.3999998934932538, which
gives back `root_b = 3.8982450000000002` — also identical.

## Constructor: 17 arguments to 15

`src/tf24_strategy.cpp:918` and the generated `src/RcppR6.cpp:12` both build it
positionally.

| plant passes | phylloptim 0.6.0 takes | note |
|---|---|---|
| `vcmax_25` | `vcmax_25` | |
| `c` | `stem_c` | rename |
| `b` | `stem_P50` | **convert** — pass `p_50`, not `b` |
| `psi_crit` | — | **gone**: derived as P95 of the stem curve |
| `root_c` | `root_c` | |
| `root_b` | `root_P50` | **convert**: `root_b * (ln 2)^(1/root_c)` |
| `root_psi_crit` | — | **gone**: derived |
| `beta2` | `TF24_beta2` | rename |
| `jmax_25`, `a`, `curv_fact_elec_trans`, `curv_fact_colim` | same | |
| `GSS_tol_abs`, `vulnerability_curve_ncontrol`, `ci_abs_tol`, `ci_niter` | same | |
| `g1_TF24` | `TF24_cost_scale` | rename |

plant's own `psi_crit` / `root_psi_crit` derivations become dead once they are no
longer passed. `b` likewise. Keep `p_50` and `c` — they are the parameters now.

Four traits are **not** constructor arguments and must be assigned afterwards if
plant wants to set them: `R_d_25`, `JS22_gamma`, `CMax_a`, `CMax_b`. Their defaults
leave the TF24 path unchanged, so plant can ignore all four for now — but note that
`leaf_traits(R_d_25 = )` being silently ignored is a real trap on the R side.

## Names plant's generated glue refers to and that no longer exist

`inst/RcppR6_classes.yml` lists 55 fields as `access: field`, and RcppR6 pastes
`name_cpp` verbatim after `->`, so these break the *generated* `src/RcppR6.cpp`
rather than only plant's own sources.

| plant binds | do this |
|---|---|
| `lambda_` | `name_cpp: "CF77_lambda_"` |
| `lambda_analytical_` | delete the binding; nothing read it |
| `optimise_psi_stem_TF` | `set_model("TF24", "stem")` then `optimise()` |
| `optimise_psi_stem_Sperry` | `set_model("ProfitMax", "stem")` then `optimise()` |
| `hydraulic_cost_Sperry` | delete, or bind `hydraulic_cost_norm_` |
| `profit_psi_stem_Sperry` | `profit_psi_stem_ProfitMax` |

`perturb_stem_b` is now `perturb_stem_P50`; plant does not use it.

Those are the only six R-bound names that disappeared. Everything else plant binds
survives, `find_root_collar_psi()`, `evaluate_root_collar_psi()` and
`dprofit_droot_collar_psi()` among them — so `tf24_strategy.cpp` and
`tf24f_strategy.cpp` need no change beyond the constructor, and
`optimise_psi_stem_TF` at `tf24_strategy.cpp:45` is a comment.

For `access: field` the R-side name comes from the YAML key rather than from
`name_cpp`, so plant's R API can stay put:

```yaml
lambda_: {type: double, access: field, name_cpp: "CF77_lambda_"}
```

Whether plant's *R-side* names should follow the model-scoped convention is a
separate decision and a plant API break in its own right. Not this change.

## Also new, and optional

- **`gradient.hpp`** did not exist at plant's pin. `gradient::batch()` gives exact
  trait gradients over many observations in one call, and `gradient::n_pars` is 18.
  Nothing in plant needs it yet; traitecoevo/plant#614 is where it would land.
- **Six more cost curves** (`CF77`, `JS22`, `CMax`, `SOX`, `JW26`, `ProfitMax`) reachable
  through `set_model(curve, route)`. The default is `("TF24", "collar")`, which is
  plant's path, so plant gets them for free and uses none of them.
- **`weibull_p50_c()`** and friends convert published vulnerability curves into
  `(P50, c)`, which is what plant's hyper-parameterisation is doing by hand.

## Suggested order

1. Bump the `Remotes:` pin to phylloptim `master` after #126 merges.
2. Fix the constructor in `tf24_strategy.cpp` — pass `p_50` and `c`, drop
   `psi_crit`/`root_psi_crit`, convert `root_b`.
3. Fix the six names in `inst/RcppR6_classes.yml`, then regenerate:
   `Rscript -e 'RcppR6::RcppR6()'`.
4. Update `tests/testthat/test-leaf.r`: 23 references across `lambda_`,
   `lambda_analytical_`, `hydraulic_cost_Sperry` and `optimise_psi_stem_TF`. Most are
   `expect_true(is.na(l$lambda_))` assertions that a freshly built leaf has not
   solved yet, so they carry over unchanged once the binding is renamed.
5. `rm -f src/*.o src/*.so` before reinstalling. R does not track header
   dependencies, so a stale object silently runs the old model.
6. **Confirm with the SCM regression rather than assuming.** The conversion above is
   bit-identical at the defaults, so the expectation is that nothing moves at all.
   A moved SCM node means the conversion was applied wrongly — most likely `b` into
   the `stem_P50` slot.

dfalster:
Leaf() takes p_50 and root_p_50; phylloptim derives b and psi_crit from them,
reproducing plant's curves bit for bit. set_model() + optimise() replace the
optimise_psi_stem_* calls, and the collar helpers are templated on the curve.

TF24 output moves +0.36%, bisected to one upstream constant: kg_to_mol_h2o,
55.4939 -> 1/molar_mass_h2o = 55.509298. FF16 and K93 are identical().

scientific_version deliberately not bumped. Closes #622.


dfalster:
## The migration, and the one thing #622 did not predict

### #622's conversion is exactly right

Handing over `(p_50, c)` and letting phylloptim derive the rest reproduces plant's curves bit for bit:

| | plant derives | phylloptim derives from `(P50, c)` |
|---|---|---|
| `stem_b` | 2.5894373183081751 | 2.5894373183081751 |
| `psi_crit` | 7.0854928681315172 | 7.0854928681315172 |
| `root_b` | 3.8982450000000002 | 3.8982450000000002 |
| `root_psi_crit` | 5.8702825428827037 | 5.8702825428827037 |

So the `b` vs `stem_P50` trap #622 warns about is avoided, and the root conversion `root_P50 = root_b·(ln 2)^(1/root_c)` round-trips.

### But TF24's numbers do move, and #622's "the numbers do not have to move" is wrong

Reference SCM run: net reproduction **3.671011e-25 → 3.684168e-25, +0.36%**; 1055 → 1065 ODE steps; first divergence at step 50, t ≈ 0.011 yr. FF16 and K93 are `identical()`.

Bisected rather than guessed:

| change | TF24 |
|---|---|
| odelia 0.2.1 → 0.3.1, plant untouched | identical |
| phylloptim 037673b7 → e8f4a7f (16 commits, incl. #84 and #86), plant untouched | identical |
| phylloptim → 0.6.0 / #126 | **moves** |

Then localised with a leaf-level probe at plant's TF24 configuration:

| quantity | 0.2.0 → 0.6.0 |
|---|---|
| vulnerability curve over ψ 0–6 | **max delta = 0** — #126's `vulnerability.hpp` rewrite is behaviour-preserving |
| `opt_root_psi`, `transpiration_`, `hydraulic_cost_`, `R_d_` | identical |
| `stom_cond_CO2_` | **rel 2.7747e-04** |
| `assim_colimited_`, `profit_` | follow, with the same absolute delta as each other |

The cause is one constant:

```cpp
- inline constexpr double kg_to_mol_h2o = 55.4939;
+ inline constexpr double molar_mass_h2o = 0.018015;             // kg mol^-1
+ inline constexpr double kg_to_mol_h2o  = 1.0 / molar_mass_h2o; // 55.509298
```

`55.509298 / 55.4939 − 1 = 2.7747e-04`, matching the measured `stom_cond_CO2_` shift **to five significant figures**. It enters the E → g_sc conversion; conductance carries it exactly, ci and assimilation follow, and the demographic integration amplifies it to 0.36%.

**This looks like a correction, not a regression.** The old 55.4939 implies a molar mass of 0.0180200 kg/mol; the new one implies 0.018015, which is water's actual value.

Two near-misses worth recording, because both would have been invisible:

- `R_d` is now a trait with default `R_d_25 = 1.44` rather than `0.015·vcmax_25`. plant's `vcmax_25` is **96**, and `0.015 × 96 = 1.44` exactly — so plant is unaffected *by coincidence*. Any strategy with a different `vcmax_25` would silently change respiration.
- `ko_`/`kc_` shift in the 5th significant figure off the 25 °C reference. plant pins its leaf at 25 °C, where the Arrhenius factors are 1, so nothing moves — but the leaf tests that deliberately go hot do.

### Three places #622's spec had gone stale

#126 grew after the issue was written, so:

- `optimise_psi_stem_TF` is **not** kept as a C++ alias. It is gone; `set_model("TF24", "stem")` + `optimise()` replaces it, and both are newly bound.
- `prepare_collar_solve` and `profit_at_collar_psi` are now **templated on the cost curve**, so `tf24f_strategy.cpp` needed changing after all (`<Leaf::CostCurve::TF24>`).
- `hydraulic_cost_Sperry(psi_stem, psi_upstream)` has no successor of the same shape — `hydraulic_cost_norm_` is a normalised *field* written by the ProfitMax solve. Four assertions were dropped, with the reason recorded in place; the `hydraulic_cost_TF` assertions beside them cover the same property for the curve plant runs.

### Test changes, and why each moved

Not blanket-updated. Three of them were **stale references**, not shifted numbers:

- `test-leaf.r` hand-rolls an R replica of the E ↔ ψ conversion with `kg_to_mol_h2o = 55.4939` **hard-coded**. Left alone it would have biased the round-trip by exactly the 2.77e-4 under investigation. Updated to `1/0.018015`.
- Two assertions restated formulas that are no longer true — `psi_crit` as a supplied constant, and `R_d = 0.015·vcmax_25`. Both now read the value off the leaf, so they are identities again rather than duplicated arithmetic.
- The out-of-domain message is now phylloptim's own, naming the spline and the overshoot rather than odelia's generic text.

The remaining six are genuine pinned values, updated with the reason in a comment beside them.

`psi_crit` and `stem_b` are newly exposed on plant's `Leaf`, because they are derived now and R has no other way to see the curve the leaf is using.

### Verification

Full suite `FAIL 0 | WARN 0 | SKIP 4 | PASS 2923`. The FF16/K93 `identical()` check and the TF24 comparison both run against baselines captured on `develop` under the old pins, before any code change.

`LinkingTo phylloptim (>= 0.6.0)` earned its keep during this work: it stopped a build dead when the bisection had left 0.2.0 installed, instead of failing deep in a template instantiation.

### Two open items

- **`scientific_version` not bumped**, by decision. Archived TF24 results from before this PR differ by ~0.36% but carry the same `TF24@v8`. `logpile` has not been used yet, so nothing is currently mis-served.
- **`agents.md` said `TF24@v2`; it is `v8`.** Six bumps behind. Replaced with a pointer to `model_id()` so it cannot drift again.

PR #635, opened 2026-08-26
Seat TF24 on the TF24_floor cost curve

### LINKED ISSUE #634: [TF24 hydraulics] Move the TF24 strategies onto the TF24_floor cost curve, with the water price at zero by default
## Why

Every conductance-loss cost prices water at **zero** as transpiration goes to zero — no conductivity is lost when nothing flows — so water is free precisely when it is abundant. `TF24_floor` is the repair: it splits the cost into a part depending on potential alone and a part linear in the flux,

```
Theta(E) = Theta~(psi) + lambda_o * E
```

with `Theta~` being TF24's own cost at TF24's own traits. So `lambda_o` is the curve's only parameter and **`TF24` is `TF24_floor` at `lambda_o = 0`, at identical parameter values.**

Nothing needs implementing here. plant already consumes phylloptim through `LinkingTo` and `plant::Leaf` *is* `phylloptim::Leaf` — see the shim in `inst/include/plant/leaf_model.h`. The curve and the new `shadow_cost` output arrive with a version bump. What this issue asks for is that plant **seat** the curve and, more importantly, get the accounting right.

## ⚠️ Step 0: pin the phylloptim version, don't just raise the floor

`DESCRIPTION:22` currently reads `phylloptim (>= 0.6.0)`, and 0.6.0 has neither the curve nor `shadow_cost` — both landed in **phylloptim 0.7.0** (traitecoevo/phylloptim#132). Because plant consumes this package as *headers*, an unraised bound fails as a compile error at best and a silently absent field at worst.

But raising it to `>= 0.7.0` is not enough, and the reason is specific to this dependency:

```
LinkingTo: Rcpp, BH, odelia (== 0.3.1), phylloptim (== 0.7.0)
```

**plant's regression baselines are bit-exact, and phylloptim is compiled into plant.** Every number in plant's golden files is a property of the phylloptim headers that were in the include path when plant was built. Under `>=`, installing any later phylloptim silently changes plant's compiled arithmetic — a re-parameterisation, a reordered expression, a new default — and the first sign is a red baseline with no local change to explain it. That is a bisect through two repositories.

Under `==` the same upgrade fails at install time, naming the version, before anything is built. The upgrade then becomes a deliberate act: bump the pin, rebuild, read the baseline diff, and record it — which is what the golden files exist to make possible. A version range is the right default for a dependency you call; it is the wrong one for a dependency you *compile in* and hold to bit-equality.

The same argument applies to `odelia`, which is also `LinkingTo` and also header-only, so it is worth pinning in the same pass.

## The safety property this rests on

`TF24_floor` at `lambda_o = 0` reproduces `TF24` **bit-for-bit** — asserted in phylloptim over 45 driver rows and eight reported fields, with no fixture and no tolerance, because each term is the parent's own expression so zeroing the price adds an exact zero to TF24's exact value. It survives fused multiply-add for the same reason.

So with the price defaulting to zero, **every existing plant result should be unchanged to the last bit**, and that is the acceptance test for the first half of this work. If a regression baseline moves, the change is wrong — do not update the baseline.

## ⚠️ The part that is easy to get wrong: the shadow price must not reduce growth

This is the substance of the issue, and it is worth reading before touching anything.

The two terms are **not the same kind of thing.** `Theta~` is a *realised* cost — carbon actually forgone, in damaged tissue and in capacity that has to be rebuilt — and belongs in a carbon budget. `lambda_o` is **not a cost at all.** It is a *shadow price*: the value of water in its best alternative use, which for a leaf is assimilation later. No carbon is lost when the plant pays it. It changes the aperture the plant chooses, and nothing else — which is what a Lagrange multiplier does.

`leaf.profit_` is the **objective**, so it deducts both terms. plant currently feeds it straight into growth:

https://github.com/traitecoevo/plant/blob/master/src/tf24_strategy.cpp#L552

```cpp
const double assimilation_ = leaf.profit_ * area_leaf_* 60*60*12*365/1e6;
```

That is correct today, because at `lambda_o = 0` the objective *is* the realised carbon. It stops being correct the moment the price is non-zero: growth would be taxed by carbon the plant never spent, and the strategy would be double-counted as a physiological loss. At `lambda_o = 1e5` the objective understates the carbon the leaf actually kept by **28%**.

The quantity plant wants is `profit + shadow_cost`. phylloptim reports `shadow_cost` as of 0.7.0 — `lambda_o * E` on this curve, exactly zero on all seven others.

### The canopy integral makes this easier than it looks

`shadow_cost()` reads the *stored* `transpiration_`, and plant overwrites both `profit_` and `transpiration_` with canopy integrals taken against the **same** weights `qi` (`src/tf24_strategy.cpp:520-521`, assigned at `:535-536`). The shadow term is linear in `E`, so

```
∫ profit·q dz  +  lambda_o ∫ E·q dz  =  ∫ (A - Theta~)·q dz
```

exactly. No second integration is needed — after the integral is assigned, `leaf.shadow_cost()` already returns `lambda_o` times the integrated transpiration. So the fix at `:552` is:

```cpp
// The carbon the plant actually kept. `profit_` is the OBJECTIVE and deducts the
// shadow price too; that price is not carbon and must not reduce growth.
const double carbon_profit_ = leaf.profit_ + leaf.shadow_cost();
const double assimilation_  = carbon_profit_ * area_leaf_ * 60*60*12*365/1e6;
```

At `lambda_o = 0` this is `leaf.profit_ + 0.0`, which is `leaf.profit_` exactly — so the line can be changed *before* the curve is switched, with no effect on any baseline. That is the safer ordering: land the accounting fix first, on its own, where it provably changes nothing.

## Seating the curve

⚠️ **`leaf.find_root_collar_psi()` is hard-wired to TF24 and does not read the seated curve.** It is a fixed shortcut:

```cpp
inline void Leaf::find_root_collar_psi() {
  find_root_collar_psi_for<CostCurve::TF24>();
}
```

So `set_model("TF24_floor", "collar")` alone will do nothing at `TF24_Strategy::solve_leaf()` (`src/tf24_strategy.cpp:577`) — it will keep solving TF24 while reporting a `shadow_cost` computed from a `lambda_o` the solve never saw, which is worse than not switching. Use `leaf.optimise()` instead, which dispatches on the seated curve and route through `with_curve` and covers all eight curves on both routes.

Swapping `find_root_collar_psi()` for `optimise()` with TF24 seated is the same code path plus one `bool` write that no baseline reads, so that step is also bit-neutral and can be verified on its own.

`src/tf24f_strategy.cpp` selects the curve at **compile time** in four places (`:108`, `:120`, `:121`, `:123`, all `<Leaf::CostCurve::TF24>`). Those need the same treatment — either templated on the seated curve via `with_curve`, or switched wholesale.

## Suggested order of work

Each step is separately verifiable, and steps 1–4 should move no numbers at all.

0. Pin `phylloptim (== 0.7.0)` in `DESCRIPTION` (and `odelia` likewise), and rebuild against it. Nothing below compiles without this.
1. Change `:552` to `profit_ + shadow_cost()`. Bit-neutral while the curve is TF24, since `shadow_cost()` is then exactly `0.0`.
2. Replace `find_root_collar_psi()` with `optimise()` in `solve_leaf()`, seating `("TF24", "collar")`. Bit-neutral.
3. Do the same for `tf24f_strategy.cpp`'s four template sites. Bit-neutral.
4. Add a strategy parameter — `TF24_floor_lambda_o`, in `umol C (kg H2O)^-1`, **defaulting to 0** — and seat `"TF24_floor"`. Bit-neutral at the default, by phylloptim's reduction assertion.
5. Only then, exercise a non-zero price.

⚠️ On step 4: phylloptim refuses `TF24_floor_lambda_o` **with no default**, deliberately — at zero the curve *is* TF24, so a caller who seated it and never named a price would be running the production model under a new name. An explicit zero is accepted. So plant's parameter defaults to 0 and must be passed explicitly, not omitted.

## Also worth doing

`src/tf24_strategy.cpp:173` records `profit` as an aux. That is fine as the objective, but consider adding `shadow_cost` beside it — otherwise the two quantities cannot be separated after the fact from plant's own output, and the distinction is the whole point of the curve.

## One thing that is *not* a bug

`shadow_cost` is exactly zero on every curve other than `TF24_floor`, and that means **"this curve does not separate the two"**, not "this curve's cost is all realised carbon". `CF77` is the case worth knowing: its whole cost is `lambda * E`, and reading that as a shadow price is an ordinary thing to do — but the model supplies one number and nothing to attribute it with, and behaves identically whether it is read as carbon lost to water already taken or as the value of water withheld for later. Reporting it as all-shadow would present one reading as a fact.

## Verification

- plant's existing regression baselines, unchanged through steps 1–4. A moved baseline means a mistake, not a new expected value.
- A non-zero `lambda_o` should *reduce* conductance and transpiration and *raise* the implied `g1` at low light, while leaving `carbon_profit >= profit`.
- `carbon_profit == profit` exactly when `lambda_o == 0`.

## References

- traitecoevo/phylloptim#132 — the curve, the reductions, and `shadow_cost`, with the reasoning behind the realised/shadow split. Released as phylloptim 0.7.0.
- The conceptual argument is Isaac Towers': the hydraulic term is a realised carbon cost, the floor is a strategic trait that makes the plant weigh tomorrow's assimilation against today's and carries no lost profit at all. Read that way `lambda_o` indexes a continuum between carbon maximisation at zero and water-conserving strategies above it — which is a prediction about aridity rather than a fitted constant.

dfalster:
TF24's hydraulic cost prices water at zero as transpiration goes to
zero, so water is free precisely when it is abundant. phylloptim's
TF24_floor adds a price that does not vanish, and TF24 is that curve
at lambda_o = 0 at identical parameter values. New parameter
TF24_floor_lambda_o defaults to 0, so nothing moves: the whole suite
passes with every pinned baseline unchanged.

The price is a shadow price, not a cost, so growth is now billed on
profit + shadow_cost -- a new aux reported beside profit. Every
collar-solve call site routes through the seated curve instead of
naming TF24. TF24's leaf parameters take phylloptim's names, and both
LinkingTo dependencies are pinned with ==.

closes #634


dfalster:
Closes #634. Detail, measurements and the two things the issue got slightly differently.

## Order of work, and what each step was verified against

Steps 1–4 of the issue's plan should move no numbers, and they don't. The suite went **2954 pass / 0 fail** with only the accounting fix and the curve seated (the 2 reds at that point were the model-version snapshot picking up the new parameter, nothing numeric), and **2974 / 0** with the 18 new tests and the rename.

| step | what | verified by |
|---|---|---|
| 0 | pin `phylloptim (== 0.7.0)`, `odelia (== 0.3.1)` | `R CMD INSTALL` clean; and the pin *bites* — see below |
| 1 | `:552` → `profit_ + shadow_cost()` | bit-neutral: `shadow_cost()` is exactly `0.0` at the default |
| 2 | `find_root_collar_psi()` → `optimise()` | bit-neutral: same path plus one `bool` write |
| 3 | TF24f's four template sites → `with_curve` | bit-neutral |
| 4 | `TF24_floor_lambda_o`, default 0, curve seated | every pinned baseline unchanged |
| 5 | non-zero price exercised | new `test-tf24-floor-price.R`, 18 checks |

## ⚠️ The reduction assertion the issue cites is on the STEM route, not the collar one

phylloptim's `test_tf24_floor_reduces` calls `set_model(..., false)` — the second argument is `collar_route`, so those 45 rows are the **stem** route. plant runs the **collar** route, and the two are not the same code on this point: `no_flow_profit<TF24>` is spelled out as `-R_d_ - hydraulic_cost_TF(psi)` while every other curve goes through `profit_psi_stem_for<K>(psi, psi)`. So the feasibility exits are a *different expression* on the two curves and the reduction is not automatic there.

Measured before touching plant, with a standalone C++ probe against the phylloptim headers:

```
378 collar-route rows (108 with E == 0), 0 rows differ, 0 values differ
 96 TF24f-entry rows, 0 differ
```

Grid: PPFD {100, 900, 1500} × VPD {1, 2, 4} × T {25, 40} × ψ_soil {0.5, 1, 2, 4, 5.5, 6.5, 8} × {1, 3, 5} layers, nine reported fields, compared with `memcmp`. The ψ_soil values deliberately run past `root_psi_crit` (5.87 MPa) so the shutdown exits are exercised rather than assumed — 108 of the 378 rows have `E == 0`. The second block covers `evaluate_root_collar_psi_for<K>`, `dprofit_droot_collar_psi_for<K>`, `prepare_collar_solve<K>` and `profit_at_collar_psi<K>`, i.e. the three TF24f entry points, none of which the first block reaches.

So the collar-route reduction holds too. Worth recording in phylloptim as a second arm of that test, since nothing there covers it.

## The `==` pin bites, and here is what it looks like

The load-bearing claim in the NEWS entry, so it was checked rather than asserted. With `phylloptim (== 0.7.1)` against the installed 0.7.0:

```
Error: package ‘phylloptim’ 0.7.0 was found, but == 0.7.1 is required by ‘.’
```

At install time, naming the version, before anything is built — which is the whole point over a red baseline with no local change to explain it.

## The shadow price, measured on plant

5 m plant, PPFD 1800, θ = 0.25 across 5 layers:

| `lambda_o` | `profit` | `shadow_cost` | carbon kept | objective understates by | E | g_sc |
|---|---|---|---|---|---|---|
| 0 | 11.13207 | 0.00000 | 11.13207 | 0.00% | 2.71e-05 | 0.0914 |
| 1e4 | 10.85611 | 0.25555 | 11.11166 | 2.30% | 2.56e-05 | 0.0860 |
| 5e4 | 9.67099 | 1.00629 | 10.67728 | 9.42% | 2.01e-05 | 0.0678 |
| 1e5 | 8.00996 | 1.46214 | 9.47211 | **15.44%** | 1.46e-05 | 0.0492 |
| 2e5 | 4.07738 | 1.18872 | 5.26610 | 22.57% | 5.94e-06 | 0.0200 |

The issue quotes 28% at `lambda_o = 1e5`; on plant's own defaults at this operating point it is **15.4%**, rising to 22.6% by 2e5. Different leaf, same conclusion — the error is the same order as the effect being modelled, which is why growing on the objective is not a rounding concern.

Conductance and transpiration fall monotonically and `psi_stem` gets wetter, as the issue predicted.

## `shadow_cost` needed no second canopy integral

Confirmed as the issue argued: `shadow_cost()` reads the *stored* `transpiration_`, and the DeepCrown branch overwrites `profit_` and `transpiration_` with integrals against the **same** weights `qi`, so linearity in E makes `∫profit·q + λ_o∫E·q = ∫(A − Θ̃)·q` exact. There is a comment at the assignment saying that moving either integral apart from the other breaks it.

## TF24f: one dispatch for the whole body, not seven

The issue names four template sites in `tf24f_strategy.cpp`. There are seven once the AD branch is counted: `find_root_collar_psi()` in the `initializing_` branch, and `dprofit_droot_collar_psi(used)` — both TF24-hardwired *wrappers* rather than explicit `<CostCurve::TF24>`, so they don't show up in a grep for the template argument.

Rather than convert seven call sites, `solve_leaf()` now resolves the seated curve **once** and hands the entire body to `solve_leaf_for<K>`. That makes it structurally impossible for two calls inside one solve to disagree about which model is being solved, which was the actual failure mode: before this, `set_model()` had no effect on TF24f at all, so seating TF24_floor would have moved `shadow_cost` while the tracked state went on climbing TF24's gradient.

`test-tf24-floor-price.R`'s last test is the guard. It cannot use E or g_sc as the instrument — TF24f *evaluates* at the tracked collar rather than optimising it, so at a given tracked state those are curve-independent by construction. It checks `shadow_cost` and the tracked state's **rate** instead, and seeds the tracked state from TF24's optimum, because a fresh individual holds 0 (a collar at saturation) where no water moves and every price is worth nothing times zero.

## The parameter rename

Not in the issue; requested during review. `TF24_Pars`: `p_50 → stem_P50`, `c → stem_c`, `b → stem_b`, `beta2 → TF24_beta2`, `g1_TF24 → TF24_cost_scale`, and `Leaf()`'s arguments move with them (`root_p_50 → root_P50`).

The reason it matters rather than being cosmetic: these are handed to `Leaf()` **positionally**, so a name that disagrees with the slot it lands in is silent. `g1_TF24` was the old name for what phylloptim calls `TF24_cost_scale` — the TF24 hydraulic cost scale — and it reads as a stomatal slope, a quantity the leaf separately has under `g1` on the Medlyn path.

**No value moved.** Checked directly rather than by eye: the old and new model-version snapshots were parsed and compared as name→value maps under the rename table —

```
TF24   added: pars.TF24_floor_lambda_o=0 | dropped: none | values moved: NONE
TF24f  added: pars.TF24_floor_lambda_o=0 | dropped: none | values moved: NONE
```

Two things deliberately **not** renamed:

- **`root_b` stays `root_b`.** It is the Weibull *scale*; `root_P50 = root_b·(ln 2)^(1/root_c)` is the 50%-loss potential. Different quantities, so there is nothing to align — phylloptim's own `root_b` and `root_psi_crit` already carry these names, and what differs is that it makes them derived and read-only while taking `root_P50` as the settable trait. Which of the pair plant should expose is an interface question of its own and would change a default *value*, not a name. Worth its own issue.
- **`beta1`** (20000) is declared, exposed to R, and read by nothing in C++. Left alone here; separate cleanup.

`stem_b` and `psi_crit` are worth knowing about: `TF24_hyperpar` computes both, `prepare_strategy()` passes neither, and phylloptim derives its own from `(stem_P50, stem_c)` by the same formulas. They are reporting copies, not inputs — setting either does not change the curve. That is now said at the struct.

## Not bumped: `scientific_version`

Stays at `TF24@v9` / `TF24f@v9.1`. No equation changes and no default output changes, which is the stated bump rule. The model-version snapshot was accepted rather than the version bumped, on the strength of the values-moved check above.

The one thing a reviewer might disagree with: `aux_size` goes 11 → 12, so the aux *vector* changes shape even though no number in it moves. Stored outputs read positionally would shift. Under the documented rule that is not a bump; say so if you want it treated as one.

## Also

- `notes/plan-tf24-scenario-framework.md` carries a marker that its parameter names are pre-#634 — it is the record of #548, not a live plan.
- Two conventions added to `agents.md` §11: the curve is seated once in `prepare_strategy()` and no call site may name one; and the leaf parameters must keep phylloptim's names, with the positional-hand-over reason.
- `notes/penman-monteith/step5_fick_vs_pm.R` was left alone. It calls `Leaf()` with `b =` and `root_psi_crit =`, arguments that stopped existing in #622 — it is already broken against the current API and half-updating it would hide that.

dfalster:
Fyi @itowers1 @elijahmagistrado -- the new stomatal model with shadow price on water now live and the default in plant. It's set up so that the default shadow price is zero, that way the model behaves similarly to what it did before, but you can increase it as a trait.

dfalster:
## The scenario gateway fails, and it is not this PR — measured, not argued

`PLANT_RUN_SCENARIOS=1` reports `persists S01: TRUE -> FALSE` against the recorded baseline. **Unmodified `develop` at `2c270be1` reports byte-for-byte the same failure**, so the flip is pre-existing drift.

Built and installed `origin/develop` in a separate worktree (`git stash` cannot revert your own commits) and ran the same gateway:

```
FAILURE: 'test-scenario-gateway.R:55:5'
Scenario outcomes changed vs baseline: persists S01: TRUE -> FALSE
[ FAIL 1 | WARN 0 | SKIP 0 | PASS 1 ]
```

Then the stronger statement, because matching *classifications* is weaker than matching *numbers*. Full scorecards from both builds, via `scripts/run_scenario_gateway.R`:

| scenario | `develop` | this branch | identical |
|---|---|---|---|
| S01 | 0.47977632012566557 | 0.47977632012566557 | ✓ |
| S02 | 0.011533148189433219 | 0.011533148189433219 | ✓ |
| S03 | 1.6654162185594561e-10 | 1.6654162185594561e-10 | ✓ |
| S04 | 1.3178150166448428e-11 | 1.3178150166448428e-11 | ✓ |
| S05 | 2.4700083582999845e-11 | 2.4700083582999845e-11 | ✓ |
| S06 | 6.5700725263637376e-12 | 6.5700725263637376e-12 | ✓ |
| S07 | 810.70501840763779 | 810.70501840763779 | ✓ |
| S08 | 2.0255780484570568e-14 | 2.0255780484570568e-14 | ✓ |

`offspring_production` is **bit-identical on all 8 scenarios**, and every classification column (`observed`, `match`, `outcome`, `crashed`, `persists`, `finite`, `error_message`, `warnings`) differs on **no** row. The only numeric column that differs at all is `run_seconds` — wall clock, worst 4.7%.

So this PR is bit-neutral through full SCM runs as well as through the unit suite, which is the strongest form of the claim the issue asked for.

### Where the drift comes from, and why I did not re-bless

The baseline was last blessed at `133499b8` ("Score the gateway on two axes, in birth date", #583). Seven commits have landed on `develop` since, and one of them is `b374ce3b` ("Migrate to phylloptim 0.6.0 and odelia 0.3.1", #633) whose own NEWS entry records **TF24 output moving +0.36%** with `scientific_version` deliberately not bumped — and no re-bless. S01 is the scenario both `test-scenario-gateway.R` and `scripts/run_scenario_gateway.R` single out by name as sitting on the `R0 = 1` threshold ("switching the density coordinate ... moves S01 across R0 = 1"). Its R0 is 0.4798 now, so it is comfortably below 1 today.

`make bless-scenarios` records a changed outcome as expected. The drift is not this PR's, and folding it in would bury an unexamined TF24 movement inside a change whose entire claim is that it moves nothing. **Worth its own issue** against #633's drift.

## The price works as a trait (second commit)

Asked during review. It does, and it is now pinned — `trait_matrix()` -> `generate_strategy()` -> `pars[trait_names] <- xi`, so no extra wiring was needed. Verified and tested:

| route | result |
|---|---|
| `trait_matrix(c(0, 5e4, 1e5), "TF24_floor_lambda_o")` | 3 strategies at those prices |
| the leaf actually solves at it | `shadow_cost` 0 / 1.087 / 1.624; E falls 2.76e-05 -> 2.18e-05 -> 1.62e-05 |
| `add_strategies` across species | 2 species, distinct λ_o |
| `TF24_hyperpar` | passes through untouched |
| TF24f | works (shares `TF24_Pars`) |

⚠️ **Being a `TF24_Pars` field does not by itself make a parameter reachable as a trait**, which is why this is tested rather than assumed: `TF24_hyperpar` refuses to overwrite anything it derives ("Attempt to overwrite generated parameters"), and that is exactly what makes `stem_P50`, `stem_c`, `stem_b`, `psi_crit` and `TF24_cost_scale` unusable as *input* traits. The price is derived from nothing, so it survives. If someone later adds a `lambda_o ~ aridity` relationship to the hyperpar, the new test is what will tell them they have just made it unsettable.

The leaf-side half is tested separately from the strategy-side half on purpose: a value can land on `pars` and never reach `leaf.TF24_floor_lambda_o`, and that looks entirely fine from R.

Suite now **2985 pass / 0 fail**.

## One thing left open, deliberately

**A misspelt trait name is accepted silently.**

```r
generate_strategy(p0, trait_matrix(1e5, "TF24_floor_lambda"))
# -> accepted; lambda_o still 0, full run at the default
```

`generate_strategy` does `pars[trait_names] <- xi` and an unknown name on an RcppR6 list-class assign is dropped rather than refused. Pre-existing, and it affects every model and every trait — but this PR renames five TF24 parameters, so anyone migrating `g1_TF24 -> TF24_cost_scale` who mistypes gets a plausible run at the old default with no warning. It is the same failure mode phylloptim already has on record for a renamed trait.

A guard in `generate_strategy` (refuse a trait name not present in `pars`) closes it in ~3 lines. **Left out of this PR**: it changes shared behaviour for FF16 and K93 too, and I cannot rule out existing code that passes extra columns through. Happy to add it here or file it — say which.

dfalster:
## Pin moved to phylloptim 0.8.0

`LinkingTo: phylloptim (== 0.8.0)`, `Remotes: traitecoevo/phylloptim@845390c`.

This is the pin doing its job rather than getting in the way. Before the bump, building against the newly installed 0.8.0 failed at install time:

```
Error: package ‘phylloptim’ 0.8.0 was found, but == 0.7.0 is required by ‘.’
```

Which is the whole argument for `==` over `>=`, demonstrated: under `>=` this upgrade would have been silent and the first sign of any movement would have been a red baseline with no local change to explain it.

## Inert for plant, and by construction rather than by luck

phylloptim 0.8.0 (traitecoevo/phylloptim#133) reports the seated curve's `lambda_emergent` through `leaf_solve()`, and fixes a stale `lambda_emergent_` after the two terminal exits on a reused `Leaf`. The header diff plant compiles is **24 lines**, and all of it is:

1. one element appended to `operating_point_values()` — after `shadow_cost`, so nothing shifts position;
2. two `lambda_emergent_ = util::na_value;` writes, in `set_shutdown_state` and the `assim_max_ < 0` exit.

`git grep` over `src/`, `inst/`, `R/` and `tests/`: plant reads **neither** `lambda_emergent` nor `operating_point_values`. So nothing plant touches changed. Measured regardless, because that reasoning is exactly what the pin exists not to have to trust:

- Suite **2985 pass / 0 fail** — same count as on 0.7.0.
- Scenario scorecard, same plant source on 0.7.0 vs 0.8.0 — a clean A/B on phylloptim alone:

| scenario | phylloptim 0.7.0 | phylloptim 0.8.0 | identical |
|---|---|---|---|
| S01 | 0.47977632012566557 | 0.47977632012566557 | ✓ |
| S02 | 0.011533148189433219 | 0.011533148189433219 | ✓ |
| S03 | 1.6654162185594561e-10 | 1.6654162185594561e-10 | ✓ |
| S04 | 1.3178150166448428e-11 | 1.3178150166448428e-11 | ✓ |
| S05 | 2.4700083582999845e-11 | 2.4700083582999845e-11 | ✓ |
| S06 | 6.5700725263637376e-12 | 6.5700725263637376e-12 | ✓ |
| S07 | 810.70501840763779 | 810.70501840763779 | ✓ |
| S08 | 2.0255780484570568e-14 | 2.0255780484570568e-14 | ✓ |

`offspring_production` bit-identical on every row; every classification column unchanged.

Installed from `git archive` of the pinned sha rather than from the sibling working tree, so the build is the commit named in `Remotes:` and not whatever branch happened to be checked out.

## ⚠️ One thing 0.8.0 makes available that this PR does not take up

`lambda_emergent` is the **seated curve's** own `(dC/dpsi)/(dE/dpsi)`. `marginal_cost_water()` — reported as `lambda` — is *TF24's* price whatever curve actually ran, so on `TF24_floor` the two differ by the price floor. From phylloptim's own table at `lambda_o = 5e4`: `lambda` 32996, `lambda_emergent` 82996.

That is directly relevant to this PR, since it moves plant onto `TF24_floor`: anything reporting a marginal cost of water from this model now wants the emergent one, and reconstructing it by hand means knowing that `lambda` carries only the hydraulic half.

plant reports no lambda aux today, so nothing is currently wrong. But `lambda_emergent` beside `shadow_cost` would complete the picture the same way `shadow_cost` completes `profit` — three lines in `RcppR6_classes.yml` plus an aux. **Not done here**, since the ask was the version bump; say the word and I'll add it to this PR, or file it.

Recorded in NEWS, since the point of `==` is that an upgrade is a deliberate act with its baseline diff written down.

Issue #525, opened 2026-06-24
[acclimation] [speed] Track optimal root-collar ψ as a relaxing ODE state to avoid nested hydraulics root-finding


dfalster:
Specific instance of #406, applied to the hydraulics solve.

TF24's per-step cost is dominated by **nested root-finding inside the leaf
model**, re-solved from scratch on every ODE RHS evaluation, per individual
(`Leaf::find_root_collar_psi`):

- golden-section search maximising profit over leaf ψ (~10–15 profit evals)
- TOMS748 root-finds (×3) for soil→root-collar water continuity (~6–8 evals each)
- TOMS748 again for the ci/assimilation solve

→ ~60–120 `E_from_Soil_to_Root_Collar` evaluations per individual per RHS eval
(×15–21 again under DeepCrown shading).

### Idea

Carry the optimal hydraulic state — `opt_root_psi` (the supply–demand matching
variable) or `opt_psi_stem` — as an **extra ODE state** (per #406).

This **must use the gradient-ascent form**, not relaxation:

```
dx/dt = k · ∂profit/∂x          # evaluated at the current x
```

The relaxation form `dx/dt = (x* − x)/τ` is a non-starter here: it needs
`x* = argmax profit` explicitly, which *is* the GSS/root-find we are trying to
delete. Gradient-ascent never computes `x*` — its fixed point is `x*` (where
`∂profit/∂x = 0`), so `x` tracks the moving optimum without solving for it. Cost
is the current value plus one gradient (a finite-difference perturbation or one
AutoDiff pass, #491). This removes the GSS + collar TOMS748 layers. As `k`→∞ we
recover today's quasi-steady-state behaviour; finite `k` is acclimation.

### Key risks to resolve

- **Stiffness vs tracking** — the effective relaxation rate is
  `k · (curvature of profit)` near the optimum, *not* a freely tunable τ. Sharp
  optima → stiff (tiny steps under the explicit fixed-step solver, eating the
  speedup); flat optima → sluggish tracking. Whether a usable window exists is
  make-or-break, and it is partly set by the biology.
- **Gradient reliability is load-bearing** — leaf solver tolerances (~1e-3/1e-4)
  can swamp a finite-difference gradient and push `x` the *wrong* way, not just
  slow it. This moves AutoDiff (#491) from nice-to-have toward required.
- **DeepCrown multiplicity** — optimisation runs per crown light point; one scalar
  state can't track per-point optima. Scope initial work to MeanLight.
[...]

Issue #174, opened 2016-02-10
Problem growing plants via `grow_plant_to_size`


dfalster:
Hi @richfitz ,
I'm hoping you might be able to take a look at this. I'm using the `grow_plant_to_size` a lot and mostly it's great. However, this error has come up a couple of times now (when running plant with stateline) : 

```
> res <- grow_plant_to_size(pl, sizes, "height", env, 1000, warn = FALSE, filter = TRUE)
Error: Step size became too small
```

I suspect it's arising because some strange parameter values are being fed in. That can't be avoided when running inside stateline (bc it's searching parameter space it's bound to throw out some wilds values) . Here's a MWE:

```

library(plant)  
traits <- trait_matrix(
 c(10.8552329005728,0.0105249489979936,633.641169104633,0.00227017846696535,3.61377429973252,0.66734,1.99575855688525,0.0567301588978768,900.481925352216,0.278101287831634,0.0183902655344452,0.523884989942541,4.34167189261234,4.520285,0.824983001217059,0.0344755338576102,0.245619588820917,0.260495,5031.78411788144,1.06992910086522,1.1368443297711,50,1),
 c("eta","lma","rho","theta","a_l1","a_l2","a_r1","a_b1","r_r","k_b","k_r","omega","B_kl1","B_kl2","B_ks1","narea","B_lf1","B_lf5","B_lf4","B_rs1","B_rb1","hmat","c_r1")
  )

s <- strategy(traits, scm_base_parameters())
pl <- FF16_PlantPlus(s)

env <- fixed_environment(1)
sizes <- seq_range(c(pl$height, 50), 50)
res <- grow_plant_to_size(pl, sizes, "height", env, 1000, warn = FALSE, filter = TRUE)
```

It's failing within the Ode_Runner code:
```
1: grow_plant_to_size(pl, sizes, "height", env, 1000, warn = FALSE, filter = T
2: grow_plant_bracket(plant, sizes, size_name, env, time_max, warn)
3: runner$step()
4: OdeRunner___PlantRunner__step(self)
```

richfitz:
Even shorter code to trigger the problem

```r
library(plant)
traits <- trait_matrix(
[...]

PR #637, opened 2026-08-26
Refuse an unknown trait name

### LINKED ISSUE #636: generate_strategy() accepts an unknown trait name in silence
## Why

A trait name that does not exist is accepted in silence, and the run proceeds at the default. On `develop` at `2c270be1`:

```r
p0 <- scm_base_parameters("FF16")
generate_strategy(p0, trait_matrix(0.5, "lmaa"))[[1]]$pars$lma
#> 0.1978791     <- the DEFAULT. No warning, no error.
```

That is not a typo-catching nicety. It is a whole simulation that looks like it varied a trait and did not, and there is nothing in the output to say so. A trait sweep written that way runs the entire sweep at one parameter value; a calibration reports a plausible fit to a model it never perturbed. The class of failure is the one this family keeps meeting — a pipeline that builds clean and holds nothing — and it is worse here than in most places because the *result* is well-formed.

## The mechanism, which is not "the name is dropped"

`generate_strategy()` ([R/util_model.R](https://github.com/traitecoevo/plant/blob/develop/R/util_model.R)) does:

```r
pars <- strategy$pars
pars[trait_names] <- xi
strategy$pars <- pars
```

`pars[["lmaa"]] <- 0.5` on a list **appends** a new element. So the misspelt trait is not discarded — it is added, and `"lmaa" %in% names(s$pars)` comes back `TRUE`. Then the generated converter reads only the names it knows:

```cpp
template <> inline plant::TF24_Pars as(SEXP x) {
  ...
  Rcpp::List xl(x);
  ret.lma = Rcpp::as<double >(xl["lma"]);
  // ... one line per known field; extras never looked at
}
```

`inst/include/plant/RcppR6_post.hpp` also says `// No current support for a hook`, so there is no C++-side place to validate. **The check has to be R-side, in `generate_strategy()`**, which is the documented trait route and where the name is still attached to something meaningful.

## Why now

Nothing about this is new, but #635 renames five TF24 parameters — `p_50 -> stem_P50`, `c -> stem_c`, `b -> stem_b`, `beta2 -> TF24_beta2`, `g1_TF24 -> TF24_cost_scale`. Anyone migrating a script who mistypes one, or misses one, gets a complete run at the old default and no signal at all. The rename makes an existing hazard likely to be met rather than merely present.

## Proposed fix

Refuse a trait name that is not a field of the model's `pars`, naming the offenders and (ideally) their nearest valid matches. All four models carry a nested `pars`, so the check is uniform: FF16 32 fields, K93 11, TF24 and TF24f 62.

Check **after** the hyperpar runs, not before, since the hyperpar is what turns input traits into the parameter columns actually assigned.

## ⚠️ Two things to settle in the PR

1. **Blast radius.** The guard turns a silent no-op into an error, so any existing caller passing a column that is not a `pars` field starts failing. That is the point, but it needs measuring rather than assuming — full suite, vignettes, and `inst/scenarios/` — and if something legitimate does pass an extra column, the guard needs to accommodate it deliberately rather than be weakened.

2. **Top-level strategy fields are a genuine gap, discovered here, and should NOT be fixed in the same PR.** `pars[trait_names] <- xi` only ever writes into `pars`, so TF24f's `k_acclim`, `psi_fd_step` and `use_ad_gradient` — real, user-settable, R-exposed strategy fields — are silently ignored as traits today. `k_acclim` in particular is the acclimation gain the #525 k-sweep varies, so wanting it as a trait is not hypothetical. The guard will convert that silence into an error, which is an improvement and is the correct first step; *extending* the trait mechanism to reach top-level fields is a feature and belongs in its own issue.

## Verification

- The failing case above errors, naming `lmaa`.
- A correct trait sweep is unchanged: same strategies, same numbers.
- Full suite green, or every new failure explained as a caller that was silently doing nothing.

dfalster:
A misspelt trait was silent, and not by being dropped: assigning into
a list APPENDS, so the junk element reached the generated
Rcpp::as<*_Pars>, which reads only fields it knows. The run then
reported at the DEFAULT of the parameter the caller meant to vary, with
nothing in the output to say so -- a sweep running its whole length at
one point.

The valid set is wider than names(pars): a hyperpar may consume a trait
it never assigns, so hyperpars now declare those via an input_traits
attribute. FF16's narea is the case, and a pars-only check would have
rejected a sweep that works.

Closes #636


dfalster:
Closes #636.

## What the guard found, which is why it is worth having

The full suite came back with **one** failure, and it was a real one — `test-ode-individual-runner.R`'s regression test for issue #174 was passing ten trait names that are not traits:

```
Unknown trait names for FF16_Strategy: B_kl1 (did you mean a_l1?), B_kl2 (did you mean a_l2?),
B_ks1, narea, B_lf1 (did you mean a_l1 / a_f1?), B_lf5, B_lf4, B_rs1 (did you mean a_r1?),
B_rb1 (did you mean a_r1 / a_b1?), c_r1 (did you mean a_r1?)
```

Sorting them out is most of this PR's content, because they are three different things:

| name(s) | what it actually is | verdict |
|---|---|---|
| `narea` | read from the matrix by FF16's hyperpar, derives `a_p1`, `a_p2`, `r_l` | **legitimate trait** — must keep working |
| `B_kl1`, `B_kl2`, `B_ks1`, `B_lf1`, `B_lf4`, `B_lf5`, `B_rs1`, `B_rb1` | **constructor arguments** of `make_FF16_hyperpar()` | never did anything |
| `c_r1` | not a parameter, not an argument | never did anything |

The line above the matrix is `hyperpar <- make_FF16_hyperpar()` — **no arguments** — so the eight `B_*` columns were not configuring the hyperpar either. They were doubly inert: they never reached the trade-off coefficients, and they never reached the model. The test has been asserting a growth-run regression on a strategy it thought it had parameterised eleven ways and had parameterised two.

I removed the nine and kept `narea`. Verified rather than assumed — building the strategy from the 23-column and 14-column matrices and diffing all 32 `pars` fields:

```
pars fields differing between the 23-column and 14-column matrices: NONE — bit-identical
```

So the #174 regression still exercises exactly what it did.

## ⚠️ `narea` is why the valid set is not `names(pars)`

This is the part that would have been wrong if I had shipped the obvious guard. `narea` is **not a parameter of any model** — FF16's hyperpar consumes it via `with_default("narea", narea)` and emits `a_p1`/`a_p2`/`r_l` instead. A `pars`-only check rejects a `narea` sweep, which works today and is a documented FF16 trait.

Measured before deciding: of the ten offending columns, `narea` **alone** moved anything (`r_l`, `a_p1`, `a_p2`).

So hyperpars declare their matrix-read non-parameter traits themselves:

```r
attr(f, "input_traits") <- "narea"
```

and the guard's valid set is `names(pars)` ∪ that. Only FF16 needs it — TF24's hyperpar reads `lma`, `rho`, `omega`, `K_s`, `vcmax_25`, `jmax_25`, all of which *are* `pars` fields.

**A caller-supplied hyperpar with no declaration gets the narrow set.** Nothing can know what an arbitrary closure reads, so that is the honest default rather than an oversight; it is documented and tested.

## Two implementation notes

**`xi` is unnamed.** `matrix_to_list()` returns a bare numeric vector positionally aligned with the column names — no names at all. So restricting the assignment had to carry positions:

```r
assign_at <- which(trait_names %in% known)
pars[assign_names] <- xi[assign_at]
```

`xi[assign_names]` returns all `NA`, silently, and every parameter lands as `NA`. There is a test (`pars and consumed traits can be mixed in one matrix`) whose matrix deliberately puts the consumed column *first*, so the assigned columns are not a prefix — that is the arrangement that catches it.

**The check runs after `hyperpar()`, not before.** The hyperpar is what turns input traits into the columns actually assigned, and its own guard already refuses an input colliding with something it derives. Checking the caller's original column names would reject every legitimate hyperparameterised trait.

**`strategy$name` does not exist at R level** — it silently yields `NULL` and an unusable message. `class(strategy)[[1]]` is what names the model.

## Also fixed, since the guard made it visible

Consumed traits are no longer appended to `pars`. Previously `narea` became a junk element the C++ converter ignored — harmless to the model, but it made a strategy compare unequal to a default one over something the model never reads. Asserted with `expect_setequal(names(s$pars), names(default$pars))`.

## Verification

- **Full suite 2987 pass / 0 fail**, with the #174 test fixed and 31 new checks in `tests/testthat/test-trait-names.R`.
- `trait_matrix()` call sites outside the suite audited (`inst/reports/`, `scripts/`, `R/benchmark.R`, `R/individual.R`, `R/scenario_eval.R`): all name real parameters. `inst/scenarios/scenario_mapping.csv`'s `param` column is `K_s`, `lma`, `rho`, `theta`, `vcmax_25`, `root_depth_shape_eta` — all `pars` fields, and covered by `test-scenario-eval.R`.
- No C++ changes, so nothing numeric can move; the guard either errors or is a no-op.

## Left open, deliberately

**Top-level strategy fields are not traits, and this PR does not make them ones.** `pars[trait_names] <- xi` only ever writes into `pars`, so TF24f's `k_acclim`, `psi_fd_step` and `use_ad_gradient` — real, user-settable, R-exposed fields — cannot be varied as traits. `k_acclim` is the acclimation gain the #525 k-sweep varies, so wanting it is not hypothetical.

Before this PR that was silent; now it is an error naming `k_acclim`. That is the honest report, not the fix. Extending the trait mechanism to reach top-level fields is a feature — happy to file it as a follow-up issue if you want it.

PR #640, opened 2026-08-26
Re-bless the scenario gateway and diff its numbers

### LINKED ISSUE #639: Scenario gateway baseline predates #590 and is red on develop
`tests/testthat/test_data/scenario_baseline.rds` no longer matches any current build, so `PLANT_RUN_SCENARIOS=1` fails on an unmodified branch. Every scenario has moved, and the size of the move identifies the cause.

Measured on `feature/events-522` (`max_patch_lifetime = 100`, 1 worker), `offspring_production` current vs recorded:

| scenario | baseline | current | rel. change |
|---|---|---|---|
| S01 | 1.48872077 | 0.4797763201 | −0.68 |
| S02 | 0.09963559347 | 0.01153314819 | −0.88 |
| S03 | 5.498561487e-12 | 1.665416219e-10 | **+29.3** |
| S04 | 4.745911732e-13 | 1.317815017e-11 | **+26.8** |
| S05 | 5.888101886e-13 | 2.470008358e-11 | **+40.9** |
| S06 | 3.261465575e-13 | 6.570072526e-12 | **+19.1** |
| S07 | 1415.200202 | 810.7050184 | −0.43 |
| S08 | 3.230258314e-14 | 2.025578048e-14 | −0.37 |

The thresholded diff the test reports is just `persists S01: TRUE -> FALSE`, which understates it considerably.

19–41× on S03–S06 is the range the birth-date/density coordinate change (#590) is documented to produce — `test-scenario-gateway.R:32-34` says switching the coordinate "multiplies R0 by 2.4x to 47x and moves S01 across R0 = 1". So the recorded baseline predates it, and possibly other work since; #633's +0.36% was also never re-blessed.

## Why this is worth its own issue

The gateway is a baseline diff, so a red result is only informative if the baseline is current. As it stands every PR that runs it inherits a failure it did not cause, and the cheapest way to find that out is to build a second copy of the base branch and re-run — which is expensive and has to be redone by each reviewer. Two PRs have now paid that cost.

## What it needs

1. Confirm the cause is #590 (plus anything after it) rather than a live regression — the direction and magnitude fit, but that has not been demonstrated by bisection.
2. `make bless-scenarios` on a clean `develop`, as its own commit, with the table above recorded so the re-baselining is legible rather than a silent binary blob change.
3. Consider whether the gateway should diff the numeric scorecard alongside the two thresholded booleans. The bools reduce a 29× change to one flipped flag, and the comment at `test-scenario-gateway.R:30-32` already notes that `observed` alone cannot move.

## Not blocking #626

Raised while working #626 (PR #638), which is *not* the cause: `run_scm` on the default environment there is bit-identical (`offspring_production = 11.143684537309916`) between builds with and without that change, and every scenario runs a uniform soil column, which #626 leaves untouched. Deliberately not re-blessed inside that PR, since folding an unrelated re-baselining into it would hide what actually moved.

dfalster:
`scenario_baseline.rds` matched no build since #583, so
`PLANT_RUN_SCENARIOS=1` failed on an unmodified `develop` and every PR
that ran it inherited a failure it had not caused.

Bisected: phylloptim 0.2.0 -> 0.6.0 (in #633) moves R0 by 8x-32x on five
scenarios and takes S01 across R0 = 1; #619 moves five again on top. Not
#590. Everything else since is inert. Re-blessed on the develop tip as
its own commit, table in NEWS. The gateway now also diffs
`offspring_production` (`SCENARIO_TOL`, default 1e-3) and prints the
whole table on failure.

Closes #639.


dfalster:
## Bisection

One build per commit, each against the phylloptim/odelia sha that commit pins (installed from a `git archive` of the pinned sha into a scratch library, so no sibling working tree leaks in). macOS arm64, R 4.6.1, `-O2` via `pkgbuild::compile_dll(debug = FALSE)`, `max_patch_lifetime = 100`, `SCENARIO_WORKERS = 8`. Worker count does not enter the numbers — the scenarios are independent runs.

`offspring_production`:

| scenario | baseline | #583 `133499b8` | #606 `39c0c657` | #633 `b374ce3b` | #619 `745dd600` | develop `60708313` |
|---|---|---|---|---|---|---|
| S01 | 1.488721 | 1.488721 | 1.488721 | 0.2725883 | 0.4797763 | 0.4797763 |
| S02 | 9.963559e-2 | 9.963559e-2 | 9.963559e-2 | 2.784370e-3 | 1.153315e-2 | 1.153315e-2 |
| S03 | 5.498561e-12 | 5.498561e-12 | 5.498561e-12 | 1.745197e-10 | 1.665416e-10 | 1.665416e-10 |
| S04 | 4.745912e-13 | 4.745912e-13 | 4.745912e-13 | 1.357792e-11 | 1.317815e-11 | 1.317815e-11 |
| S05 | 5.888102e-13 | 5.888102e-13 | 5.888102e-13 | 5.052547e-12 | 2.470008e-11 | 2.470008e-11 |
| S06 | 3.261466e-13 | 3.261466e-13 | 3.261466e-13 | 5.082346e-12 | 6.570073e-12 | 6.570073e-12 |
| S07 | 1415.200 | 1415.200 | 1415.200 | 837.6889 | 810.7050 | 810.7050 |
| S08 | 3.230258e-14 | 3.230258e-14 | 3.230258e-14 | 2.711131e-13 | 2.025578e-14 | 2.025578e-14 |

Step ratios:

| scenario | -> #633 | -> #619 | net |
|---|---|---|---|
| S01 | 0.183 | 1.76 | 0.322 |
| S02 | 0.0280 | 4.14 | 0.116 |
| S03 | 31.74 | 0.954 | 30.29 |
| S04 | 28.61 | 0.971 | 27.77 |
| S05 | 8.581 | 4.889 | 41.95 |
| S06 | 15.58 | 1.293 | 20.14 |
| S07 | 0.592 | 0.968 | 0.573 |
| S08 | 8.393 | 0.0747 | 0.627 |

`persists` changes once, at #633: `S01 TRUE -> FALSE`. Nothing else moves on either flag anywhere in the range.

`b3472d43` (#608) was measured rather than `b374ce3b` (#633) — the two commits between them, #607 and #608, touch only `NEWS.md`, `test-leaf.r` and `notes/`, so they compile to the same library. The `#633` column is that measurement.

## Which half of #633

Rebuilt `b374ce3b` against **phylloptim 0.6.0 with odelia held at 0.2.1** (the `LinkingTo` floor relaxed for the probe; nothing else touched). All eight scenarios identical to the odelia-0.3.1 build, so odelia 0.2.1 -> 0.3.1 is inert on the gateway and the whole move is phylloptim 0.2.0 -> 0.6.0.

That fits odelia's two behavioural changes in the range: `stop_domain` step rejection (odelia#56) had no declared domain to act on until #619 added `Patch::ode_state_valid`, and rejecting a non-finite error estimate (odelia#54) only bites a step that was already diverging.

## Why #590 looked like the cause and is not

- `53d0a200` (#590) is an ancestor of the commit the baseline was blessed at.
- The baseline's own metadata records `node_density_in_birth_date = TRUE` — the recorded numbers were already in birth date.
- The magnitude coincidence is real but is a coincidence: #590's documented 2.4x-47x and #633's 8x-32x are both "a small perturbation to a water-limited run, read through an exponentially small survivorship integral".

## Provenance of the old baseline

`git_commit = 99e44db6`, `git_dirty = TRUE`. It could not have been produced by that commit as committed: there, `run_scenarios()` still defaulted to `control()` (height) and `scenario_run_metadata()` had no `node_density_in_birth_date` field to write. The tree that produced it is the one that landed as `133499b8`, which is why that is the control and reproduces it exactly.

## Verification of the new test

Against the stale baseline it now reports, in full:

```
Scenario outcomes changed vs baseline (re-bless via `make bless-scenarios` if intended): persists S01: TRUE -> FALSE | offspring_production moved on 8/8 scenario(s) (tol 0.001)
id         baseline        current        rel
S01        1.488721      0.4797763     -0.678 *
S02      0.09963559     0.01153315     -0.884 *
S03    5.498561e-12   1.665416e-10       29.3 *
S04    4.745912e-13   1.317815e-11       26.8 *
S05    5.888102e-13   2.470008e-11       40.9 *
S06    3.261466e-13   6.570073e-12       19.1 *
S07          1415.2        810.705     -0.427 *
S08    3.230258e-14   2.025578e-14     -0.373 *
```

which reproduces the table in #639 exactly. Against the re-blessed baseline it passes.

## Tolerance

`SCENARIO_TOL` defaults to `1e-3`. It is not measured against cross-platform drift — the gateway is opt-in and has only ever been run on macOS arm64 — so it is set from the other end: small enough to catch the smallest move anyone has wanted to see (#633's +0.36% on the reference run), large enough to sit well clear of floating-point reordering. If a platform proves noisier, raise it there rather than widening the default; the two classifications are the part that is meant to be platform-stable.

PR #643, opened 2026-08-28
Fix run_mutant() for TF24 via odelia 0.4.0

### LINKED ISSUE #642: `run_mutant()` fails for TF24 / TF24f with a spurious "TF24 storage is negative" domain error
## Summary

`SCM$run_mutant()` — the resident/mutant invasion-fitness replay path — fails
for every TF24 and TF24f strategy tested, including the **trivial identity
case** where the "mutant" strategy is identical to the resident. The failure
is the TF24 NSC storage-pool floor guard (`src/tf24_strategy.cpp:217-222`,
added for #517), which is designed to trigger the ODE stepper's
adaptive shrink-and-retry, but `run_mutant()` pins the mutant to the
resident's fixed recorded step times (`node_schedule$r_set_use_ode_times(true)`
in `SCM::run_mutant`, `inst/include/plant/scm.h:294-315`), which appears to
disable that retry mechanism. The result is that a domain violation the
resident's own adaptive run tolerated and recovered from becomes a hard,
unrecoverable error on replay.

This blocks any resident-vs-mutant invasion-fitness analysis (the standard
adaptive-dynamics / ESS workflow: grow a resident to its own patch history,
then test whether a rare mutant strategy can invade without letting it
reshape the environment) for any TF24-family strategy.

## Environment

- `plant` 2.0.0.9002, commit `c64b8d92625592bf115d77b3ac111413d4ebadec`
  (2026-08-27)
- R 4.5.2, macOS (Darwin 25.6.0)

## Minimal reproducible example

The mutant here is **identical to the resident** (`TF24_floor_lambda_o = 0`
for both), so this is the same identity check
`tests/testthat/test-mutant.R` already uses for FF16
(`"mutant method densities"`) — for FF16 it passes; for TF24 it does not.

```r
library(plant)

p0 <- scm_base_parameters("TF24")
p0$max_patch_lifetime <- 20

p_resident <- add_strategies(p0, trait_matrix(0, "TF24_floor_lambda_o"),
                              hyperpar = TF24_hyperpar, birth_rate = 1)

env <- Environment("TF24")
env$set_soil_water_state(rep(0.428 * 0.5, env$get_soil_number_of_depths()))
env$extrinsic_drivers_set_constant("rainfall", 1)

ctrl <- control()
ctrl$save_RK45_cache <- TRUE  # required for run_mutant's replay

scm <- run_scm(p_resident, env = env, ctrl = ctrl)
scm$net_reproduction_ratios[[1]]   # 22.46469 -- resident runs fine

p_mutant <- add_strategies(p_resident, trait_matrix(0, "TF24_floor_lambda_o"),
                            hyperpar = TF24_hyperpar, birth_rate = 1)
scm$run_mutant(p_mutant)
```

### Actual behavior

```
Error: TF24 storage is negative (-1.53893e-07 kg): the pool's flow does not
leave [0, capacity], so this is a step that overshot the empty boundary
```

### Expected behavior

`scm$net_reproduction_ratios` returns `c(22.46469, 22.46469)` (or something
extremely close) — a mutant strategy identical to the resident, replaying the
resident's own environment, should reproduce the resident's own R0 exactly,
per the identity `test-mutant.R` documents and tests for FF16.

## Scope of the bug

- **Independent of the strategy difference being tested.** Reproduces with
  the mutant identical to the resident (`lambda_o = 0` vs `0`), and with every
  value tried in `lambda_o ∈ {1, 10, 100, 500, 1000, 1200, ..., 3000, 5e4,
  1e5}`. The overshoot magnitude is the same tiny value
  (`-1.53893e-07` kg in this configuration) for the identity case and small
  perturbations, consistent with it being a step-size/replay artifact rather
  than anything about the mutant's physiology.
- **Independent of rainfall.** Reproduces at `rainfall = 1` and
  `rainfall = 0.75`.
- **Reproduces on both `TF24` and `TF24f`.**
- **Depends on `max_patch_lifetime`.** At `max_patch_lifetime = 10` the
  identity case above succeeds; at `20`, `50`, and `100` it fails. This is
  consistent with the resident's adaptive stepper hitting (and recovering
  from) the storage floor at some point during the run whenever the run is
  long enough, and the mutant replay then re-encountering that same point
  without the ability to recover.

## Suspected root cause

`Patch::set_mutant()` / `SCM::run_mutant()` marks the run `is_mutant_run =
true` and pins the ODE solver to the resident's exact recorded step times
(`inst/include/plant/scm.h:308-310`):

```cpp
node_schedule = make_node_schedule(parameters);
node_schedule.r_set_ode_times(patch.step_history);
node_schedule.r_set_use_ode_times(true);
```

Per the comment at `src/tf24_strategy.cpp:196-209`, the storage-floor guard is
deliberately a *retryable* domain error rather than a floor/clamp:

> "the pool is refused there rather than floored, and the stepper shrinks and
> retries (#609, #610)"

That retry-and-shrink presumably works during a normal (adaptive) resident
run. Under `run_mutant`'s pinned/fixed step schedule, the stepper has no room
to shrink below a checkpoint interval it's forced to hit exactly, so a domain
violation that the resident run absorbed becomes fatal on replay — even for a
mutant with identical parameters to the resident, at the exact same recorded
times.

## Suggested directions

- Confirm whether `use_ode_times = TRUE` runs (not just `run_mutant`, e.g.
  `run_scm(..., use_ode_times = TRUE)`) hit the same issue independent of the
  mutant machinery — that would confirm the pinned-step / domain-retry
  interaction is the real culprit rather than something specific to
  `set_mutant()`.
- Consider whether the storage-floor tolerance (`storage_domain_tol`) could be
  widened specifically for replay steps, or whether the mutant replay could
  fall back to a bounded local retry (e.g. bisecting between the two
  surrounding recorded times) instead of requiring an exact hit.
- At minimum, this should block/skip `run_mutant` support for storage-tracked
  strategies (TF24/TF24f) until resolved, since it currently fails silently
  in a way that looks like a mutant-specific numerical problem rather than a
  structural incompatibility.

## Impact

Any adaptive-dynamics / invasion-fitness analysis on TF24 or TF24f (e.g.
sweeping a trait like `TF24_floor_lambda_o` and asking whether a mutant can
invade a resident population) cannot currently use `run_mutant()` and has no
numerically-equivalent workaround within the package — the only alternative
(growing resident and mutant together via `add_strategies()` +
`run_scm()`) lets the mutant feed back on the shared environment, which
answers a different ecological question (mutual coexistence of two
established populations rather than rare-mutant invasion of a resident at
equilibrium).

dfalster:
TF24's storage pool refuses an overshoot below empty by throwing a domain error, which the adaptive stepper answers by shrinking and retrying — 480 times in a resident run that then completes normally. `run_mutant()` pins the stepper to the resident's recorded step times, and that path called the stepper with no domain handling at all, so the first refusal killed the replay.

Never mutant-specific: `run_scm(..., use_ode_times = TRUE)` failed identically replaying a resident against its own recorded times. Fixed in traitecoevo/odelia#58 by subdividing a refused pinned step to the same endpoint.

`test-mutant.R` now checks the resident-vs-mutant identity for TF24 alongside FF16.

**Blocked on traitecoevo/odelia#58 merging and being tagged `v0.4.0`** — the pin here points at that tag.

Closes #642.


dfalster:
## Verification

`test-mutant.R` is a true regression test — it fails on `develop` and passes with the fix, same test file both times:

```
# develop (odelia 0.3.1)
FAILURE 'test-mutant.R:188:3'  Expected `scm$run_mutant(p1)` not to throw any errors.
  Actually got a <odelia::util::DomainError>: TF24 storage is negative (-1.53726e-07 kg) ...
[ FAIL 2 | WARN 0 | SKIP 0 | PASS 20 ]

# this branch, built against odelia#58
[ FAIL 0 | WARN 0 | SKIP 0 | PASS 22 ]
```

Beyond the test, every case in #642's reported scope, all previously fatal:

| model | `max_patch_lifetime` | resident R0 | mutant R0 | rel. diff |
|---|---|---|---|---|
| TF24  |  10 | 1.432198588e-06 | 1.432193759e-06 | 3.4e-06 |
| TF24  |  20 | 0.1252691257    | 0.1252551481    | 1.1e-04 |
| TF24  |  50 | 9.542491626     | 9.54174979      | 7.8e-05 |
| TF24  | 100 | 22.46469146     | 22.46280171     | 8.4e-05 |
| TF24f |  20 | 0.1270659448    | 0.1270658756    | 5.5e-07 |
| TF24f |  50 | 9.816862989     | 9.816857153     | 5.9e-07 |

All inside the 1e-3 tolerance the FF16 identity block uses. The mpl=100 resident R0 reproduces #642's quoted 22.46469 exactly — worth noting, because the issue's script sets `max_patch_lifetime <- 20` but quotes the number from a lifetime of 100.

## On the test's cost

~14s, which is not free. `max_patch_lifetime = 14` is the cheapest value that covers the bug, measured rather than chosen: at 12 the replay never meets a refusal and passes with or without the fix, so a cheaper test would be vacuous. Most of the 14s is fixed setup — 20 costs 14.9s — so there is little to win by trimming further. The test's first expectation guards against the vacuous case by checking the resident actually produced a live population.

## Two things worth separate attention

**TF24's mutant identity is ~1e-6 at best, against FF16's ~1e-13.** Visible in the mpl=10 row above, which needs no subdivision at all, so it is not caused by this fix. Some of the spread in the other rows *is* attributable to subdivision — odelia#58 documents why (RK stage indices restart per sub-step, so a subdivided interval re-reads the cached environments of the original larger step). But the ~1e-6 floor predates both. Probably worth its own issue.

**The issue's suggested direction — widening `storage_domain_tol` — would have been wrong**, and it is worth recording why, since the number invites it. `-1.5e-07 kg` reads as round-off but is 6% of capacity (`S_max = 2.6e-06`), and attempts rejected during the ordinary resident run reach `r = -4.14`. Any tolerance wide enough to admit these would feed `mortality_storage_dependent_dt() = a_dG1·exp(-a_dG2·r)` exactly the unbounded argument its `[0, 1]` contract was written to prevent. The guard was right; the solver was not.

PR #645, opened 2026-09-04
Report leaf temperature as a TF24 aux

### LINKED ISSUE #625: Leaf temperature should be exposed in TF24strategy (reissued)
Addressing previously closed issues which were poorly defined:

phylloptim currently reports leaf temperature, although this does not appear to be the case in TF24strategy in plant. Ideally, for canopy-level analyses, this should be reported in a similar manner to opt_psi_stem.

Moreover, leaf temperature can vary through the crown, and if an analyses was required whereby leaf temperature variation through the canopy was to be analysed, the deep crown method could report the leaf temperature variation through depth in the canopy along with the analysed height. In principle, this could be done for most, if not all, variables reported from deep crown method.

dfalster:
Adds a `Tleaf` auxiliary variable to TF24/TF24f: the leaf's own temperature at the optimal operating point, not the `leaf_temp` driver. Integrated to the leaf-area-weighted crown mean under the deep-crown shading model, and plotted in `TF24_plot_diagnostics()`.

`aux_size` goes 12 -> 13, appended after `assimilation`, so no name lookup and no positional index below it moves.

Closes #625's first half; the crown-depth profile is left open — see the comment.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -437,8 +437,44 @@ were not previously recorded here:
   * Semantic change: `patch$ode_rates` was a cheap read of stored values and is
     now a full right-hand-side evaluation. `StochasticPatch$compute_rates()` is
     unchanged.
+* TF24/TF24f gain a `Tleaf` auxiliary variable, so their `aux_size` grows by one
+  (#625; see New features for what the slot holds). Additive — it is appended
+  after `assimilation` — so every existing name lookup and every positional index
+  below it is unchanged. Migration, for code that indexes auxs by position or
+  asserts their count:
+  * `aux_size` `12` -> `13`; with `collect_all_auxiliary`, `13` -> `14`
+  * `auxs[[13]]` (`area_sapwood`, `collect_all_auxiliary` only) -> `auxs[[14]]`,
+    or better `ind$aux("area_sapwood")`
+  * no other slot moves; `ind$aux(<name>)` needs no change anywhere
 
 ### New features
+
+* **TF24 and TF24f report leaf temperature as an auxiliary variable, `Tleaf`
+  (#625).** The leaf's own temperature at the optimal operating point, in deg C,
+  alongside `opt_psi_stem` and the other leaf outputs. `aux_size` therefore goes
+  12 -> 13 (13 -> 14 with `collect_all_auxiliary`), appended after
+  `assimilation`, so existing positional indices are unchanged except
+  `area_sapwood`, which moves from 13 to 14.
+
+  It is an **output, not the `leaf_temp` driver**, and that is the whole reason
+  for the slot. With `pars$use_energy_balance` at its default of `0` the leaf
+  runs at the prescribed driver and `Tleaf` equals it exactly — reported anyway,
+  rather than `NA`, so the column can be plotted against anything. Turn the
+  Penman-Monteith energy balance on and the leaf solves its own temperature from
+  its transpiration at every operating point; that value was computed, used to
+  re-derive the whole Farquhar temperature block, and then discarded, so the one
+  quantity the PM path exists to produce was the one a canopy-level analysis
+  could not read. Measured on a 5 m plant at PPFD 1800 with air at 30 °C, the
+  leaf sits at **39.2 °C**.
+
+  Under `shading_model = "deep-crown"` it is integrated to the leaf-area-weighted
+  crown mean alongside `profit`, `transpiration`, `assimilation` and the rest,
+  and NOT left at whichever quadrature node the crown loop ended on. Note what
+  that means for interpretation: it is a **crown mean**, not any single leaf's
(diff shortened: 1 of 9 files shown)


dfalster:
## What the slot holds, and why it needed one

phylloptim's `Leaf` has carried the leaf's operating-point temperature as `Tleaf_` for a while — an **output**, distinct from the `leaf_temp` driver it is handed. TF24 read the driver, computed `Tleaf_`, used it to re-derive the whole Farquhar temperature block, and then discarded it. So the one quantity the Penman–Monteith path exists to produce was the one a canopy-level analysis could not read: you could see the assimilation the temperature caused, but not the temperature.

With `pars$use_energy_balance` at its default `0`, `Tleaf` equals the driver exactly. That is deliberate rather than `NA` — an output column that is sometimes a number and sometimes `NA` cannot be plotted against anything. So at TF24's defaults the column is flat at 25 °C, and the slot only earns its keep with the energy balance on.

Measured, 5 m plant, PPFD 1800, air 30 °C, well watered:

| `use_energy_balance` | `Tleaf` (°C) |
|---|---|
| 0 (default) | 30.00 |
| 1 | 39.17 |

## Deep crown

Under `shading_model = "deep-crown"` the crown loop runs one hydraulic optimisation per quadrature node and integrates each leaf output to a leaf-area-weighted crown mean. `Tleaf_` was not in that set, so left alone it would have reported whichever node the loop happened to end on — a finite, plausible, wrong number, and not the crown top or the centre either. It is now integrated alongside `profit`, `transpiration`, `assimilation` and the rest.

⚠️ **It is a crown mean, not a leaf's temperature.** A canopy with a hot top and a cool base reports the average. Under a light gradient (`0.08 + 0.9·(z/h)²`, h = 15 m, PM on, air 30 °C) the three shading models genuinely separate:

| shading model | `Tleaf` (°C) | `assimilation` |
|---|---|---|
| mean-light | 36.822 | 1.346 |
| crown-centre | 36.761 | 1.381 |
| deep-crown | 36.813 | 1.501 |

Under a **uniform** light environment all three agree exactly, which is the documented invariant (see `test-canopy-methods.R`) — and the reason the crown-mean test had to impose a light spline on the environment directly. A stand-alone `Individual` sits in no patch, so its light is uniform, and a version of that test without the gradient passed while proving nothing.

## What is NOT in this PR

The issue's second half — reporting each leaf output *against crown depth* rather than as a crown mean — is left open, and deliberately. An aux slot is fixed width; a per-quadrature-node profile is a vector whose length is a property of the integration rule. Exposing it means a new R-facing accessor (RcppR6 regeneration), and a decision about what it should return under `mean-light`/`crown-centre`, which evaluate at one point and have no profile to report. That is a separate feature with its own design questions, not a line to add here.

## Verification

- Full suite green: **3202 pass, 0 fail**, 1 skip (the opt-in scenario gateway).
- Scenario gateway (`PLANT_RUN_SCENARIOS=1`): matches the recorded baseline, no re-bless needed (`FAIL 0 | SKIP 0 | PASS 2` — SKIP 0 confirms it actually ran rather than skipping on the env var). Both classifications and every `offspring_production` are within the 1e-3 tolerance.
- No `scientific_version` bump, and the argument is structural rather than empirical: aux slots are write-only diagnostics that no rate reads, and `Tleaf_` has no reader outside the function that writes it — every read sits immediately below its own write inside `set_leaf_states_rates_from_psi_stem`. Setting it after the crown loop therefore cannot move a simulation number.
- Built against the pinned `phylloptim@v0.8.1` (`git archive` of the tag, not the sibling checkout, which is on 0.9.0).

## Also in here

`TF24_plot_diagnostics()` gains a leaf-temperature panel, on the reading that "reported in a similar manner to `opt_psi_stem`" includes the figure that already plots `opt_psi_stem`. The existing test for that function only checked the returned object's class, so a missing column would not have surfaced until the plot was drawn — it now asserts the aux columns the figure consumes are present in `results$species`.

⚠️ **Unrelated, noticed and deliberately NOT included:** `man/rainfall_pulse.Rd` and `man/run_scm.Rd` are stale against their roxygen sources on `develop` — `rainfall_pulse()` gained an `env` argument and `run_scm()`'s `events` moved position, and neither `.Rd` was regenerated. Running `devtools::document()` here rewrites them. I reverted both so this PR stays one change; they want their own commit.

## Migration

Additive, appended after `assimilation`. `ind$aux("<name>")` needs no change anywhere. Only code that indexes auxs positionally or asserts their count is affected:

* `aux_size` `12` → `13`; with `collect_all_auxiliary`, `13` → `14`
* `auxs[[13]]` (`area_sapwood`, `collect_all_auxiliary` only) → `auxs[[14]]`

itowers1 (approved):
Looks good to me!

Issue #614, opened 2026-08-10
[AutoDiff] [acclimation] TF24f trait gradients need a dpsi/dtheta companion state


dfalster:
#527 gave TF24f an exact `∂profit/∂ψ`, which made the tracking ODE smooth and
beat the finite difference on both speed and accuracy. That is a gradient in
**ψ**.

A gradient in **traits** is a different object, and it does not exist — for TF24f
or anywhere else in plant. The only gradient plant consumes from phylloptim is
that one scalar `dprofit_droot_collar_psi`; `Node::growth_rate_gradient`
(`inst/include/plant/node.h:228-256`) is a finite difference in *height*, and
`optimise_individual_rate_at_size_by_trait` is derivative-free.

## Why phylloptim's trait gradient does not transfer

`leaf_gradient()` gets `dψ*/dθ` by applying the implicit function theorem to the
first-order condition:

```
dψ*/dθ = −M/H,   M = ∂²profit/∂ψ∂θ,   H = ∂²profit/∂ψ²
```

That derivation assumes `∂profit/∂ψ = 0`, which for TF24f is **false for every
individual after birth, by construction** — the residual is the acclimation rate
(`src/tf24f_strategy.cpp:37`). Using `−M/H` anyway is the same class of error
phylloptim's own active-set guard exists to prevent; at a pinned point that guard
measures the bare composite wrong by up to 3.5e+07.

## What is actually needed

Differentiate the tracking law `dψ/dt = k·∂profit/∂ψ(ψ, θ)` in θ. With
`s ≡ dψ/dθ`:

```
ds/dt = k_acclim · (M + H·s),      s(0) = −M/H
```

One extra scalar ODE per traced parameter. `M` and `H` are exactly the two
quantities phylloptim already forms inside its gradient path; exposing them is
traitecoevo/phylloptim#88.

Two things fall out that make this checkable:

- **The birth seed is the IFT value**, because `set_initial_states()` genuinely
  does solve the optimum once (`src/tf24f_strategy.cpp:111-121`).
- **The fixed point is `−M/H`.** So large `k_acclim` must reproduce phylloptim's
  answer, mirroring the existing `k_acclim`-convergence test at
  `tests/testthat/test-strategy-tf24f.R:265-299`. That is the acceptance test.

## Depends on traitecoevo/phylloptim#87

Without `profit` among phylloptim's differentiated outputs there is nothing on
plant's side of the boundary for `s` to propagate into: plant's carbon comes from
[...]

Issue #512, opened 2026-06-23
[TF24 allometry] Review empirical data on plasticity in allometry (height-to leaf area for individuals) in response to environment (light or water)


dfalster:
We know plants become a little straggly as they begin to snuggle. This should manifest as variation in leaf area at a given height. 

What data exists on how big the variation is?

Could also research leaf loss as a drought response?

PR #647, opened 2026-09-07
Flexible allometry: leaf and sapwood as ODE states

dfalster:
Frees leaf and sapwood area as ODE states, so replacement of turned-over
leaf becomes conditional on the marginal leaf's carbon balance and the stem
can track the Huber value that maximises growth.

Both are OFF at the defaults (`a_pl0 = 0`, `a_sw = 0`) and the model is then
bit-exact: the scenario gateway passes unmodified against develop.

TF24 `v10 -> v11`; TF24f `v10.1 -> v11.1`. Merges #617.

Draft: the demo's "thinning does not buy survival" section still quotes
pre-#617 numbers. Detail, measurements and retractions in the first comment.


dfalster:
## What this does

Two mechanisms, both defaulting to off.

**1. Conditional leaf replacement (`a_pl0`).** Leaf turnover is currently charged as a cost and the tissue restored unconditionally by the allometry, so a struggling plant cannot economise by carrying less leaf. Replacement now depends on whether the *marginal* leaf is paying for itself. Carbon not spent and leaf mass not replaced are two halves of one decision and conserve exactly by construction.

**2. Sapwood tracks its growth optimum (`a_sw`).** `dψ/dt` gains `a_sw · R_s`, an integral controller on the marginal **growth** return of sapwood area. It converges onto `R_s = 0`, and because integrating is averaging, a small `a_sw` makes the stem follow the long-run mean of a signal that swings with the weather — no extra tracked state.

States are carried as departures, `log(actual/preferred)`, which is what makes the fixed-allometry model recoverable bit-for-bit rather than approximately.

## The objective is growth, not production

This is the design decision everything else follows from. Against **production**, extra sapwood is nearly always worth building: it raises `k_max`, and nothing in `P` charges the plant for the leaf area it did not build instead. `R_s` subtracts that price. The optimum is not one number — after #617 it rises with height and again as soil dries:

| stem area per leaf area, ×pipe model | 5 m | 10 m | 16 m | 20 m |
|---|---|---|---|---|
| soil 0.30 | 0.87 | 1.03 | 1.17 | 1.26 |
| soil 0.20 | 0.89 | 1.05 | 1.21 | 1.30 |
| soil 0.15 | 1.04 | 1.28 | 1.54 | 1.68 |

**The defining correctness property is that `R_s` crosses zero exactly where `dh/dt` peaks** — verified in all twelve cells above on a 0.01 grid in ψ. A controller built on production instead is still monotone, still converges, and is still wrong, so the test asserts the crossing on the grid rather than by interpolation.

## ⚠️ Two retractions

**The envelope theorem does not apply here.** I claimed the sensitivity was nearly free: profit is already maximised, so re-evaluate at the collar potential already found and the indirect term vanishes. That holds for `opt_psi_stem_`, which is genuinely maximised. It does **not** hold for `opt_root_psi_`, which is a **root find** — the collar potential where the soil-root network's supply matches the leaf's demand. A constraint is not an optimum, so `d(profit)/d(collar) ≠ 0`.

Holding it fixed under-reported `d(profit)/d(k_max)` by **15–19%**, putting the controller's zero at 1.22× where growth peaks at 1.28×. A full re-solve matches a finite difference of the model's own assimilation to **0.09%**. Consequence: `a_sw > 0` costs a second full leaf solve per rate evaluation; `a_sw = 0` skips the block. plant #614 (analytic `d(profit)/d(k_max)`) would remove that cost and is worth more than it looked.

**Why it survived so long is worth its own warning.** In wet soil the growth peak is nearly flat — at 16 m, `dh/dt` varies by 6e-4 relative between 1.13× and 1.22× — so a 0.025 scan grid could not resolve a 0.03 offset in ψ and the zero *appeared* to land exactly on the peak. A coarse grid here does not merely lose precision; it manufactures an exact-looking agreement.

**Shedding's benefit is a function of size, and my first measurement was taken at one size.** An early sweep at 10 m found 1.001–1.06× and I concluded thinning does not buy survival. That is true at 10 m and false at 20 m+.

## What #617 changed, measured before and after

The benefit band sits at the height-resistance relation's viability ceiling, so replacing that relation moved both:

| | before #617 | after |
|---|---|---|
| viability ceiling (`P > 0` in wet soil) | 17–18 m | **30–36 m** |
| benefit at soil 0.16, 4 yr, any height ≤24 m | up to 4.4e9× | **exactly 1.00** |
| largest benefit found | 6.1e9× (both arms dead) | 84× at 30 m |

TF24 is now much more drought-tolerant: a 4-year drought at soil 0.16 kills nothing up to 24 m, where before it took a 20 m plant to 3e-19. The size dependence survives, shifted — nothing below 13 m, 2.4× at 20 m, 84× at 30 m at soil 0.14.

**Shedding still never converts death into survival.** Zero cells of the post-#617 grid have the fixed plant dead and the shedding plant alive; the large ratios sit between two very small numbers. The benefit is also largest at *moderate* drought, not the harshest — at soil 0.12 everything dies whatever it does. This is the open scientific question, and it is why the demo carries a section on it.

## Merge resolutions worth reviewing

- **`k_max` takes #617's path-integral denominator and this branch's actual Huber value as numerator**, and is now stored rather than recomputed. The controller had been recomputing it on the old height-linear relation — wrong by a factor 0.35 at 16 m, and silent. This is the resolution I would most like a second pair of eyes on.
- **`scientific_version → 11`.** Both branches independently claimed v10; a merged TF24 carries both changes and is comparable to neither.
- **#617's exactness test is pinned to the height coordinate.** This branch changes TF24's default, so a bare `Control()` compared two coordinates rather than two hydraulic models (404.5 against 30.2) — a failure that reads exactly like a broken exactness test.
- **The TF24f `k_acclim` bracket moved rather than its threshold relaxed.** The peak shifted from k≈0.1 to k≈1 under the birth-date coordinate, so 0.1-vs-10 straddled the maximum and read 3% — a badly-placed bracket, not weak acclimation. Spanning 0.001–1 gives 13.2%.

## Also fixed along the way

- **Sapwood turnover is gated, so its ψ-derivative is not `k_s·m_s`.** The charge falls as `exp((1 − 1/a_pl2)ψ)`. Differentiating the ungated cost dropped a term worth 6.8 of 13.1 kg/yr — invisible to any check on sign or monotonicity.
- **The controller's denominator must stay positive**, and `profit` is not: it goes negative in exactly the drought where the sign is most needed. It now divides by the maintenance bill.
- **Re-proportioning has to be paid for.** The reserve gate `G` is *not* the right switch — at soil 0.13 reserves are still full and it is production that has gone to zero.
- **Deep-crown is refused, not skipped** (`prepare_strategy()` throws). Skipping leaves the controller with only its negative cost term, and the stem shrinks without bound while the run looks plausible the whole way down.

## Status

Full suite green at 3738, and **the scenario gateway passes unmodified** — the strongest available statement that `a_sw = 0` is bit-exact against develop.

Draft because the demo's "thinning does not buy survival" section still quotes pre-#617 numbers. Its chunks run, so nothing fails; the prose beside them is what is stale.

dfalster:
@itowers1 @elijahmagistrado — feedback welcome on this one, whenever suits.

**Please run the demo rather than only reading the diff.** It is the honest account of what this does and, more usefully, of where it does not yet work:

```
git fetch && git checkout feature/flexible-allometry-516
R CMD INSTALL --preclean .          # header + yml changed, so a clean build
quarto render overstorey_staging/TF24_flexible_allometry_demo.qmd
```

It runs in a few minutes and needs nothing beyond the branch. If you would rather not build it, say so and I will render and attach the HTML.

**What I would most like challenged, in priority order:**

1. **The claim that shedding does not buy survival.** Across a 90-cell grid of size, drought intensity and duration, there is no cell where the fixed plant dies and the shedding plant lives. That is the central negative result and I would like it attacked — either a regime I have not tried, or a reason the experiment is the wrong one. The demo's section on this is the part still carrying pre-#617 numbers, so read the prose sceptically.

2. **Whether the growth rate is the right objective.** `R_s` is zero where `dh/dt` peaks. Height growth is one choice; `dA_leaf/dt`, or the relative rate `(dA/dt)/A`, are defensible and give different optima. Daniel's view is that this is ultimately an empirical question for evolutionary analysis of competing strategies, but if either of you has a prior I would rather hear it before we run that.

3. **The merged `k_max`.** It takes #617's path-integral denominator with the plant's actual Huber value as numerator. I think that is right — supply per unit leaf area should follow the stem the plant actually has — but it is the one place where this branch and #617 genuinely interact, and it was silently wrong once already.

4. **Whether the parameters are the right ones to expose.** `a_pl0`–`a_pl4` and `a_sw` are all `TF24_Pars` fields and so settable as traits. Defaults reproduce the current model exactly. Whether any of them should instead be *derived* from an existing trait is open, and #512 is the input.

The first comment above has the measurements and two retractions. No rush on any of this — it is draft and the demo needs a pass before it is merge-ready.

dfalster:
Ready for review — the stale section is re-measured.

@itowers1 @elijahmagistrado the demo is now safe to read end to end. It renders clean and the full suite is green at 3753.

**One argument in it is retracted outright**, and it is worth knowing about before you read the section. It used to say that thinning "cuts income just as fast" — that shrinking the canopy shrinks both sides of the budget together, with per-leaf production worsening monotonically from −3.81 to −4.68 as the canopy went to a fiftieth. **That is no longer true after #617.** With resistance derived from stem anatomy, the remaining leaves inherit the whole stem's hydraulic supply, so thinning *improves* the per-leaf balance by 43% and there is an interior optimum near a tenth of full canopy:

| canopy fullness | 1.00 | 0.50 | 0.20 | 0.10 | 0.05 | 0.018 |
|---|---|---|---|---|---|---|
| net production per m² leaf | −1.47 | −1.17 | −0.94 | **−0.84** | −0.86 | −0.99 |

The conclusion survives the argument being replaced: a struggling plant can now *narrow* its deficit by shedding, but it still cannot close it. Thinning slows the burn; it never reaches balance.

**The stand-level cost also shrank a lot.** Switching the mechanism on used to cost 0.55–0.64× lifetime offspring production on seasonal stands. It is now 0.92–1.00×, so the mechanism is close to free rather than expensive — but still buys no survival, which is what the section is about.

**The decisive measurement is unchanged and now guarded.** The two thresholds are in the wrong order in stress space: at soil 0.145 nothing thins because the leaves are still profitable, and by 0.140 a 10 m plant is at a survivorship of one in a million *with its canopy still at 1.000*. Mortality saturates at a milder drought than shedding begins at, so there is no intensity where a plant is both stressed enough to thin and alive enough to profit. Across the grid the largest gain is 20×, and it moves survivorship from 2.8e-08 to 5.6e-07 — both dead.

**Why this rotted, and what stops it next time.** The section's tables were prose, not chunks, so every number in it survived #617 while the page rendered clean. `test-allometry-demo.R` now asserts the ordering, the absence of any rescue cell, the interior optimum, and the trajectory's height cost — against `demo_soil_history`'s own arguments, since quoting a number from a nearby-but-different run is how an earlier Huber-value claim went wrong.

The four things I would most like challenged are unchanged, in the comment above.

src/tf24_strategy.cpp
itowers1: Unclear what Ppos does, it appears to ensure that there is a small amount of production when P is 0?

src/tf24_strategy.cpp
```diff
@@ -246,18 +308,70 @@ void TF24_Strategy::compute_rates(const TF24_Environment& environment,  Internal
   const double growth_flux = Ppos * G;
```
itowers1: Ppos requires documentation regarding role in smooth negative function

itowers1:
Just adding my current understanding of how this all works via a flowchart, this can be edited in powerpoint as I learn more.
Slide1

itowers1:
In terms of optimisation of allometry, if the purpose of thinning is to stave off mortality in drought periods, would there be a functional difference to maximising survivorship as opposed to growth. Is survival maximised where height growth is greatest anyway?

elijahmagistrado:
I think once the trees reach their height at maturation and af_1 = 1, then the purpose of thinning would be exclusively for reducing mortality. So in wet periods, survival would be the highest where height growth is also the highest when there is still room to grow, otherwise the trade-off is between survival and reproduction.

PR #648, opened 2026-09-19
Differentiate a stand's census by trait

aornugent:
**Needs** odelia `v0.5.0` and phylloptim `v0.9.0` tagged first

A leaf slightly cheaper to build shades its neighbours slightly more, so
they grow slightly less, so they shade it back slightly less. Fitting a
trait means asking what a century of that does to a stand summary, for
all forty-eight TF24 traits at once. Differencing costs forty-eight
re-runs and does not converge where a trait moves an event in the run.

`stand_gradient(scm)` returns that derivative exactly, from one run, as
`list(value, gradient, refusal, control)`. The model is templated on its
scalar: `Internals` becomes `Internals<double>` and `Individual::state`
returns `const value_type&`. `run_scm()` takes `record_trajectory` where
it took `use_ode_times`; `NodeSchedule` holds one row per introduction
time, so `$size` counts those; `Interpolator` is gone and
`ResourceSpline` carries height, value and slope. TF24 numbers move:
`GSS_tol_abs` 1e-3 to 1e-1, `vulnerability_curve_ncontrol` 100 to 400.


aornugent:
## High level overview

Run a stand, then ask how its census responds to every trait at once.

```r
p   <- scm_base_parameters("TF24")
p   <- add_strategies(p, trait_matrix(c(0.0825, 5.13), c("lma", "hmat")),
                      hyperpar = TF24_hyperpar, birth_rate = list(1.10))
scm <- run_scm(p, Environment("TF24"),
               Control(node_density_in_birth_date = TRUE),
               refine_schedule = TRUE, record_trajectory = TRUE)

g <- stand_gradient(scm)

dim(g$gradient)
#> [1]  3 48        three metrics by forty-eight trait columns, from one pass

g$gradient["mass_above_ground", "1.lma"]
#>                  d(mass above ground)/d(lma) for species 1, over the whole run
```

One run forward, one sweep back, and every cell is filled. Differencing would
cost forty-eight re-runs and would not converge where a trait moves an event.

Three things a caller must know before the rest makes sense.
`Control(node_density_in_birth_date = TRUE)` is required and defaults to `FALSE`.
`refusal` carries a stated reason where a metric has no derivative, and NaN
without a reason is a bug rather than an answer. And `control` records the five
settings the gradient was taken at, so `stand_gradient_compare()` can refuse two
gradients taken differently.

The number is a derivative of the emergent stand, not of a plant in isolation.
Environmental feedback suppresses the response to leaf mass per area and
reverses the sign of the response to seed mass — for seed mass, a calculation
holding the neighbours fixed gives the wrong direction.

## Terms

- **census metric** — a stand reduced to one scalar. TF24 declares three: `leaf_area`, `mass_above_ground`, `area_stem`.
- **row** — one output's derivatives against a list of inputs. phylloptim supplies the leaf's rows; the sweep never records its solve.
- **refusal** — a declared reason a metric has no derivative, carried beside NaN.
- **range** — consecutive recorded steps at constant state width. A System of fixed width has one; plant opens a new one at every cohort introduction.

Of TF24's 67 parameter-table entries, 19 are `undifferentiable`, leaving 48
columns per species. Columns are named `"1.lma"` — species-major, because two
species carrying the same trait need distinguishing and R's character indexing
would otherwise return species one's column for both.

## Control flow

```
  stand_gradient(scm)                         R/stand_gradient.R
    │  resolve names, refuse ones the model lacks
    ▼
  census_trait_gradient_tf24                  src/census_gradient.cpp
    │  the only place double becomes an active scalar
    ▼
  SCM::census_trait_gradient                  scm.h
    │
    ├─► census_state_and_trait_rows()
    │     one recording at the final time, swept for two things:
    │       state rows   d(metric)/d(final state)
    │       trait rows   d(metric)/d(trait), read directly
    │
    ├─► trait_adjoint STARTS at the trait rows; lambda at the state rows
    │
    ├─► solver.solve_adjoint(...)             odelia
    │     │
    │     ├─► range N … 1                     widest first
    │     │     │
    │     │     ├─► step k … 1                last to first
    │     │     │     └─ re-run 6 stages at the active scalar
    │     │     │          └─ Patch::compute_rates
    │     │     │               └─ Species → Node → Individual
    │     │     │                    └─ TF24_Strategy::solve_leaf
    │     │     │                         └─ record_leaf_outputs()
    │     │     │                              reads phylloptim's supplied rows
    │     │     │
    │     │     └─► at an introduction: sweep apply_insertion,
    │     │              so a newborn's initial conditions carry trait rows
    │
    └─► census_gradient{ gradient[][], why }
```

Scale: a stand of `N` cohort introductions is swept as `N` ranges, with six rate
evaluations in every accepted step. `N` is the schedule's, so it is a property of
the stand rather than of the gradient. The ratio holds across the stands measured:
a gradient over all traits costs about 2.9 times a forward run of the same
stand, with the two arms interleaved in one sitting.

## What changes for existing code

| old | new |
|---|---|
| `run_scm(..., use_ode_times, ...)` | `run_scm(..., record_trajectory, ...)` — same position, different meaning, so a positional caller gets the wrong flag |
| `Interpolator()` | gone; `ResourceSpline` carries height, value and slope |
| `sched$next_event` | `sched$next_introduction` |
| `sched$ode_times <- x` | `sched$set_ode_steps(times, sizes)` |
| `sched$size` | counts distinct introduction times, where it counted one entry per species per time |
| `Control$save_RK45_cache` | gone; `Control$gradient_curvature_floor` added |
| `patch$introduce_new_node(i)` | `patch$introduce_new_node(i, time)` |
| `plant::Internals` | `Internals<double>` |
| `Individual::state` returning `double` | returning `const value_type&` |
| `plant/adaptive_interpolator.h` | odelia's `interpolator.hpp` |
| `plant/optimize.h` | deleted; it had no callers |

Forward numbers move for every TF24 run whether or not a gradient is taken:
`GSS_tol_abs` 1e-3 → 1e-1, `vulnerability_curve_ncontrol` 100 → 400. FF16 and
K93 go `scientific_version` 1 → 2 on the birth-date coordinate.

Two pinned values move with them, and `R CMD check` is red on both:
`test-mutant.R`'s ten-mutant panels, worst 4.1e-4 and 6.0e-4 relative on FF16.
The machinery under them is exact — a strategy replayed as an invader of itself
returns the resident's own fitness to 1e-13 on FF16 and 4e-15 on TF24 — and the
drift has a referee that owes `develop` nothing.
Invasion fitness is what a strategy attains when vanishingly rare, so running the
mutant endogenously at birth rate e and letting e → 0 gives a referee the pins
cannot appeal to. That limit is this branch's replayed value, approached at
first order in e; `develop`'s pin sits off it. The same drift is already visible
in the resident pins that pass, at −2.2e-5 and +4.7e-5. Re-pinning accepts a change
to the model's science, so it wants a `scientific_version` decision ahead of any
`snapshot_accept`.

On Linux the check reads `FAIL 2 | SKIP 5 | PASS 4641`, with one NOTE and one
WARNING; the WARNING is the check host reporting it cannot set a UTF-8 locale.
The NOTE is `stderr` in `census_gradient.o` and
`gradient_ladder.o`, which are the two objects carrying an adjoint sweep. It comes from odelia's
`ODELIA_ADJOINT_TRACE` diagnostic, dead unless that variable is set in the
environment, and not from `odelia::util::warning` — `RcppR6.o` includes that
helper through `scm.h` and references no `stderr` at all.

The gradient is TF24's. `census_gradient.cpp` names `TF24_Strategy` throughout
and `tf24_strategy.h` is the only file declaring `census_metrics`, so an FF16
stand handed to `stand_gradient()` fails in an `Rcpp::as` type error and not
a model-level refusal.

## Files

| file | lines | what |
|---|---|---|
| `models/tf24_strategy.h` | +2365 −211 | the strategy, templated; absorbs the deleted `.cpp` |
| `scm.h` | +964 −98 | the solver and `census_trait_gradient` |
| `patch.h` | +881 −358 | the stand; `at_scalar`, `plant::tangent` |
| `species.h` | +484 −214 | per-species accumulation |
| `models/tf24_environment.h` | +396 −136 | light and soil, templated |
| `census.h` | 59 | new — `census_metric<Strategy>`, `Censusable` |
| `census_gradient.h` | 59 | new — `census_gradient`, `refusal` |
| `clamp_sites.h` | 95 | new |
| `with_slope.h` | 38 | new — alias of odelia's |

Deleted: `adaptive_interpolator.{h,cpp}`, `optimize.h`, and the two strategy
bodies `tf24_strategy.cpp` and `tf24f_strategy.cpp`, which the header absorbed.
`NAMESPACE` gains six and loses one: `stand_gradient`, `stand_census`,
`stand_census_state_adjoint`, `stand_gradient_compare`, `stand_gradient_refused`
and `gradient_control` arrive, `Interpolator` goes. That is the whole of the
public R surface this adds. `R/` gains 82 functions and loses 22; of the 76 that
are not those six, 74 are `RcppExports.R`'s, regenerated from the export
attributes, and two are private helpers in `stand_gradient.R`.

Read `census.h` and `census_gradient.h` first — 118 lines, and they say what a
metric and a refusal are. Then `SCM::census_trait_gradient` end to end,
which is the shortest complete path through the change. Then
`census_state_and_trait_rows`, then `solve_leaf` and `record_leaf_outputs` for
the seam with phylloptim. `patch.h`, `species.h` and `node.h` are mechanical after that.

aornugent:
## How it works 

A census metric integrates a per-plant quantity over the size distribution, and a
trait reaches the answer by two routes.

The **trajectory term** is what the sweep produces. Changing `lma` changes growth
and mortality, and it changes the light and water every other plant is competing
for, all the way along the run — so it changes the size distribution standing at
the end. This is the term that needs a whole reverse pass.

The **allometric term** is what survives when that distribution is held
identical. A cohort of a given height reads a different leaf area under a
different `lma`, because leaf area is itself a function of the trait
in `area_leaf`. No sweep produces this; it comes from differentiating
the census expression directly. The boundary recruit node belongs here too — it
is the trapezium's lower grid point and is not ODE state.

Both come out of one recording at the final time, swept twice with different
seeds. The allometric term could then be added to the sweep's
answer at the end. It is used as the walk's starting value instead, and
`census_trait_gradient` gives the reason at the site: *"a term added last is a term that can be left
out, and a gradient missing it is a plausible number rather than an error."*
A missing trajectory term produces obvious nonsense; a missing allometric term
produces a number of the right order that is quietly wrong.

## Why the model became a template

`Internals` holds the state, rates and auxiliary values a strategy reads and
writes. The sweep re-enters `Patch::compute_rates` at an active scalar, so
`Internals` has to exist at that scalar — and so does everything that touches it:
`Individual`, `Node`, `Species`, `Patch`, `SCM`, the environment and the strategy
itself.

Twenty-three headers transitively include `internals.h`. That is why this arrives
as one commit: any split leaves between 3 and 17 headers that do not compile, and
a sequence of commits that do not build is worse to review than one that does.

FF16 and K93 are not templated. They pin `using value_type = double` and never
instantiate the active path, so only TF24 pays the compile cost — which is real,
since `tf24_strategy.h` grew from 771 lines to 2,925 by absorbing its `.cpp`, and
twelve of the twenty-six translation units now recompile it.

`Individual::state` returns `const value_type&` where it returned `double`.
Copying an active scalar registers a tape slot and records an operation, once per
element per stage, for a value the caller was about to assign anyway.

## Where a derivative stops, and how to lose one by accident

The boundary between the differentiated interior and the plain-`double` world is
`write_iterator_scalar`:

```cpp
  if constexpr (std::floating_point<
                  typename std::iterator_traits<It>::value_type>) {
    *it++ = odelia::util::to_passive(value);   // derivative dropped here
  } else {
    *it++ = value;
  }
```

At the R boundary this is exactly right: nothing in R holds an active scalar, so
there has to be one place where the derivative is deliberately discarded, and
keying it on the destination's type means the conversion cannot be forgotten.

It is also type-directed and completely silent, which makes it the easiest way to
lose a gradient in this codebase. A `std::vector<double>` scratch buffer
introduced anywhere on the rate path zeroes every derivative passing through it,
with no diagnostic and every number still finite.
`TF24_Environment::resource_uptake` is already such a buffer. Any new container
on an active path is worth checking for its `value_type` before anything else.

## Refusal takes the whole metric

A census metric is a sum over cohorts, and a sum has no defined value when one of
its terms is undefined. So if any cohort's contribution has no derivative, the
metric's whole row is NaN with one stated reason,

aornugent:
*Note for maintainers*

This is my first go writing something this big with Claude Code. Big enough that I haven't read and understood every part of it completely.

There are more changes than I would like. Where there was a blocker, Claude proposed options and I considered the different trade-offs, but it was often iterative. I did not succeed in implementing a preplanned design. Working across multiple sessions, trying and reverting different approaches, means there might be rough edges or dead code. There are parts that could be done simpler or better.

I am a little concerned about increasing the maintenance debt of the plant project. My experience so far is that coding agents/Claude don't have the wherewithall to say 'this is poorly designed' but bend over backwards to accomodate the status quo. As a consequence, residue and poor abstractions accumulate, with volume and indirection making it hard to read and understand.

Hopefully, though, we can build upon this implementation and continue to mature and refine plants reverse mode. Please take this message as an invitation to change and improve (or reject!) things.

